PR #1, opened 2021-11-11
Topology change

bdeggleston:


accord-core/src/main/java/accord/coordinate/Agree.java
```diff
@@ -45,11 +49,76 @@ public boolean hasMetFastPathCriteria()
         }
     }
 
+    static class PreacceptTracker extends FastPathTracker<ShardTracker>
+    {
+        volatile long supersedingEpoch = -1;
```
belliottsmith, line 54: I assume this is volatile so the configuration service (which is unrelated to the `Instance`) may invoke directly, but would it be easier to ask the `Instance` to perform epoch work so we can avoid worrying about the concurrency?

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -116,37 +165,43 @@ public Timestamp uniqueNow()
...
-            // TODO: this diverges from proof; either show isomorphism or make consistent
-            long now = nowSupplier.getAsLong();
-            if (now > cur.real) return new Timestamp(now, 0, id);
-            else return new Timestamp(cur.real, cur.logical + 1, id);
-        });
+            now.accumulateAndGet(atLeast, (current, proposed) -> {
+                Timestamp timestamp = proposed.compareTo(current) <= 0
+                        ? new Timestamp(current.epoch, current.real, current.logical + 1, id)
+                        : proposed;
+                return timestamp.withMinEpoch(topology.epoch());
```
belliottsmith: Could we just return `proposed.compareTo(current) <= 0 ? new Timestamp(max(current.epoch, proposed.epoch), max(current.real, proposed.real), current.logical + 1, id) : proposed;`?

Then we could call `uniqueNow` on witnessing a new epoch with `new Timestamp(epoch, 0, 0, id)` and it should take care of itself without polling this value on each update? We probably anyway want to ensure we don't go back in `real` time so we don't mess up any timestamps we'll generate for Cassandra

accord-core/src/test/java/accord/burn/TopologyUpdate.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+public class TopologyUpdate
+{
+    private static final Logger logger = LoggerFactory.getLogger(TopologyUpdate.class);
+
+    private static class CountDownFuture<T> extends CompletableFuture<T>
```
belliottsmith, line 33: I think we should consider inserting a cassandra.jar in our dependencies for now, so we can use the same facilities we do in mainline until we can figure out a good way to share code. WDYT?

belliottsmith:
<test email doesn't go to commits@>

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -274,22 +278,36 @@ public Topologies forKeys(Keys keys)
             for (int i=1; i<current.epochs.length; i++)
             {
                 epochState = current.epochs[i];
-                if (i > 1 && epochState.syncCompleteFor(keys))
+                if (i > 1 && epochState.syncCompleteFor(keys) && epochState.epoch() < minEpoch)
```
belliottsmith, line 281: Should this be `<=` ?

accord-core/src/main/java/accord/topology/Topologies.java
```diff
@@ -0,0 +1,272 @@
...
+    // topologies are stored in reverse epoch order, with the highest epoch at idx 0
+    Topology get(int i);
+
+    int size();
+
+    int totalShards();
+
+    Set<Node.Id> nodes();
+
+    Topologies withMinEpoch(long epoch);
```
belliottsmith: Perhaps `truncateWithMinEpoch`? Or `removeEpochsBefore`?

accord-core/src/main/java/accord/coordinate/tracking/AbstractResponseTracker.java
```diff
@@ -2,22 +2,26 @@
...
 import java.util.function.Predicate;
 
-abstract class AbstractResponseTracker<T extends AbstractResponseTracker.ShardTracker>
+public abstract class AbstractResponseTracker<T extends AbstractResponseTracker.ShardTracker>
 {
-    private final Shards shards;
+    private static final int[] SINGLETON_OFFSETS = new int[0];
+    private final Topologies topologies;
     private final T[] trackers;
+    private final int[] offsets;
```
belliottsmith, line 22: suggestion: can simply compute the `sizePerTopology` and leave the last elements of `trackers` null for those topology with fewer trackers, as most topology will have around the same number of trackers?

accord-core/src/main/java/accord/api/ConfigurationService.java
```diff
@@ -0,0 +1,56 @@
...
+        /**
+         * Informs listeners of new topology. This is guaranteed to be called sequentially for each epoch after
+         * the initial topology returned by `currentTopology` on startup.
+         */
+        void onTopologyUpdate(Topology topology);
+
+        /**
+         * Called when accord data associated with a superseded epoch has been sync'd across current replicas
+         */
+        void onEpochSyncComplete(Node.Id node, long epoch);
```
belliottsmith, line 55: Maybe add some additional explanation about what implementations need to guarantee before they invoke this?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -0,0 +1,325 @@
...
+        private boolean syncComplete = false;
+        private boolean prevSynced;
+
+        private boolean updateState(boolean prevSynced)
+        {
+            this.prevSynced = prevSynced;
+            if (!syncComplete)
+            {
+                syncComplete = syncTracker.hasReachedQuorum();
+                return syncComplete;
```
belliottsmith: Should this be `prevSynced && syncComplete`? Or perhaps we should call the method `onPrevSynced()` so that we cannot provide a value of `false`?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -0,0 +1,325 @@
...
+            return !syncTracker.unsafeGet(idx).hasReachedQuorum();
+        }
+    }
+
+    private class Epochs
+    {
+        private final long maxEpoch;
+        private final long minEpoch;
+        private final EpochState[] epochs;
+        private final List<Set<Node.Id>> pendingSyncComplete;
```
belliottsmith, line 98: A comment indicating what this collection maintains, and in what epoch order, could be helpful

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -0,0 +1,325 @@
...
+        public Epochs add(Topology topology)
+        {
+            Preconditions.checkArgument(topology.epoch == nextEpoch());
+            EpochState[] nextEpochs = new EpochState[epochs.length + 1];
+            List<Set<Node.Id>> pendingSync = pendingSyncComplete;
+            if (!pendingSync.isEmpty())
+            {
+                boolean prevSynced = epochs.length <= 1 || epochs[1].syncComplete();
+                EpochState currentEpoch = epochs[0];
+                pendingSync.remove(0).forEach(id -> currentEpoch.recordSyncComplete(id, prevSynced));
```
belliottsmith: Why do we remove this element?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -0,0 +1,325 @@
...
+
+                pendingSyncComplete.get(idx).add(node);
+            }
+            else
+            {
+                boolean prevSynced = epoch == minEpoch || get(epoch - 1).syncComplete();
+                boolean hasChanged = get(epoch).recordSyncComplete(node, prevSynced);
+                for (epoch++ ;hasChanged && epoch <= maxEpoch; epoch++)
+                {
+                    prevSynced = get(epoch - 1).syncComplete();
```
belliottsmith: why do we keep going if `!prevSynced`?

accord-core/src/main/java/accord/messages/TxnRequestScope.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Indicates the ranges the coordinator expects the recipient to service for a request, and
+ * the minimum epochs the recipient will need to be aware of for each range
+ */
+public class TxnRequestScope
+{
+    public static class EpochRanges
```
belliottsmith: `EpochDomain`?  `KeysForEpoch`?

accord-core/src/main/java/accord/messages/TxnRequestScope.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Indicates the ranges the coordinator expects the recipient to service for a request, and
+ * the minimum epochs the recipient will need to be aware of for each range
+ */
+public class TxnRequestScope
```
belliottsmith: Make inner class of `TxnRequest`?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -0,0 +1,325 @@
...
+
+        private EpochState get(long epoch)
+        {
+            if (epoch > maxEpoch || epoch < maxEpoch - epochs.length)
+                return null;
+
+            return epochs[(int) (maxEpoch - epoch)];
+        }
+
+        long canProcess(TxnRequestScope scope)
```
belliottsmith: Document what this returns? Or perhaps call it `maxUnknownEpoch`?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -0,0 +1,325 @@
...
+                    return scope.maxEpoch();
+                }
+
+                // validate requested ranges
+                KeyRanges localRanges = epochState.local.ranges();
+                if (!localRanges.intersects(requestRanges.keys))
+                    throw new RuntimeException("Received request for ranges not replicated by this node");
+            }
+            if (scope.maxEpoch() > 0)
+                missingEpochNotify.accept(scope.maxEpoch());
```
belliottsmith: Are we invoking this unnecessarily?

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -103,40 +108,27 @@ static void burn(TopologyFactory topologyFactory, List<Id> clients, List<Id> nod
...
-
-    static void burn(Random random, TopologyFactory topologyFactory, List<Id> clients, List<Id> nodes, int keyCount, int operations, int concurrency)
-    {
-        burn(random, topologyFactory, clients, nodes, keyCount, operations, concurrency, System.out, System.err);
+        burn(random, topologyFactory, clients, nodes, keyCount, operations, concurrency);
     }
 
     static void reconcile(long seed, TopologyFactory topologyFactory, List<Id> clients, List<Id> nodes, int keyCount, int operations, int concurrency) throws IOException, ExecutionException, InterruptedException, TimeoutException
     {
-        ReconcilingOutputStreams streams = new ReconcilingOutputStreams(System.out, System.err, 2);
```
belliottsmith, line 125: Was this intentionally removed?

accord-core/src/main/java/accord/coordinate/tracking/ReadTracker.java
```diff
@@ -109,7 +111,7 @@ private int compareIntersections(Id left, Id right, Set<ReadShardTracker> target
     public Set<Id> computeMinimalReadSetAndMarkInflight()
     {
         Set<ReadShardTracker> toRead = accumulate((tracker, accumulate) -> {
-            if (!tracker.shouldRead())
+            if (tracker.epoch != topologies().currentEpoch() || !tracker.shouldRead())
```
belliottsmith: Should we just accumulate on the trackers for the relevant epoch directly?

accord-core/src/test/java/accord/burn/BurnTest.java
bdeggleston: yes it was removed when I added logging. Looking closer, it appears to be comparing the output of 2 identical runs in the reconcile method?

belliottsmith: yeah, it's to ensure runs are deterministic/reproducible

accord-core/src/main/java/accord/topology/TopologyManager.java
bdeggleston: no, but the variable name and documentation on the method it is meant to ultimately call are unclear

accord-core/src/main/java/accord/topology/TopologyManager.java
bdeggleston: `pendingSync` holds sync notifications we've received for future epochs (or the current epoch), and is indexed from the current epoch. Since we're adding an epoch here, the contents of pendingSync at index 0 are no longer for the current epoch, so they're removed from the list and applied to the epoch state.

belliottsmith: Thanks, I'm just confused why that translates into removing index 0 rather than pushing a new collection at index 0 and shifting the remainder down? If we're adding an epoch aren't we invalidating the whole sequence. I will be reading this all again more carefully today, hopefully with my brain working better, so hopefully I'll grok it properly.

accord-core/src/main/java/accord/topology/TopologyManager.java
belliottsmith: Thanks. The confusion for me was that we also seemed to invoke this in `Node.receive` if we _didn't_ know an epoch. I can see now that this method is intended to be used differently by implementations, but it seems to also perform the same fetching by default. Are we relying on this behaviour? If not, might it be better (particularly from a test perspective) to have this no-op by default?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -0,0 +1,326 @@
...
+
+        boolean shardIsUnsynced(int idx, Shard shard)
+        {
+            return !prevSynced || !syncTracker.unsafeGet(idx).hasReachedQuorum();
+        }
+    }
+
+    private class Epochs
+    {
+        private final long maxEpoch;
```
belliottsmith: Perhaps call this `currentEpoch` for consistency with the comments and other methods? Or migrate those to `maxEpoch`?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -0,0 +1,326 @@
...
+        private final long minEpoch;
```
belliottsmith: unused now

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -0,0 +1,326 @@
...
+    {
+        return epochs.get(epoch);
+    }
+
+    public Topologies forKeys(Keys keys, long minEpoch)
+    {
+        Epochs current = epochs;
+
+        EpochState epochState = current.get(current.maxEpoch);
+        Topology topology = epochState.global.forKeys(keys);
```
belliottsmith: `currentTopology`?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -0,0 +1,326 @@
...
+
+    @VisibleForTesting
+    EpochState getEpochStateUnsafe(long epoch)
+    {
+        return epochs.get(epoch);
+    }
+
+    public Topologies forKeys(Keys keys, long minEpoch)
+    {
+        Epochs current = epochs;
```
belliottsmith: `snapshot`?

accord-core/src/main/java/accord/topology/TopologyManager.java
bdeggleston: pendingSync is indexed with the current epoch at 0, plus epochs we haven't seen yet indexed after that. So if the current epoch is 5, the epochs represented in pending sync would be [5, 6, 7]. If we receive a new epoch topology (6), the pending sync entry representing epoch 5 must be removed for this indexing assumption to be valid

accord-core/src/main/java/accord/txn/Keys.java
```diff
@@ -115,4 +132,52 @@ public static Keys of(Key k0, Key... kn)
...
+        int keyLB = 0;
+        int keyHB = size();
+        int rangeLB = 0;
+        int rangeHB = ranges.rangeIndexForKey(keys[keyHB-1]);
+        rangeHB = rangeHB < 0 ? -1 - rangeHB : rangeHB + 1;
+
+        for (;rangeLB<rangeHB && keyLB<keyHB;)
+        {
+            Key key = keys[keyLB];
+            rangeLB = ranges.rangeIndexForKey(rangeLB, ranges.size(), key);
```
belliottsmith, line 150: ranges.size()->rangeHB?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -0,0 +1,324 @@
...
+ */
+public class TopologyManager implements ConfigurationService.Listener
+{
+    class EpochState
+    {
+        private final Topology global;
+        private final Topology local;
+        private final QuorumTracker syncTracker;
+        private boolean syncComplete = false;
+        private boolean prevSynced = false;
```
belliottsmith, line 34: always provided in constructor, so perhaps simply set it to the parameter's value?

accord-core/src/main/java/accord/topology/Topology.java
```diff
@@ -61,25 +58,93 @@ public Topology(Shard... shards)
...
+
+        return new Topology(epoch, subset);
+    }
+
+    public boolean isSubset()
+    {
+        return supersetRangeIndexes.length < shards.length;
+    }
+
+    public boolean isSubsetOf(Topology topology)
```
belliottsmith, line 121: unused

accord-core/src/main/java/accord/messages/WaitOnCommit.java
```diff
@@ -49,46 +53,55 @@ public synchronized void onChange(Command command)
 
         private void ack()
         {
-            if (--waitingOn == 0)
+            if (waitingOn.decrementAndGet() == 0)
                 node.reply(replyToNode, replyToMessage, new WaitOnCommitOk());
         }
 
-        synchronized void setup(TxnId txnId, Keys keys)
+        void process(CommandStore instance)
```
belliottsmith, line 60: `process` is an ambiguous name - ideally this would also be called `setup` or something equivalent, since it is only used on setup.

accord-core/src/main/java/accord/messages/BeginRecovery.java
```diff
@@ -118,32 +132,46 @@ public void process(Node node, Id replyToNode, long replyToMessage)
...
+            if (ok.executeAt.epoch > configService.currentEpoch())
+            {
+                configService.fetchTopologyForEpoch(ok.executeAt.epoch, () -> disseminateApply(node, ok));
+                return;
+            }
+            disseminateApply(node, ok);
         }
     }
 
+    private void disseminateApply(Node node, RecoverOk ok)
```
belliottsmith, line 160: Are we confident it is safe to only send to the latest epoch, and not the possibly earlier `executeAt` epoch, whose members might be waiting on the result to e.g. finish transactions they must sync to the new members?

I can always assess as part of liveness work when I rebase onto this patch.

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -116,84 +161,74 @@ public Timestamp uniqueNow()
...
         if (now.get().compareTo(atLeast) < 0)
-            now.accumulateAndGet(atLeast, (a, b) -> a.compareTo(b) < 0 ? new Timestamp(b.real, b.logical + 1, id) : a);
-
-        return now.updateAndGet(cur -> {
-            // TODO: this diverges from proof; either show isomorphism or make consistent
-            long now = nowSupplier.getAsLong();
-            if (now > cur.real) return new Timestamp(now, 0, id);
-            else return new Timestamp(cur.real, cur.logical + 1, id);
-        });
+            now.accumulateAndGet(atLeast, (current, proposed) -> {
```
belliottsmith, line 174: I think this has changed the semantics of this step? It seems to have been flipped, to take `atLeast` itself if it is larger (which might have the wrong `node`, but fortunately this is fixed in `uniqueNow`) but to increment current unnecessarily otherwise.

I suspect previously there was a missing `return` also, as the prior increment to `proposed` (and overwriting of `node` makes it unique immediately.

We could keep the basic flow of this approach, but remove the `logicalNext(id)` application to `current`, but I would probably prefer at least overwriting `id` of `proposed` to ensure the global counter always has the local node `id`.

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -91,19 +86,69 @@ public static int numCommandShards()
...
+        this.configService = configService;
+        this.topology = new TopologyManager(id, configService::reportEpoch);
+        Topology topology = configService.currentTopology();
+        this.now = new AtomicReference<>(new Timestamp(topology.epoch(), nowSupplier.getAsLong(), 0, id));
         this.nowSupplier = nowSupplier;
         this.scheduler = scheduler;
-        this.commandStores = new CommandStores(numCommandShards(), id, this::uniqueNow, agent, dataSupplier.get(), commandStoreFactory);
-        this.commandStores.updateTopology(cluster.forNode(id));
+        this.commandStores = new CommandStores(numCommandShards(),
+                                               id,
```
belliottsmith, line 102: I know this is technically the spec for formatting in the project, but it's my pet peeve as I think it really harms legibility. I haven't really wanted to bring up revisions, but maybe I will do soon. In this case, we only slightly overspill the one line limit so one line would be fine, but two or three lines of grouped-related parameters would be much more legible IMO.

e.g.

```
        this.commandStores = new CommandStores(numCommandShards(), id, this::uniqueNow, 
                                               agent, dataSupplier.get(), commandStoreFactory);
```

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -283,13 +260,35 @@ void processInternal(Consumer<? super CommandStore> consumer, CompletableFuture<
...
 
+    public void processBlocking(Consumer<? super CommandStore> consumer)
+    {
+        try
+        {
+            process(consumer).toCompletableFuture().get();
+        }
+        catch (InterruptedException e)
+        {
+            Thread.currentThread().interrupt();
```
belliottsmith, line 271: Should we propagate some kind of exception so this doesn't effectively report success?

accord-core/src/main/java/accord/messages/WaitOnCommit.java
```diff
@@ -1,27 +1,31 @@
...
+        final AtomicInteger waitingOn = new AtomicInteger();
 
-        LocalWait(Node node, Id replyToNode, long replyToMessage)
+        LocalWait(Node node, Id replyToNode, TxnId txnId, long replyToMessage)
         {
             this.node = node;
             this.replyToNode = replyToNode;
+            this.txnId = txnId;
             this.replyToMessage = replyToMessage;
         }
```
belliottsmith, line 31: Should `onChange` no longer be synchronised?

accord-core/src/main/java/accord/coordinate/Agree.java
```diff
@@ -45,11 +49,76 @@ public boolean hasMetFastPathCriteria()
...
+        private final boolean fastPathPermitted;
+        private final Set<Id> successes = new HashSet<>();
+        private final Set<Id> failures = new HashSet<>();
```
belliottsmith, line 57: Initialise lazily?

accord-core/src/main/java/accord/coordinate/Agree.java
```diff
@@ -137,13 +232,13 @@ private void onPreAccepted()
...
 
-            startAccept(executeAt, deps);
+            startAccept(executeAt, deps, tracker.topologies());
         }
     }
 
     private boolean shouldSlowPathAccept()
     {
-        return !tracker.hasInFlight() && tracker.hasReachedQuorum();
+        return (!tracker.fastPathPermitted || !tracker.hasInFlight()) && tracker.hasReachedQuorum();
```
belliottsmith, line 241: Perhaps encapsulate some of this in `tracker`?

belliottsmith (changes requested):
Final round of review, mostly looks good just some minor things to look at. I also notice there is some feedback from the last round of review that has not been responded to?

accord-core/src/main/java/accord/messages/BeginRecovery.java
bdeggleston: that was fixed in the topology-integration branch since it indeed turned out to be unsafe

accord-core/src/main/java/accord/local/Node.java
bdeggleston: heh, I prefer the former since I find it easier to read :). Anyway, topology-integration refactoring reduced the number of commands to make it a one liner

bdeggleston:
merged as part of https://github.com/apache/cassandra-accord/pull/2

PR #2, opened 2022-03-01
Initial Cassandra integration

bdeggleston:
I'll rebase onto trunk once https://github.com/apache/cassandra-accord/pull/1 is merged


accord-core/src/main/java/accord/messages/ReplyContext.java
```diff
@@ -0,0 +1,5 @@
+package accord.messages;
+
+public interface ReplyContext
```
belliottsmith, line 3: Could you link to the C* implementation part this maps to? I had sort of imagined it would be a `long` on the C* side as well, given the `RequestCallbacks` uses a long for `messageId`

belliottsmith (approved):
LGTM pretty much, once the prior PR has been finished up. Only major suggestion is that, in retrospect, I think it would be easier to just include a cassandra-all.jar file for `Future` etc for now as a compile-time only dependency, as maintaining a commented-out copy is less than ideal, on top of which there will probably be other things we want in future and bringing in updates will be easier as well. WDYT?

belliottsmith:
I have some suggested improvements to how we manage command stores: https://github.com/belliottsmith/accord/tree/topology-integration-suggest

Also, I think this patch continues to have some issues passing the burn test

accord-core/src/main/java/accord/messages/ReplyContext.java
bdeggleston: here you go: https://github.com/bdeggleston/cassandra/blob/accord-integration/src/java/org/apache/cassandra/service/accord/AccordMessageSink.java#L100

In C*, the reply context is the incoming message. After the netty refactor, the message is used to create the reply, and auto-fills stuff like the response verb, the id of course, and some timeout related stuff iirc

bdeggleston:
I reworked this a bit to prevent assigning ranges to nodes that were previously replicas of said ranges. Supporting that opens a can of worms where you have to purge stale data which other nodes might still need. It's probably tractable today, but it will be much easier once repair is merged and we can require purgeable ranges are fully replicated. I also added a throttle on topology changes since the topology randomizer was keeping some burn tests busy indefinitely all by itself.
Regarding depending on cassandra-all, In principle I don't have a strong preference... both seem kind of janky. In practice, I'd prefer sticking with the copy paste until it becomes a problem, just because it's done. I spent about 20 minutes trying to get the cassandra-all approach to work locally and the compiler couldn't find any of the classes we're using. I fiddled around with it a bit, but didn't want to spend too much time on it.
Finally, about suggestions regarding command stores. Let's revisit this when the metadata persistence patch is ready for review. A lot of what you're suggesting is already done in that patch in one form or another.

belliottsmith:
I think it would be great to pause and let review and feedback catch up, even if there may be some duplicated work or rebasing involved. We haven't yet addressed all feedback from the last PR, and to daisy-chain to a third PR would be less than ideal IMO.

Regarding cassandra-all, there are already C* concepts that I think we would benefit from having access to, but sure I can bring it in when I need it.

bdeggleston:
I guess that's fair. It looks like this addresses more things introduced in this patch that I'd realized initially. Would you mind rebasing it onto this branch? Could we also leave the various CommandStore[s] changes in nested classes? This has been reworked for metadata persistence and that would simplify rebasing.

belliottsmith:
Yep, rebased and pushed.

I think there remain some unaddressed comments from the last PR it would be great to close-out, perhaps we can address them in this PR and close that one so we can get to a single point to cleanly move forwards from?

I cannot find it now, also, but I remember commenting additionally about the `TxnRequest.Scope` and whether we actually need to send every `KeysForEpoch` or just the superset of `Keys` and the range of epochs we cross?

bdeggleston:
Ok pulled in the suggestions and addressed the missed feedback. For the suggestions, I reverted how command stores detect reintroduced ranges, as the changes would not have detected them properly. Also changed some minor stuff for clarity

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -228,7 +228,7 @@ public void processBlocking(Consumer<? super CommandStore> consumer)
         }
         catch (InterruptedException e)
         {
-            Thread.currentThread().interrupt();
+            throw new AssertionError(e);
```
belliottsmith: would be nice to have `UncheckedInterruptedException` here...

Humbedooh:
testing gitbox notification target, please ignore this

belliottsmith:
I think we're pretty much there, just a couple of things to talk about. 

1) it would be great to consider my proposed simplification to `TxnRequest.Scope`
2) we should probably decide whether to call `foldl` accumulate or `accumulate` foldl. Basically they're the same, `foldl` is just the functional name for the construction (and it communicates in what order the accumulation occurs). accumulate is easier for people to grok in many ways, but is less communicative and probably less standard, so I opted for `Fold` when I introduced similar methods this time around, but happy to migrate back. We should probably pick one or the other though.

bdeggleston:
1. can you point me to the `TxnRequest.Scope` proposal? I don't remember seeing it and couldn't find it in either PRs or JIRAs. Fully expect it to be somewhere obvious though :)
2. let's go with foldl

belliottsmith:
> I cannot find it now, also, but I remember commenting additionally about the `TxnRequest.Scope` and whether we actually need to send every `KeysForEpoch` or just the superset of `Keys` and the range of epochs we cross?

Basically I think we pretty much only ever _functionally_ use the union of all `Keys` - we do use the individual ones for a kind of sanity check, but I think the union would be fine for this too. It would probably simplify a bunch of things, including serialisation. WDYT?

belliottsmith:
Also I realise it was unclear in retrospect but I was mostly lamenting we couldn't easily use `UncheckedInterruptedException` - but thanks for copy pasting it so we can now!

bdeggleston:
> Basically I think we pretty much only ever functionally use the union of all Keys 

One place that would be an issue is when a node reacquires ranges it previously replicated, then lost, and how that will interact with the epoch checks being permissive with unknown epochs.

For instance, with a (current) scope:

```
1: [a, b]
2: [a, b, c]
3: [a, b, c]
```

If the node is aware of epoch 2, but not 3, the check can infer that it’s not being asked to work on ranges it doesn’t already replicate and pass the check.

On the other hand, it can infer from this scope that it’s missing something if it’s not aware of epoch 3 yet:

```
1: [a, b, c]
2: [a, b]
3: [a, b, c]
```

Transmitting just a min and max epoch, and a union of keys, the check wouldn’t be able to distinguish between these 2 states, and would have to pessimistically assume the second scenario, though it’s much less common than the first.

bdeggleston:
We might be able to address this by diffing keys per epoch when the scope is being built, and communicating the max epoch as the highest one where the keys being replicated changed instead of the actual max epoch we're aware of. I'll try that out tomorrow.

accord-core/src/main/java/accord/messages/TxnRequest.java
```diff
@@ -33,113 +29,51 @@ public Scope scope()
...
             {
                 Topology topology = topologies.get(i);
-                KeysForEpoch keysForEpoch = KeysForEpoch.forTopology(topology, node, keys);
-                if (keysForEpoch != null)
+                KeyRanges topologyRanges = topology.rangesForNode(node);
+                if (topologyRanges == null)
+                    continue;
+                topologyRanges = topologyRanges.intersection(txnKeys);
+                Keys epochKeys = txnKeys.intersection(topologyRanges);
+                if (lastKeys == null || !lastKeys.equals(epochKeys))
```
belliottsmith: could this be `lastKeys.containsAll(epochKeys)`?

belliottsmith:
LGTM besides that minor suggestion / question. Let's get this merged.

PR #5, opened 2022-06-28
CASSANDRA-17718: Transaction Invalidation

belliottsmith:


bdeggleston:
Is there a reason recovery shouldn’t invalidate any operations it can’t determine were already accepted/committed, ie: `acceptOrCommit==null` in Recovery#recover? Presumably recovery is being run because we believe the coordinator has failed and the client would therefore not be expecting their operation to have been completed.

bdeggleston:
If not, I think we could simply set an `invalidated` flag on the accept message we send out in the NotWitnessed/Preaccepted case. We might also be able to just interpret a non-zero accepted ballot as having invalidated a transaction if we adjust how we deal with previously accepted operations in recover. We could dispense with most/all of the parallel invalidating proposal/commit stuff and simplify the progress log state in that case.

belliottsmith:
Fast path decisions aren’t necessarily recorded in this way, but may have returned to the client already - we may have triggered recovery erroneously, or the coordinator may have since recovered from a temporary fault, or the client may have simply received a success result without any execution information prior to the coordinator failing.

bdeggleston:
Sure, but in that case we should either be able to detect that a fast path decision could have been taken, or see at least one accept message from the coordinator, in which case we would not invalidate

bdeggleston:
And regarding erroneous recovery or temporary coordinator faults, if the coordinator couldn't have taken the fast path, I believe it's doomed to timeout anyway since it won't be able to make any progress once recovery has taken over.

accord-core/src/main/java/accord/coordinate/Coordinate.java
```diff
@@ -1,46 +1,283 @@
...
+    }
+
+    final Node node;
+    final TxnId txnId;
+    final Txn txn;
+    final Key homeKey;
+
+    private PreacceptTracker tracker;
+    private final List<PreAcceptOk> preAcceptOks = new ArrayList<>();
+    private boolean isPreAccepted;
```
bdeggleston: If we decide to not separate preaccept from coordinate: `isDone` might be too generic for the Coordinate class, but `isPreaccepted` implies (to me) that preaccept was successful, and is odd to see set to true in `onFailure`. Maybe `preacceptDone` or something similar?

accord-core/src/main/java/accord/coordinate/Coordinate.java
```diff
@@ -1,46 +1,283 @@
...
+
+import org.apache.cassandra.utils.concurrent.AsyncFuture;
 import org.apache.cassandra.utils.concurrent.Future;
 
-public class Coordinate
+/**
+ * Perform initial rounds of PreAccept and Accept until we have reached agreement about when we should execute.
+ * If we are preempted by a recovery coordinator, we abort and let them complete (and notify us about the execution result)
+ */
+public class Coordinate extends AsyncFuture<Result> implements Callback<PreAcceptReply>, BiConsumer<Result, Throwable>
```
bdeggleston, line 34: WDYT about pulling the preaccept phase into its own class and composing preaccept and other phases in the coordinate class? On one hand, the top level operations, coordinate, recover, etc each have a unique first stage, so I understand the reasoning for bundling them into their respective classes. OTOH, each operation is a composition of several phases, and it would make more sense to me if they were modeled that way.

accord-core/src/main/java/accord/impl/SimpleProgressLog.java
```diff
@@ -129,6 +131,7 @@ boolean isAtLeast(GlobalStatus equalOrGreaterThan)
         Set<Id> globalNotPersisted;
         GlobalPendingDurable globalPendingDurable;
 
+        Object debugInvestigating;
```
bdeggleston, line 134: I'm assuming this can be removed?

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -229,6 +233,7 @@ static void burn(Random random, TopologyFactory topologyFactory, List<Id> client
     public static void main(String[] args) throws Exception
     {
         Long overrideSeed = null;
+//        Long overrideSeed = -7320078316311161123L;
```
bdeggleston, line 236: leftover seed

bdeggleston (changes requested):
No major objections to anything here. Invalidation logic is pretty straightforward. I do like wrapping up the check for epoch first pattern into `node.withEpoch`. 
One thing that should be cleaned up by the time the partial replication patch is merged is the number of `TODO (now)` comments that were added here. Some were addressed in that patch (and then others added), so I haven't paid too much attention to them here, but there are still quite a few in that branch.
Finally, I know this is easier said than done, but keeping refactors and functional changes in separate commits would really help speed up review by limiting the scope of each sub-review. I realize they usually happen during implementation, but something for the wish list at least :)

accord-core/src/main/java/accord/coordinate/Coordinate.java
belliottsmith: Yeah, I'd be happy simply renaming `Coordinate` to `PreAcceptPhase` (can't be `PreAccept` as we have that already), as that's really all `Coordinate` is right now (plus an `AsyncFuture`, which could be allocated separately if we wanted). Each phase just invokes the next and passes on the callback now.

FWIW, I also think it's fine as it is, as `PreAccept` can only be started by the initial coordinator, and we do allocate slightly less by bundling this into the same class. But I don't really have a strong preference for either, so happy to take yours.

accord-core/src/main/java/accord/impl/SimpleProgressLog.java
belliottsmith: I'd prefer to keep it, it can be helpful for debugging issues thrown up by the burn test. This isn't going to be a production implementation after all.

accord-core/src/test/java/accord/burn/BurnTest.java
belliottsmith: I leave some seed commented out to make it easier to switch. Not sure if it matters much at this point in development, but perhaps I should make it cleaner going forwards.

belliottsmith:
> One thing that should be cleaned up by the time the partial replication patch is merged is the number of TODO (now) comments that were added here.

I think we need e.g. `TODO (soon)`. I've triaged and completed the most important `TODO (now)`, leaving ones I don't have time for but still want to get done ahead of many of the other `TODO`. I'll clean up either way, at least.

> Finally, I know this is easier said than done, but keeping refactors and functional changes in separate commits would really help speed up review by limiting the scope of each sub-review. I realize they usually happen during implementation, but something for the wish list at least :)

Yeah, sorry. Time pressure leads to corner-cutting here. It's time consuming to rebase and separate things, but perhaps a false economy if it steals your time. I'll have a look over the next patch sets to see what I can do.

bdeggleston:
> I think we need e.g. TODO (soon)
That sounds good, some sort of TODO prioritization.

> Yeah, sorry. Time pressure leads to corner-cutting here. It's time consuming to rebase and separate things, but perhaps a false economy if it steals your time. I'll have a look over the next patch sets to see what I can do.
No worries, it's hard to know which is the better use of time as well. Happy medium is probably different commits where splitting out is easy, and a talk before review to go over the major points.

belliottsmith:
I've made the requested (minor) changes, let me know if you think they're OK and I'll get to merging this. Thanks.

bdeggleston (approved):
Point taken about Coordinate/Preaccept, there's not much left :). I reverted the split, renamed the tracker test and renamed isPreaccepted.

PR #8, opened 2022-08-02
CASSANDRA-17796: Add CircleCI configs to build the project

dcapwell:


accord-maelstrom/src/main/java/accord/maelstrom/Packet.java
```diff
@@ -16,33 +17,49 @@
...
+        CheckStatus(CheckStatus.class),
+        CheckStatusOk(CheckStatus.CheckStatusOk.class),
+        CheckStatusOkFull(CheckStatus.CheckStatusOkFull.class),
+        InformOfTxn(InformOfTxn.class),
+        InformOfTxnOk(InformOfTxn.InformOfTxnOk.class),
+        InformOfPersistence(InformOfPersistence.class),
+        SimpleProgressLog_ApplyAndCheck(SimpleProgressLog.ApplyAndCheck.class),
+        SimpleProgressLog_ApplyAndCheckOk(SimpleProgressLog.ApplyAndCheckOk.class);
+
+        private static final Map<Class<?>, Type> LOOKUP = Arrays.stream(Type.values())
```
belliottsmith: perhaps `LOOKUP_MAP`, so we can call the exported property `LOOKUP`?

belliottsmith: Or else could have a static method to transform the prior function construction to introduce the null check

dcapwell: switched to `LOOKUP_MAP`

dcapwell: > Or else could have a static method to transform the prior function construction to introduce the null check

I did both.  Looked and saw this `Function` is only used within this class, so made it a static function rather than `Function` object where we call `.apply`

accord-maelstrom/src/main/java/accord/maelstrom/Body.java
```diff
@@ -23,7 +24,7 @@
 
     public Body(Type type, long msg_id, long in_reply_to)
     {
-        this.type = type;
+        this.type = Objects.requireNonNull(type);
```
dcapwell, line 27: no longer needed, was used to find where this was happening; can revert if desired

accord-maelstrom/src/main/java/accord/maelstrom/Cluster.java
```diff
@@ -163,15 +163,17 @@ public boolean processPending()
                     }
                     err.println(clock++ + " RECV " + deliver);
                     err.flush();
-                    if (deliver.body.in_reply_to > Body.SENTINEL_MSG_ID)
+                    Object body = ((Wrapper)deliver.body).body;
+                    //TODO : REVIEW : Don't know the point of deliver.body.in_reply_to > Body.SENTINEL_MSG_ID, but without it InformOfTxnReply fails
+                    if (deliver.body.in_reply_to > Body.SENTINEL_MSG_ID || body instanceof Reply)
```
dcapwell, line 169: Don't know the point of `deliver.body.in_reply_to > Body.SENTINEL_MSG_ID`, but without it InformOfTxnReply fails

accord-maelstrom/src/main/java/accord/maelstrom/Body.java
belliottsmith: It's not Guava so I don't mind, though I prefer annotations (and we can perhaps later use byte weaving or something to insert run-time asserts during tests)

belliottsmith: It _looks_ like Guava so I did originally mind a little :)

accord-maelstrom/src/main/java/accord/maelstrom/Cluster.java
dcapwell: leaving as is, moved TODO to a comment documenting the RHS of the else

dcapwell:
@belliottsmith @bdeggleston let me know what you think, if its +1 or more feedback needed

belliottsmith (approved):
Sorry David, it looks fine to me. +1.

PR #10, opened 2022-09-13
Metadata persistence

bdeggleston:


accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -356,42 +315,110 @@ public synchronized void shutdown()
...
 
-    public <T> T mapReduce(Keys keys, long epoch, Function<CommandStore, T> map, BiFunction<T, T, T> reduce)
+    private static Fold<TxnOperation, Void, List<Future<Void>>> forEachFold(Consumer<CommandStore> forEach)
     {
-        return mapReduce(keys, epoch, epoch, map, reduce);
+        return (store, op, i, t) -> { t.add(store.process(op, forEach)); return t; };
     }
 
-    public <T> T mapReduce(Keys keys, long minEpoch, long maxEpoch, Function<CommandStore, T> map, BiFunction<T, T, T> reduce)
+    public void forEach(TxnOperation operation, Keys keys, long minEpoch, long maxEpoch, Consumer<CommandStore> forEach)
```
dcapwell: "forEach" takes both a "TxnOperation" and "Keys", but "TxnOperation" holds ref to "Keys", so shouldn't we drop "Keys"? Looking at the usage, it seems like we could...

```
node.forEachLocalSince(TxnOperation.scopeFor(txnId), full.txn.keys(), ...
```

```
TxnOperation scope = TxnOperation.scopeFor(txnId, max.txn.keys());
...
node.forEachLocalSince(scope, max.txn.keys(), ...
```

This change trickled throughout a lot of different places, so would be good to grasp the expected semantics... if the TxnOperation has keys, and you provide keys... "should" they be released or "must" they be related? or something else?

dcapwell: Reply from Blake in JIRA

The keys in each perform different functions. The keys in the TxnOperation indicate which CommandsForKey we need access to, this let's C* know which ones it needs to load. The Keys passed into forEach determine which command stores we need to talk to, since we need to talk to "replicas" of the given keys, whether we need the commands for key to be in memory or not. Of course, this isn't obvious without more complete comments on TxnOperation

> This change trickled throughout a lot of different places, so would be good to grasp the expected semantics... if the TxnOperation has keys, and you provide keys... "should" they be released or "must" they be related? or something else?

I added some comments. The name TxnOperation is a little unclear. Let me know if you have any suggestions for better names. Maybe TxnOperationScope? Maybe it could extend Function<CommandStore, T> and bundle the operation and the metadata into the same interface? Calls would be a bit more cumbersome, but would make the interdependency of the 3 parts more obvious

accord-core/src/main/java/accord/coordinate/Execute.java
```diff
@@ -70,6 +71,12 @@ private void start()
     {
         Set<Id> readSet = readTracker.computeMinimalReadSetAndMarkInflight();
         Commit.commitAndRead(node, topologies, txnId, txn, homeKey, executeAt, deps, readSet, this);
+        // skip straight to persistence if there's no read
+        if (txn.read().keys().isEmpty())
```
dcapwell, line 75: We are not good with nullability in Accord, so its not clear if read is allowed to be null... Based off Caleb's branch org.apache.cassandra.cql3.statements.TransactionStatement#createRead won't be null in a write only txn, but is null allowed or not? If so then we need a null check at "if (txn.read().keys().isEmpty())"

dcapwell: Reply from Blake in JIRA

I've added some annotations based on how things are implemented now

accord-core/src/main/java/accord/impl/InMemoryCommand.java
```diff
@@ -0,0 +1,314 @@
...
+    private final Listeners listeners = new Listeners();
+
+    public InMemoryCommand(CommandStore commandStore, TxnId txnId)
+    {
+        this.commandStore = commandStore;
+        this.txnId = txnId;
+    }
+
+    @Override
+    public boolean equals(Object o)
```
dcapwell, line 65: eq is missing homeKey, progressKey, isGloballyPersistent, is this intentional?

dcapwell: This was fixed

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    @Override
+    public CommandsForKey maybeCommandsForKey(Key key)
+    {
+        return commandsForKey.get(key);
+    }
+
+    // TODO: command store api will need to support something like this for repair/streaming
+    public void forEpochCommands(KeyRanges ranges, long epoch, Consumer<Command> consumer)
```
dcapwell, line 98: I still need to look closer at our epoch logic but if we have multiple epochs that are not full acknowledged then wouldn't forEpochCommands have the wrong min epoch? This is only used in tests and not in the C* patch... if this is only for tests then I am cool with this, but it would be good to start annotating what methods are not meant to be used in `src/main`

Same comment for forCommittedInEpoch

dcapwell: Reply from Blake in JIRA

No I don't think so, these 2 commands just return witnessed/committed commands belonging to the given epoch. These (and all InMemory* implementations) are only meant to be used in tests. These methods in particular are placeholders to range movements in the library can be validated and should be superseded by the work to "officially" support bootstrap, low bounds, etc.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -0,0 +1,371 @@
...
+                                                                           range.endInclusive()).values();
+            for (CommandsForKey commands : rangeCommands)
+            {
+                commands.forWitnessed(minTimestamp, maxTimestamp, consumer);
+            }
+        }
+    }
+
+    // TODO: command store api will need to support something like this for repair/streaming
+    public void forCommittedInEpoch(KeyRanges ranges, long epoch, Consumer<Command> consumer)
```
dcapwell, line 115: Nit: forCommittedInEpoch and forEpochCommands are mostly copy/paste of each other, can we create a new util function such as

```
public void forEach(KeyRanges ranges, long epoch, Consumer<CommandsForKey> consumer)
    {
        Timestamp minTimestamp = new Timestamp(epoch, Long.MIN_VALUE, Integer.MIN_VALUE, Node.Id.NONE);
        Timestamp maxTimestamp = new Timestamp(epoch, Long.MAX_VALUE, Integer.MAX_VALUE, Node.Id.MAX);
        for (KeyRange range : ranges)
        {
            Iterable<CommandsForKey> rangeCommands = commandsForKey.subMap(range.start(),
                                                                           range.startInclusive(),
                                                                           range.end(),
                                                                           range.endInclusive()).values();
            rangeCommands.forEach(consumer);
        }
    }
```

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -0,0 +1,371 @@
...
+            {
+
+                Collection<Command> committed = commands.committedByExecuteAt()
+                        .between(minTimestamp, maxTimestamp).collect(Collectors.toList());
+                committed.forEach(consumer);
+            }
+        }
+    }
+
+    protected void processInternal(Consumer<? super CommandStore> consumer, Promise<Void> promise)
```
dcapwell, line 135: nit: "protected void processInternal(Consumer<? super CommandStore> consumer, Promise<Void> promise)" could be the following to avoid copy/paste

```
protected void processInternal(Consumer<? super CommandStore> consumer, Promise<Void> promise)
    {
        processInternal(cs -> {
            consumer.accept(cs);
            return null;
        }, promise);
    }
```

dcapwell: resolved

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -0,0 +1,371 @@
...
+            T result = function.apply(this);
+            promise.setSuccess(result);
+        }
+        catch (Throwable e)
+        {
+            promise.tryFailure(e);
+        }
+    }
+
+    public static class Synchronized extends InMemoryCommandStore
```
dcapwell, line 156: Synchronized doesn't use "synchronized" in every method... so name feels off...

dcapwell: resolved

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -0,0 +1,371 @@
...
+                            Node.Id nodeId,
+                            Function<Timestamp, Timestamp> uniqueNow,
+                            LongSupplier currentEpoch,
+                            Agent agent,
+                            DataStore store,
+                            ProgressLog.Factory progressLogFactory,
+                            RangesForEpoch rangesForEpoch)
+        {
+            super(generation, index, numShards, uniqueNow, currentEpoch, agent, store, progressLogFactory, rangesForEpoch);
+            executor = Executors.newSingleThreadExecutor(r -> {
```
dcapwell, line 255: SingleThread should expose ability to control thread name... in C* we have a "global prefix" for jvm-dtest

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -0,0 +1,371 @@
...
+            Thread expected;
+            while (true)
+            {
+                expected = expectedThread.get();
+                if (expected != null)
+                    break;
+                expectedThread.compareAndSet(null, Thread.currentThread());
+            }
+            if (expected != current)
+                throw new IllegalStateException(String.format("Command store called from the wrong thread. Expected %s, got %s", expected, current));
```
dcapwell, line 331: nit: can we add a useful message? When something like this fails its too hard to figure out what's going on, so we should be nice and provide a useful error.

dcapwell: Resolved

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
dcapwell: Reply from Blake in JIRA

I'd tried that, but you also need to pass in the min and max timestamps, which makes 3 arguments so you need a special interface, and the duplicated code isn't too complex, so I figured it was a net negative for readability in exchange for a few less lines of code

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
dcapwell: Reply from Blake in JIRA

This is just for testing, I don't think we need anything more sophisticated here unless it's serving an immediate purpose.

accord-core/src/main/java/accord/impl/InMemoryCommandStores.java
```diff
@@ -0,0 +1,110 @@
...
+    {
+        super(num, node, agent, store, progressLogFactory);
+    }
+
+    public static InMemoryCommandStores inMemory(Node node)
+    {
+        return (InMemoryCommandStores) node.commandStores();
+    }
+
+    public void forEachLocal(Consumer<? super CommandStore> forEach)
```
dcapwell, line 46: It would be nice to move "forEachLocal" to lower level... seems that we try to overload fold in cases where forEach would be simpler.

dcapwell: Reply from Blake in JIRA

these are only there for testing and assume in-memory implementation details (since none of them require a TxnScope)

accord-core/src/main/java/accord/impl/InMemoryCommandsForKey.java
```diff
@@ -0,0 +1,136 @@
...
+        private final NavigableMap<Timestamp, Command> commands = new TreeMap<>();
+
+        @Override
+        public Command get(Timestamp timestamp)
+        {
+            return commands.get(timestamp);
+        }
+
+        @Override
+        public void add(Timestamp timestamp, Command command)
```
dcapwell, line 44: should check to see if a conflict was detected

dcapwell: Resolved

accord-core/src/main/java/accord/impl/SimpleProgressLog.java
```diff
@@ -812,18 +812,26 @@ public void onCallbackFailure(Id from, Throwable failure)
 
     public static class ApplyAndCheck extends Apply
     {
-        final Set<Id> notPersisted;
+        public final Set<Id> notPersisted;
```
dcapwell, line 824: For consistency shouldn't notPersisted be private with a getter? this patch switches many things to getters yet makes this public field access.

dcapwell: Reply from Blake in JIRA

The integration of the progress log hasn't started yet. I'd expect the interface and reference implementation to be refined as that happens, but the objective for this patch to do the bare minimum to prevent build issues and test failures.

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -18,129 +18,117 @@
...
-import static accord.local.Status.Committed;
-import static accord.local.Status.Executed;
-import static accord.local.Status.Invalidated;
-import static accord.local.Status.NotWitnessed;
-import static accord.local.Status.PreAccepted;
-import static accord.local.Status.ReadyToExecute;
-
-// TODO: this needs to be backed by persistent storage
-public class Command implements Listener, Consumer<Listener>
+public abstract class Command implements Listener, Consumer<Listener>, TxnOperation
```
dcapwell: nit: given the refactor to make all state abstract wouldn't it be best if this became an Interface?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -274,99 +289,110 @@ public boolean commit(Txn txn, Key homeKey, Key progressKey, Timestamp executeAt
...
+        commandStore().progressLog().invalidate(txnId(), isProgressShard, isProgressShard && progressKey().equals(homeKey()));
+        logger.trace("{}: committed invalidated", txnId());
 
-        listeners.forEach(this);
+        notifyListeners();
         return true;
     }
 
-    public boolean apply(Txn txn, Key homeKey, Key progressKey, Timestamp executeAt, Deps deps, Writes writes, Result result)
+    public Future<Void> apply(Txn txn, Key homeKey, Key progressKey, Timestamp executeAt, Deps deps, Writes writes, Result result)
```
dcapwell, line 325: apply now returns a Future but some code paths ignore the output, so if apply fails or is async, there may be unexpected results? Examples: accord.coordinate.CheckOnCommitted#onSuccessCriteriaOrExhaustion(accord.messages.CheckStatus.CheckStatusOkFull), and accord.impl.SimpleProgressLog.ApplyAndCheck#process


If apply is still being "applied" then the call to "commit" may cause a race condition as it could be called while the writes are happening, leading to the following timeline:

```
T1 apply -> sets status to "Executed", send writes to be written in the background
T2 commit -> set status to "Committed"
T3 apply complete, call "postApply" -> set status "Applied"
```

specifically this comment applies to "accord.coordinate.CheckOnCommitted#onSuccessCriteriaOrExhaustion(accord.messages.CheckStatus.CheckStatusOkFull)"... I might not be grasping the "epoch - 1" part

accord-core/src/main/java/accord/local/Command.java
dcapwell: Reply from Blake in JIRA

I don't think so. There's a lot of functionality in there, some of which needs to be overridden in the C* implementation

accord-core/src/main/java/accord/local/Command.java
dcapwell: Reply from Blake in JIRA

So I don't think your example timeline is possible, since Command methods like commit are idempotent and the command store is single threaded. T2 would be a noop. However, I don't think there's anything to prevent maybeExecute to fire off duplicate writes in the time between Command#apply dispatches writes and the time Command#postApply sets the status to Applied in Command#postApply., and the same goes for Command#read. I'm going to think about this a bit, since it depends on the concurrency of the implementation I'm not sure if it would be better to handle this in the library, or in the implementation....

accord-core/src/main/java/accord/messages/Apply.java
```diff
@@ -54,15 +63,59 @@ public Apply(Node.Id to, Topologies topologies, TxnId txnId, Txn txn, Key homeKe
...
+        this.txnId = txnId;
+        this.txn = txn;
+        this.homeKey = homeKey;
+        this.executeAt = executeAt;
+        this.deps = deps;
+        this.writes = writes;
+        this.result = result;
+    }
+
+    static Future<Void> waitAndReduce(Future<Void> left, Future<Void> right)
```
dcapwell: nit: "waitAndReduce" throws on future error rather than return a failed future, this feels off to me but looks to be because "mapReduceLocal" doesn't allow changing the return type; we want Void in this context but that API doesn't allow... would it be better to allow a different return type? Caller never checks return so Future<?> is brittle...

dcapwell: Reply from Blake in JIRA

Good point. Changed return type to Future<Void>

accord-core/build.gradle
```diff
@@ -29,11 +29,13 @@ java {
   withSourcesJar()
 }
 
+compileJava {
```
dcapwell: can you move the 1.8 logic to buildSrc/src/main/groovy/accord.java-conventions.gradle

accord-maelstrom/build.gradle
```diff
@@ -20,8 +20,13 @@ plugins {
     id 'accord.java-conventions'
 }
 
+compileJava {
```
dcapwell: can you move the 1.8 logic to buildSrc/src/main/groovy/accord.java-conventions.gradle

dcapwell (commented):
pushed my comments from JIRA to here to better track.

accord-core/src/main/java/accord/messages/Apply.java
```diff
@@ -54,15 +63,59 @@ public Apply(Node.Id to, Topologies topologies, TxnId txnId, Txn txn, Key homeKe
...
+    {
+        try
+        {
+            if (left != null) left.get();
+            if (right != null) right.get();
+        }
+        catch (InterruptedException e)
+        {
+            throw new UncheckedInterruptedException(e);
```
dcapwell: In testing accord.messages.Apply#waitAndReduce swallows the InterruptedException, causing Instance.shutdown to hang. This should re-add the interrupt to the thread

accord-core/src/main/java/accord/api/Read.java
```diff
@@ -29,5 +35,42 @@
 public interface Read
 {
     Keys keys();
-    Data read(Key key, Timestamp executeAt, DataStore store);
+    Future<Data> read(Key key, CommandStore commandStore, Timestamp executeAt, DataStore store);
+
+    class ReadFuture extends AsyncPromise<Data> implements BiConsumer<Data, Throwable>
```
belliottsmith: Should this extend `SyncPromise` since it appears to be synchronous? Or perhaps `AbstractFuture` with no synchronisation (since we should probably be invoking callbacks on the home shard `CommandStore`)

I think it would also be preferable not to have this defined in the interface.

accord-core/src/main/java/accord/api/Write.java
```diff
@@ -27,5 +30,6 @@
  */
 public interface Write
 {
-    void apply(Key key, Timestamp executeAt, DataStore store);
+    Future<Void> SUCCESS = ImmediateFuture.success(null);
```
belliottsmith: Again, not sure it is a great idea to put this in the interface, this case more so as it potentially breaks the simulator in future by pulling in implementation classes alongside interface classes.

accord-core/src/main/java/accord/messages/Apply.java
belliottsmith: Are we sure this is the best way to do this? We're occupying a thread waiting for these results. My preference would be to supply a callback to the apply method, as this is also easier for me to refactor to enrich with extra parameters explaining the outcome in my partial replication patch, but alternatively listeners would suffice?

If I don't hear anything I will go ahead with refactoring to callbacks in my follow-up work.

belliottsmith: In fact, in this particular case, do we even need the execute outcome? If the apply is successful we can respond immediately to the peer?

accord-core/src/main/java/accord/local/Command.java
belliottsmith: It seems totally reasonable to introduce a new status `Applying` or something, that avoids duplicated work.

accord-core/src/main/java/accord/local/CommandStores.java
belliottsmith: I think we should rename `TxnOperation` to something like `PreLoadContext` to indicate its purpose, which is to ensure the necessary state is ready for the operation before it executes.

accord-core/src/main/java/accord/primitives/Deps.java
```diff
@@ -388,7 +390,8 @@ public static <T> Deps merge(List<T> merge, Function<T, Deps> getter)
     // Lazy loaded in ensureTxnIdToKey()
     int[] txnIdToKey; // TxnId -> [Key]
 
-    Deps(Keys keys, TxnId[] txnIds, int[] keyToTxnId)
+    @VisibleForImplementation
```
belliottsmith: Why is this being made public? Can we use the static constructor methods that enforce the constraints?

accord-core/src/main/java/accord/primitives/KeyRanges.java
```diff
@@ -27,22 +28,24 @@
...
+import static accord.utils.Utils.toArray;
 
 public class KeyRanges implements Iterable<KeyRange>
 {
     public static final KeyRanges EMPTY = ofSortedAndDeoverlappedUnchecked(new KeyRange[0]);
 
     final KeyRange[] ranges;
 
-    private KeyRanges(KeyRange[] ranges)
+    @VisibleForImplementation
```
belliottsmith: Why is this being made public? Can we use the static constructor methods that enforce the constraints?

accord-core/src/main/java/accord/primitives/Deps.java
```diff
@@ -1022,15 +1049,27 @@ public TxnId txnId(int i)
...
     }
 
+    @VisibleForImplementation
+    public int keyToTxnIdCount()
+    {
+        return keyToTxnId.length;
+    }
+
+    @VisibleForImplementation
+    public int keyToTxnId(int idx)
```
belliottsmith: Perhaps prefix with `unsafe` or `serialization` as this is only for serialisation, and is not to be interpreted directly? As named it might be misinterpreted as a function for external use, which would be unlikely to go well.

accord-core/src/main/java/accord/messages/Apply.java
bdeggleston: Good point about blocking the request thread, I'll fix that. Regarding callbacks vs futures (here and returned from Read#read), I usually prefer futures since returning them is enforced by the compiler, whereas forgetting to call a callback may not be obvious until runtime. Command#maybeExecute is a good example of a method where it’s easier to make mistakes with callbacks that the compiler won’t let you make with futures. I also find it a bit easier to reason about which method needs to worry about completing a future vs callbacks. Why do you prefer callbacks for the read and write methods? Is it because you can require an arbitrary number of arguments to the callback?

Also, in this case we don't need the execute outcome (and it's Void) we just need to know when we're done so we can respond to the coordinator.

belliottsmith: I don’t see how Futures provide any additional compiler enforcement? The future here is simply a carrier for a callback, since we must not block the thread - if anything we have demonstrated a very easy way we can misuse the Future API, that isn’t possible with a callback. The callback can be required as a parameter, so it is enforced to be provided the same as a Future must be returned. The Future itself must be completed with a parameter and this is just as weakly enforced (in fact identical in requirements) to invoking a callback with the parameters, so I don’t understand the perceived benefit here.

A Future is however:
1. Harder to debug, especially when chaining several together as the callstack becomes very polluted
2. Much less efficient (in both allocations and in cost to maintain and invoke)
3. Much less flexible - as mentioned, I cannot readily add any additional parameters. We have to introduce an additional layer of indirection to carry additional information - and additional allocations
4. Most of the time - including in this instance - just serves as an additional layer of indirection to implement a callback.

I don’t think Future should never be used, but I think it’s use should be minimised to cases where it’s really needed, or to APIs where we need to provide greater flexibility to consumers.

bdeggleston: The compiler helps in the `maybeExecute` scenario I was talking about, where there are cases where the method bails out before the main event. For example:

```java
// it's easy to forget to invoke the callback here
if (shouldBailOut)
{
    callback.accept(null, null);
    return;
}

// but you have to return something here
if (shouldBailOut)
    return ImmediateFuture.success(null);
```

Of course, you have to remember to invoke the callback on the "main event" of execution as you do the future, although oftentimes things executor provided wrappers will do that for you, but I do believe the return signature does help prevent mistakes like the one above. 

Also, to be fair, it's possible to abuse the a callback in the same way as it was abused in this instance, by blocking the thread while waiting for it to be invoked.

In the case of the apply method here, in C* this terminates in a call submitting a mutation to the mutations stage, which returns a future. This future is then cached so we can de-dupe multiple apply requests for a given command (same on the read side). This _could_ be achieved by manually managing callbacks, but it will be more complicated and will end up reimplementing a lot of the future callback logic.

> Harder to debug, especially when chaining several together as the callstack becomes very polluted

The callstack does get deeper, but I can usually skip over the future specific stuff without too much trouble

> Much less efficient (in both allocations and in cost to maintain and invoke)

Maybe in raw opcode/byte count and/or some microbenchmarks... but in the context of the larger system, I don’t think ditching futures for callbacks would be a big performance win.
> Much less flexible - as mentioned, I cannot readily add any additional parameters. We have to introduce an additional layer of indirection to carry additional information - and additional allocations

How many extra parameters are we talking about here? The future isn’t returning a value as is, so you get one for free.

> Most of the time - including in this instance - just serves as an additional layer of indirection to implement a callback.

I wouldn’t call them “just an additional layer of indirection”, they’re tools that simplify programmer management of callbacks and their edge cases, albeit at the cost of another layer of indirection.

belliottsmith: > The callstack does get deeper, but I can usually skip over the future specific stuff without too much trouble

Unfortunately, I can't. I find it very challenging to debug nested `Future` callbacks, and I think this should by itself be a strong reason to avoid them except where they provide significant ergonomic advantages. But, if this were the only consideration I would perhaps accept it on taste grounds. Most of the other points fall similarly... But

> Maybe in raw opcode/byte count and/or some microbenchmarks... but in the context of the larger system, I don’t think ditching futures for callbacks would be a big performance win.

This is the main thing I have to disagree with strongly. This kind approach really hamstrings Cassandra's ability to improve. Each Future with callback is an additional three CAS instructions and ~100 bytes of memory. This is a hot path, where we already removed a chain of about four `Future` when internally processing requests. This is the removal of perhaps a dozen CAS instructions and several hundred bytes allocated. This is unlikely to be very measurable _today_, but only because of hundreds of accumulated decisions that deemphasised this kind of thing.

I am in favour of this kind of inefficiency when there's _strong justification_. But when it's at best a toss-up (in my opinion, it is worse), and you can convert a callback to a `Future` _for free_ but can never save the `Future` cost once the API has committed you to it, then I feel strongly that this isn't the right approach.

>In the case of the apply method here, in C* this terminates in a call submitting a mutation to the mutations stage, which returns a future.

This is true today, but it feels like a bit of a tail wagging the dog situation. I would prefer not to base the API on the current implementation. It is not clear this is how we _will_ be doing things (esp. with improved commit log likely preceding query processing, so that we only have to touch memtables).

bdeggleston: > This kind approach really hamstrings Cassandra's ability to improve. 

You mean a detailed and informed discussion of the pros and cons of different solutions? I feel like the project could use more of this ;)

> Each Future with callback is an additional three CAS instructions and ~100 bytes of memory. This is a hot path, where we already removed a chain of about four Future when internally processing requests. This is the removal of perhaps a dozen CAS instructions and several hundred bytes allocated. 

I understand this is a hot path, and prior to pushing back on this, I implemented a callback class that would meet the needs of apply, and the cas/bytes numbers above are way overstated. The way we distribute work to command stores negates a lot of the potential simplicity of callbacks. Specifically, we may have to talk to multiple command stores, and we don’t know how many until we’ve dispatched our tasks.

In the case where we interact with a single command store, futures actually use slightly less memory than callbacks. In the worst case, a future will use 24 more bytes and one more cas operation than a callback. 

Here’s the callback class if you’d like to take a look, it’s possible I’ve overlooked something: https://gist.github.pie.apple.com/beggleston/56e540bdb6408dd7f22ec88d095c42e2

In the more complex use of futures, like this, callback state starts to look a lot like future state.

bdeggleston: To articulate my position here in some more detail, I believe we should avoid practices that are hazardous, or that make it easier for bugs to be introduced in the future. From this perspective, I see callbacks as inferior to futures because they make control flow the responsibility of the programmer in a way that is less obvious, easier to forget about, and more difficult to reason about than futures. 

While futures also make control flow the responsibility of the programmer, they’re typically on a much shorter leash. Often being instantiated and completed in a much narrower scope, like an executor, or in some critical path, where the focus very clearly includes correctly handling control flow via the future.

In the case of apply, we’re talking about instantiating a callback in the message process method, plumbing it through a decent chunk of the command code, and into the storage layer. In my view, this is an unacceptably large surface area to be requiring programmers to be handling an opaque callback correctly.

I agree that the project has historically had a problem with not taking performance seriously, but I think it has had an even bigger problem with not taking safely seriously, and has also suffered more because of it.

I'm not saying performance and safety are mutually exclusive, though there's often a trade off between the two. I do think it’s fine to trade safety for efficiency when the risk is low and the gains are non-trivial, but I feel pretty strongly that that isn’t the case here.

belliottsmith: > You mean a detailed and informed discussion of the pros and cons of different solutions? I feel like the project could use more of this ;)

No, absolutely :)

> In the case where we interact with a single command store, futures actually use slightly less memory than callbacks.

That definitely means we've done callbacks wrong :) ... bear in mind, the `Future` has to allocate a list entry for the callback, potentially allocates a `WaitQueue` for consumers like those above, and it has to perform a CAS to add the callbacks, signal the callbacks, and to set the result (also at least four for each waiting `Thread`). At _most_ a callback should have to do the last of these (and usually not even that).

> Here’s the callback class if you’d like to take a look

This looks like it is trying to implement a map/Reduce chain over a list of callbacks, i.e. the `waitAndReduce` method - rather than a single callback? But we have to solve this anyway still above, as we shouldn't be blocking, I think. So we will need some kind of mechanism for this either way, don't we? I think it can be a lot simpler though.

> Often being instantiated and completed in a much narrower scope, like an executor, or in some critical path

So, literally anywhere you can have a `Future` you can have a callback with exactly the same control flow. There's no need for extra plumbing. Wherever you would invoke `AbstractFuture.setSuccess` or `AbstractFuture.setFailure` you can invoke a callback.

It is true that an `Executor` lets you hand this off, but we can do that for callbacks too. We have the power! `ExecutorPlus` can easily be given a callback `execute` method variation, in fact this seems like a really good idea. But also, a lambda can be submitted that simply has a try/catch branch doing the obvious things.

> I'm not saying performance and safety are mutually exclusive, though there's often a trade off between the two.

If I thought `Future` afforded more safety, I'd definitely agree it is a cost worth paying, so in that respect I can absolutely see your point of view. I just still don't see the extra safety here. 

Perhaps we do as a project need to offer as much scaffolding and support for callbacks as we already do for `Future`. Perhaps we can also consider variants of `Future` that are more restricted but efficient (and have easier call stacks!), so we can begin walking the two extremes closer towards each other as well.

bdeggleston: > That definitely means we've done callbacks wrong :) ... bear in mind, the Future has to allocate a list entry for the callback, potentially allocates a WaitQueue for consumers like those above, and it has to perform a CAS to add the callbacks, signal the callbacks, and to set the result (also at least four for each waiting Thread). At most a callback should have to do the last of these (and usually not even that).

Sure, I know, and I’m not arguing that a simple callback is less efficient than a future. The point I’m trying to make is that you can’t use a simple callback for apply. If you want to use callbacks with apply, avoid locking, have minimal memory allocation, and prevent multiple calls from the same command store from causing problems, you need this extra state and cas operations to do it, and once you’ve added all that, the efficiency argument becomes a lot thinner.

> So, literally anywhere you can have a Future you can have a callback with exactly the same control flow. There's no need for extra plumbing. Wherever you would invoke AbstractFuture.setSuccess or AbstractFuture.setFailure you can invoke a callback.
 
In the places where you’d be calling `setSuccess` or `setFailure`, that’s true. My issue with callbacks applies to everywhere else. Assume that the originator of the future or the destination of the callback will do the right thing, and think of the rest of the program. Since Future encapsulates the future result of an operation you can do one of two things with it, if you need the result, get it, otherwise return it down the stack. 

With a callback, there are more options:
 * If an exception is thrown, do you need to call it? 
 * How do you make sure you don’t call it twice? 
 * What happens if you call it twice? Does it throw an exception, or totally break your program? 

With callbacks, these are now things you need to think about. None of this is intractable, but it creates an unnecessary distraction from whatever consensus thing you’re doing anyway, and with your attention split between consensus and a callback you’re supposed to know what to do with, you’re more likely to make a mistake that you and the reviewer miss.

> Perhaps we can also consider variants of Future that are more restricted but efficient (and have easier call stacks!), so we can begin walking the two extremes closer towards each other as well.

I was going to suggest this as a middle ground, a lot of the inefficiency of our futures classes come from their needing to work properly in all cases

belliottsmith: >  you need this extra state and cas operations to do it, and once you’ve added all that, the efficiency argument becomes a lot thinner.

Except I think all of this extra work needs to be done for `Future` too? I'm pretty sure the correct approach here is simply using the `Future` as a carrier for a callback? So in this case we have two sides of the safety coin: the side we pass to the `Executor` which can be easily encapsulated safely (or done for us by the `Executor` even more cheaply), and the consumer side which I think will be identical, as we need to manage the concurrency of multiple asynchronous results being accumulated in the object that is reducing them, and we need to do it without blocking the thread?

> With a callback, there are more options:

I don't think there are? The only affordance I see here is the `trySuccess` and `tryFailure` methods, but we can readily have a very lightweight class (a single boolean) that a callback can extend (or be passed to as a lambda) that ensures this isn't a problem.

Obviously there's a huge spectrum of possible functionality/performance points, and we have to be selective so as not to overwhelm our lingua franca. I think we might have room for a couple of additional points in the spectrum though: lightweight `Future` objects designed to only support at most one follow-up action (either a caller waiting on the result, or a listener of some kind); and better callback support.

But I think we should always reach for callbacks first, and only upgrade to `Future` where we find it specifically is well justified by the call-site, so that the trade-offs are well considered and we don't become overly dependent on the additional machinery when it isn't needed (the caller I think can always be made equivalent to `Future`, as discussed above). I think by far the majority of use-cases in Cassandra (esp. until we get virtual threads) are best expressed as single submit/consume transactions with an object accumulating one or more outcomes - and callbacks fit this niche perfectly.

I guess we have accidentally trodden on a minor landmine about project code style and direction :)

belliottsmith: I guess I should say, though, that my protestations do become a lot more muted with a more efficient `Future`. I do wonder how efficient we can really go, though, having pondered it a little. We can reduce the memory footprint a little, but it is hard to improve the performance due to multi-threading - and anything we do to improve the multi-threading is likely to involve memory consumption trade-offs.

I have a general preference to use the minimal API required to implement a feature, but a really cheap and ergonomic future might be compelling anyway. Perhaps I will have a play with `AbstractFuture` and see what I can extract if we impose some restrictions such as a single follow-up action.

bdeggleston: > Except I think all of this extra work needs to be done for Future too?

Hmm good point. So that would penalize futures an additional cas instruction when there is more than one future to listen to (assuming we don't use any lighter/efficient future implementations).

> I don't think there are?

There definitely are. The Future interface(s) do not have trySuccess or tryFailure methods, or any other methods for setting a result, you’re thinking of the implementations. Again, I’m not talking about the places responsible for completing an operation, I’m talking about the rest of the program. Futures only let you receive the outcome of an operation, they’re one sided. They separate the responsibility of performing an operation from the ability to consume its result. The lack of any methods for completing the operation clearly communicate to the programmer that it’s not their responsibility, the originator of the future is responsible for doing the right thing. With callbacks, this separation does not exist.

> But I think we should always reach for callbacks first

I agree, but I think we should upgrade to Futures as soon as the callback needs to be passed around to areas of the program not responsible for completing it.

accord-core/src/main/java/accord/primitives/Deps.java
bdeggleston: for serializer support, it doesn't look like there's a comparable static constructor

accord-core/src/main/java/accord/messages/Apply.java
belliottsmith: > So that would penalize futures an additional cas instruction when there is more than one future to listen to

Three, I think? One to set the `Future` result, one to register a listener, and one to notify the listeners. I am playing with a `Future` implementation that could at least downgrade that to a couple of volatile writes in the common case though.

> With callbacks, this separation does not exist.

Could you give me an example, as I'm not seeing it? The callback is explicitly the responsibility _only_ of the producer of the outcome, it's just flipping the interface but otherwise is isomorphic. While the `Future` is passed back to the caller, the callback is only passed to the producer. The main benefit of flipping this is that we declare our receiver before the work begins, so we don't have any race condition where the work may have completed before the next step is registered; and by not maintaining a reference to the outcome we save some additional work and memory.

> I think we should upgrade to Futures as soon as the callback needs to be passed around to areas of the program not responsible for completing it.

I think this is the source of our disconnect. Where would this happen? In my mind, the callback is never passed to such areas - the callback may be a conduit to propagate that value elsewhere, but it should only ever be the glue (propagating it to some other object/recipient), and should never itself escape the scope of the method that declares it...

I agree that there are cases where it is convenient to return some encapsulation of the result. But `Future` is really for synchronous consumers of a result, that want to invoke `get` - as async users really just use it as a conduit for a callback. Some method nesting may make it easier to pass back a `Future` to attach this to rather than passing down a callback, but I think it is probably not all that commonly essential. Most of the time the same chain of `return` types that yields a `Future` can be transformed into an extra parameter that accepts a callback.

I actually think `Future` will become much more useful again when virtual threads are a thing, as synchronous programming is fundamentally easier. We're already avoiding the OG pattern for `Future` for even bigger perf reasons (of thread management). I think in future we'll go back to submitting the work and invoking `get`. But today they're sort of a bodge of asynchronous on-top of a synchronous API, IMO.

bdeggleston: > I think this is the source of our disconnect. Where would this happen? 

In our apply example, a callback would be passed up the entire call stack

```
Apply#process   <-- instantiate
Command#apply(Txn, Key, Key, Timestamp, Deps, Writes, Result)
Command#maybeExecute  <-- maybe call
Command#apply(void)
Command#postApply <-- call
```

So that’s 3 methods it has to go through before it gets to postApply, which is supposed to call it. 2 of those methods (both Command#apply) are never supposed to call it, but can do so erroneously. Then there’s maybeExecute, which is supposed to call it sometimes. maybeExecute is particularly tricky because there are multiple points where you need to call the callback _and_ return, and if you do one and not the other, you’ve introduced a bug.

belliottsmith: In this example, I don't think we need to invoke a callback at all in `postApply`. There's no need for the caller to be informed when `Apply` completes - only execute? There might be some error handling to perform in the case that apply later fails, but why are we informing the caller about apply here?

But, leaving that aside, I don't really see the problem? Each method has its own responsibility - if we need to scope it better, we can create a new callback for the relevant scope, that translates the semantics (as we do for `Future`). But I don't see any additional safety - the `Future` we return is just as prone to being done incorrectly at each step? If anything, to me this just highlights additional efficiencies of callbacks, as we _can_ quite safely pass them down without allocating additional objects.

This is exactly what we do for coordination, we pass through the callback until either we hit an exception or we pass responsibility of the callback onwards.

accord-core/src/main/java/accord/primitives/Deps.java
belliottsmith: Perhaps let's add one prefixed by `unsafe`? At present we ensure we impose our constraints (per @dcapwell's excellent feedback on my earlier patch) and it would be a shame to make it again easy to bypass accidentally.

accord-core/src/main/java/accord/messages/Apply.java
belliottsmith: For reference, here's a patch introducing a minimal future that is limited to one listener and as many waiters as we like, with zero CAS operations in the normal path, and no unnecessary allocations (but still more weight than a callback)*: https://github.com/belliottsmith/cassandra/tree/minimalfuture

I still think these use cases are better with callbacks though!

*need to double check implementation with a clearer head

bdeggleston: > But I don't see any additional safety - the Future we return is just as prone to being done incorrectly at each step?

No, you can't erroneously complete a future at each step like you can with a callback.

bdeggleston: > No, you can't erroneously complete a future at each step like you can with a callback.

This is the safety difference I keep bringing up and I don't feel like it's really being engaged with in this disussion.

The advantages of futures is that they leverage the type system to eliminate a class of bugs that callbacks make possible.

belliottsmith: I am genuinely struggling to see it. What class of erroneous completions are we avoiding? If maybeExecute does the wrong thing, why would it do the right thing with a Future? Usually we aim to defend against erroneous use by external code, not code we have control over - once we distrust ourselves all bets are off, and compiler assistance here seems very weak.

Is it that you think the returning of an object somehow ensures semantic scope, or that composition methods available to Future somehow make it easier to guarantee correct completion?

If it’s that you may unintentionally complete a callback more than once (eg with exceptional completion), this can be guarded against where we perceive a risk of it with a single cheap encapsulating object, as discussed.

if it’s that you are concerned that a higher method (maybeExecute) may accidentally invoke a callback even after it has passed ownership to a lower method (apply), this can also be cheaply guarded against (by eg nulling out the local parameter when you pass control) - but it’s also not very hard to simply verify it by code inspection. This is a very weak class of problem IMO, and does not justify the additional costs or machinery to me.

If it’s guarding against literally doing the wrong thing, I just don’t see how a Future prevents a programmer error. 

Can you give me an example of the kind of erroneous completion you imagine, that a Future would prevent?

belliottsmith: If the issue is that you are worried about _not_ completing, or about ensuring each `return` has to have considered whether it passes responsibility down or invokes itself, we could introduce a special marker type returned by callbacks, so that the caller must either invoke the callback, invoke a method that invokes the callback (and returns its marker), or expressly indicate it has passed off responsibility.

e.g.
```
public class SafeConsumer<T>
{
    enum Safety { INVOKED, DEFERRED }
    Safety accept(T input)
}
```

bdeggleston: You’ve just listed 3 safety issues callbacks have that futures don’t. There are probably a few more if you re-read my older responses. None of these are exactly explosion factories, but they’re also unnecessary in light of the alternative. They require programmer and reviewer attention that futures don’t. Consensus protocols are difficult to understand and work on, and this code will likely be read and poked at by a lot of engineers over the coming years. To my mind, that justifies paying the (imo reasonable) performance tax for using safer concurrency primitives that are easier to reason about.

belliottsmith: I really don’t agree. Performance is also critical and these are *not* where the difficulty lies with distributed consensus.

I also outlined trivial solutions to these concerns?

A higher method doing the wrong thing is also just as possible with Future: you can start some work and just return another Future, for instance. This is identical to invoking the callback erroneously, and I don’t see why one is much more likely than another?

If more work is being done after invoking the lower method, what stops you doing the wrong thing with a Future?

I really couldn’t disagree more here. If you really think this class of problem is a serious issue, we just implement code style requirements that you null out a callback after passing it, and use the type system as suggested above to ensure it must be invoked before returning. A little bit of coding discipline with no cognitive burden, like all good engineering.

What additional class of problem are we unable to prevent here, with these accordances?

This isn’t a small thing either. Like streams, Future are prone to very easily disguising very costly behaviours, even ignoring their inherent costs. Once something is a Future there is a tendency to chain, and this is costly and also something we have to burn mental cycles addressing or considering. 

So the question is: are these really weak concerns, that we seem able to address quite easily, more important than ensuring the performance of the system? And I am emphatically answering no.

I am not opposed to costs for safety, but I see little to no additional safety here.

bdeggleston: Fair enough. I think we need to agree to disagree and find a compromise though. No hard feelings of course, I think our perspectives and concerns are both valid. We’re not going to see eye to eye on everything, and having a diversity of perspectives is a good thing.

I’ll look into it some and see if I can find a solution we can both be happy with. Or at least be equally displeased with :)

bdeggleston: So the C* integration actually needs something future like to properly de-dupe multiple calls to apply and read. This was discussed here https://github.com/apache/cassandra-accord/pull/10#discussion_r969953934. I ended up handling it by returning the future from the initial apply/read. This case does require supporting multiple listeners though.

I haven't tried wiring it up, but I think this class would support that use case, and address my concerns with minimal memory or cas operations: https://gist.github.com/bdeggleston/bf791b54946427b8a1f453910e7c318f

bdeggleston: Reviewing the C* integration more, we do need to use an actual future. It hooks into the future handling that cleans up completed futures in the single threaded cache without submitting a task to do so. It does use isDone in addition to needing multiple listeners, but doesn’t need blocking get.

I reworked the previous class into an actual future implementation that’s a bit lighter than the standard future here: https://gist.github.com/bdeggleston/3d083b0baf4d4b810d5c27f7c8de0357

Also, it seems like there’s a need for a class _like_ a partial future but without the blocking parts, which I’ve sketched out an interface for here: https://gist.github.com/bdeggleston/5992d921fe158cd3453c66c60cb3aee3. It provides support for adding callbacks, determining if the operation is done, and for retrieving the result/failure if it is. No thought has gone into the name, AsyncResult or something is probably better.

Our future could easily implement it for interoperability purposes, and I _think_ we could replace much of accord’s Future use with it, since I don’t think there are many places where accord actually needs to block a thread to wait for a result, and we should be thinking twice in places where we do. The places that do need to wait typically don’t have multiple threads waiting on a single result, so that could even be implemented as a special listener. That also applies to the C* accord integration, including the cache needs I mentioned above.

That would save more bytes on heap and cas operations overall than using callbacks in apply, and would address my concerns as well. WDYT?

belliottsmith: Could you link me to the code you're referring to that _requires_ a `Future` in C*? 

This sort of speaks to my concerns, though: once `Future` gets its hooks in, it turns up everywhere, and it is hard to escape. I do not want to bake it into more APIs than necessary.

The paradigm of handling something _once the work is in-flight_ is inherently costlier than a callback that can be declared before work begins, as there is no concurrency to deal with in the latter case.

If we do end up using `Future` did you not see that I had implemented a lightweight future upthread that also reduces allocations? But it permits at most one listener, and expects single-threaded access by the producer and consumer (they may run concurrently though).

> That should save more bytes on heap and cas operations overall than using callbacks in apply

I think it is impossible for a `Future` approach to be lighter weight than a callback approach.

> I think we need to agree to disagree and find a compromise though

That would be ideal, but it might not be possible. Let's certainly try, but conflicting styles going through a library like this, and its integration with Cassandra, can cause problems. Like I've said before, we can translate _to_ a `Future` cheaply, but cannot save the cost once it is there, and `Future` infects APIs and spreads (as you have indicated above). So, if we have to choose, we should opt for the one that imposes its opinions least broadly. If anything, this conversation has made me much more strongly opposed to their use here than I was previously.

belliottsmith: Honestly, the more I think about it, the more I am simply -1 the use of `Future` where it isn't necessary. To some extent there has to be some design consistency, and by baking `Future` in here we are guaranteeing their widespread use forever more. I disagree with foisting this performance trade-off on the hot-path, even if we reduce its cost.

The efficient way to receive a result is to declare the receiver upfront, as there is no concurrency. There's no getting around this fundamental design flaw with `Future`, at least not without additional memory costs.

If we have to, we can have a project-level discussion around this. But, to the extent possible, I may simply invoke primary author privilege here.

belliottsmith: Can you link me to the code in C* that we think requires `Future`? I will see how difficult this is.

belliottsmith: Regarding correct usage, I would be happy to find some time (not immediately, as there are other pressures) to introduce linter checks for correct usage of callbacks, so that they are always either invoked or passed down to an inner method (that itself has the linter check run).

Would this satisfy your concerns for compiler assistance?

bdeggleston: > Could you link me to the code you're referring to that requires a Future in C*?

The method I’m talking about is here: https://github.com/bdeggleston/cassandra/blob/metadata-persistence-simulator/src/java/org/apache/cassandra/service/accord/AccordCommand.java#L636

> But, to the extent possible, I may simply invoke primary author privilege here.

It’s not possible to any extent, because you’re not the primary author. You and I are co-authors on this project. I don’t expect you to agree with me on everything, but you do need to interact with me professionally and respectfully.

Like you, I’ve been working on this project for years and have deep experience in the theory, design, and implementation of performant, concurrent databases and consensus protocols with a focus on correctness.

Regarding my concerns about raw callbacks, I’m not pushing back because it’s fun (it’s 100% the opposite), or in bad faith, or because I don’t know what I’m talking about. I’m pushing back because I genuinely feel this is a bad idea and pushing back is worth the effort. I spent years cleaning up subtle correctness bugs in C* 3.0, and the tendency of choosing a small performance improvements over safer code was a factor in many of the issues fixed. My experience there tells me that the performance improvement gained by using callbacks is not worth the risk, and reworking the api to effectively make them mandatory is a liability. This is subjective, but it is a valid, legitimate concern from a peer, co-author, and close co-worker. You and I are partners on this project. You don’t have to agree with me, but your disagreement doesn’t make my concern invalid, and it doesn’t mean you can unilaterally ignore it. I’d really like to find some common ground here, but if you’re not able to meet me in the middle, then your next step is to take it to the list.

belliottsmith: Dude, there is no disrespect here. I believe I am the primary author of this work, and I do not believe it is disrespectful for me to claim it (nor did I imagine it was a contentious claim). I also think when there is an _impasse_ with incompatible approaches to design, it is not unreasonable for a primary author to suggest their design approach should win out. You appear to disagree, but there is no disrespect.

Regarding this dispute, I do not enjoy insisting on changes any more than you enjoy pushing back. But, we cannot have totally conflicting approaches to design. If you disagree, and believe I have zero authority on the matter, then we may have to escalate to some other arbiter - that is another option I floated.

I will note, that your claims of improved type safety are weaker than you think. A serious problem with `Future` is failing to handle exceptional completion - and this _is_ required by the type system for callbacks. This is a much bigger issue than any of the very minor - easily linted - concerns you have raised about callbacks. So, I think even on the issue of compiler assistance `Future` is notably _worse_ than a callback, since I have provided multiple mitigations to your concerns with callbacks.

belliottsmith: I would also like to note explicitly, that I of course respect you highly, and I do not know what may have given the impression otherwise.

belliottsmith: To summarise the technical disagreement as I see it, here are the problems raised with each approach:

**Callbacks**
[1] May not be invoked in some no-op scenario, where an immediate `Future` might be
[2] May be invoked erroneously

**Future** 
[3] May block threads erroneously
[4] May not handle exceptional completion
[5] _Inherently_ less efficient
[6] Is more opinionated - once we’re working with `Future` it is hard to escape, and it gets woven through the APIs. Contrariwise, it is zero cost to convert a callback to a `Future`
[7] Worse debuggability

Now, I have demonstrated a callback API that eliminates [1], leaving only [2]. But I have yet to see an example of this kind of error, except perhaps the invoking-more-than-once variant, that we also have trivial mechanisms for.

However, I have _additionally_ proposed introducing linter support to address both [1] and [2], failing compilation if a callback is neither invoked directly nor passed to a child function on any branch.

Whereas [3] we have an example of _right here_, and [4] is something that I have seen often (indeed, it was an issue previously in this library).  I cannot see how they can be effectively addressed with linters, either.

Given the above, I think it is objectively the case that callbacks are more suitable in cases where they are sufficient. I would appreciate it if you could engage with my suggestions that address your concerns with callbacks, and justify why you think the concerns I raise with `Future` are _less_ impactful than those you have raised with callbacks.

accord-core/src/main/java/accord/api/Read.java
```diff
@@ -29,5 +32,5 @@
 public interface Read
 {
     Keys keys();
-    Data read(Key key, Timestamp executeAt, DataStore store);
+    Future<Data> read(Key key, boolean forWriteTxn, CommandStore commandStore, Timestamp executeAt, DataStore store);
```
dcapwell, line 35: why was `forWriteTxn` added?

accord-core/src/main/java/accord/primitives/Deps.java
dcapwell: I am 100% cool with `unsafe` static method as it makes the code clear.

accord-core/src/main/java/accord/messages/Apply.java
belliottsmith: So, last night I went back and revisited a thought I had previously, but discarded because it would be less efficient than callbacks. However, on further reflection I think the additional cost is probably acceptable.

The basic idea is to avoid most of the extra costs associated with a `Future` by delaying the submission of work. So instead of a `Future` chain we have essentially an async operation builder chain. I have a prototype here: https://github.com/belliottsmith/cassandra/commit/a1352abc82f030f376100e5116aa155bb5945c4f

You can only mutate the end of the chain (though we could add callbacks mid-chain if we wanted). The tail of the chain retains a reference to the head of the chain; when we add to the end of the chain, the old tail replaces its head pointer with the new tail pointer. When we execute, we replace the head pointer with the final callback, then invoke the head to start processing, which passes the result of execution down through the chain.

With respect to _safety_ I think this solves both of our concerns neatly, as we have all the same mechanisms for encapsulating responsibility that a `Future` does, but we also do not even start the work until some callback has been provided that _at least_ handles exceptions, and we have no `get` method to misuse (though we can always add this feature if we want; or add a way to convert to `Future`).

With respect to efficiency, the main additional cost is the encapsulating object that represents the deferred execution. Since there's no concurrency with consumer / producer, we can get away with minimal state, and can directly extend the implementing classes where efficiency matters (even for transformations). Importantly, we have zero concurrency overhead at any step of execution (as opposed to every step with `Future`).

We're still encouraging the use of lambdas and other allocating abstractions, but it's probably manageable, and as mentioned above we can reduce this where it matters.

_Personally_ I still prefer callbacks. But, since this makes each step of chaining relatively fixed-cost, I think the cognitive burden I experience reasoning about chains of operations with this abstraction will be much lower than with `Future`. Also importantly, it _looks_ exactly like a `Future` chain which is I think your main goal. This is not my preference for style, but that would be my main remaining gripe, besides the modest efficiency cost. So it is a much closer call to me, and I'm willing to subordinate my style preferences to yours here.

WDYT? Would this be a proper compromise between our approaches?

bdeggleston: This is really neat. Yes, this addresses my concerns, and I agree that this would be a good compromise. I appreciate you working with me to find a solution we both find acceptable.

I have 2 minor admin/project requests requests. First, do you mind if we punt integration of this and unblock merging metadata-persistence? I think this is the last real blocker and I’d like to get it committed before I go on vacation in a few days. Second, do you mind if _I_ do the accord and C* integration? You’re free to work on the AsyncChain interface/implementation as you like, but I’d like to do the actual integration work to adjust to the change in mental model. There are some slightly intricate places this applies to the C* integration, and I’d feel better a) doing it myself, and b) getting a little practice before doing it.

belliottsmith: Absolutely to both. I agree it would be better to get things unblocked at this stage.

Also happy to leave the AsyncChain stuff to you, or to collaborate more directly when you're back. I expect the integration will inform some of the design choices, so probably better to develop it more when that gets under way.

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -380,119 +402,162 @@ public void onChange(Command command)
...
         if (cur == null)
             return null;
 
-        Command next;
-        while (null != (next = cur.directlyBlockedBy()))
-        {
-            prev = cur;
-            cur = next;
-        }
+//        Command next;
```
dcapwell: can you remove?  no need to keep the code commented out.

accord-core/src/main/java/accord/messages/Apply.java
```diff
@@ -54,13 +62,42 @@ public Apply(Node.Id to, Topologies topologies, TxnId txnId, Txn txn, Key homeKe
...
-        Key progressKey = node.trySelectProgressKey(txnId, txn.keys, homeKey);
-        node.forEachLocalSince(scope(), executeAt,
-                               instance -> instance.command(txnId).apply(txn, homeKey, progressKey, executeAt, deps, writes, result));
-        // note, we do not also commit here if txnId.epoch != executeAt.epoch, as the scope() for a commit would be different
-        node.reply(replyToNode, replyContext, ApplyOk.INSTANCE);
+        Key progressKey = node.trySelectProgressKey(txnId, txn.keys(), homeKey);
+        List<Future<Void>> futures = node.mapLocalSince(this, scope(), executeAt, instance -> instance.command(txnId).apply(txn, homeKey, progressKey, executeAt, deps, writes, result));
+
+        ReducingFuture.reduce(futures, (l, r) -> null).addCallback((unused, failure) -> {
+            if (failure == null)
```
dcapwell, line 90: I am questioning what we can do about the failure other than log...  If `org.apache.cassandra.service.accord.db.AccordWrite#apply` or `accord.local.CommandStore#process(accord.local.PreLoadContext, java.util.function.Function<? super accord.local.CommandStore,T>)` fails we would get a failure log...

accord-core/src/main/java/accord/utils/ReducingFuture.java
```diff
@@ -0,0 +1,57 @@
...
+                result = reducer.apply(result, futures.get(i).getNow());
+
+            trySuccess(result);
+        }
+    }
+
+    public static <T> Future<T> reduce(List<? extends Future<T>> futures, BiFunction<T, T, T> reducer)
+    {
+        if (futures.isEmpty())
+            return ImmediateFuture.success(null);
```
dcapwell: `reduce` should take a `zero` (called `identity` in `Stream` API) argument if we want to allow empty, else we should reject as invalid... The 3 call paths ATM just happen to return `null` but that isn't universal (and this class is trying to be so).

accord-core/src/main/java/accord/utils/ReducingFuture.java
```diff
@@ -0,0 +1,57 @@
...
+    private final BiFunction<V, V, V> reducer;
+    private volatile int pending;
+
+    protected ReducingFuture(List<? extends Future<V>> futures, BiFunction<V, V, V> reducer)
+    {
+        this.futures = futures;
+        this.reducer = reducer;
+        this.pending = futures.size();
+        if (futures.size() == 0)
+            trySuccess(null);
```
dcapwell: rather than duplicate the logic, can `accord.txn.Txn.ReadFuture#ReadFuture` be rewritten to use `accord.utils.ReducingFuture#reduce`?  It looks like the whole reason the `ReadFuture` exists is because `org.apache.cassandra.service.accord.AccordCommand#read` has a cache and wants to check the `scope`... but we know the `scope` when when we are putting into the cache, so we could avoid having `ReadFuture` and move this logic to the cache in C*.

dcapwell: This also gets back to the `zero` point I made in `reduce`, `org.apache.cassandra.service.accord.db.AccordData#merge(accord.api.Data)` doesn't handle `null` so its not clear from the API that `null` will be "empty".  In this context we would call `Txn.read` and not actually read anything (no command store intersects the key), so this would be a bug... by changing to `null` we push this bug into the user of the future who then has to figure out that this happened, which I think will NPE in `accord.messages.ReadData.LocalRead#readComplete` or `org.apache.cassandra.io.IVersionedSerializer#serialize` depending on ordering...

accord-core/src/main/java/accord/api/Read.java
bdeggleston: it's for the C* integration timestamp calculation. `AccordCommandsForKey#maybeUpdateLateTimestamp` is where it's used

accord-core/src/main/java/accord/utils/ReducingFuture.java
bdeggleston: fixed so an empty list throws an exception

accord-core/src/main/java/accord/utils/ReducingFuture.java
bdeggleston: > can accord.txn.Txn.ReadFuture#ReadFuture be rewritten to use accord.utils.ReducingFuture#reduce?

The reason I didn't do that is that the cache has some generalized future handling which we lose in that case. Though I suppose a `Pair<Keys, Future>` isn't any better than a one off Future implementation that wraps the output of reduce and stores 'Keys'

accord-core/src/main/java/accord/messages/Apply.java
bdeggleston: There's not a protocol need to do anything else. From the client perspective the operation succeeded, though later operations on the key(s) will fail if apply can't succeed. I do think that logging and moving on is consistent with how we handle single node failures in things like quorum writes though.

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -380,119 +402,155 @@ public void onChange(Command command)
...
+    protected Future<Void> apply()
+    {
+        // important: we can't include a reference to *this* in the lambda, since the C* implementation may evict
+        // the command instance from memory between now and the write completing (and post apply being called)
+        return writes().apply(commandStore()).flatMap(unused ->
+            commandStore().process(this, callPostApply(txnId()))
+        );
+    }
+
+    public Future<Data> read(Keys scope)
```
dcapwell, line 452: should we remove `Keys scope`?  The previous usage was just to populate the `ReadFuture.scope` and was ignored

accord-core/src/main/java/accord/txn/Txn.java
```diff
@@ -18,93 +18,140 @@
...
                 return accumulate;
 
-            Data result = read.read(key, command.executeAt(), commandStore.store());
-            return accumulate != null ? accumulate.merge(result) : result;
-        }, null);
+            Future<Data> result = read().read(key, isWrite(), command.commandStore(), command.executeAt(), commandStore.store());
+            accumulate.add(result);
+            return accumulate;
+        }, new ArrayList<>());
+        return ReducingFuture.reduce(futures, (d1, d2) -> d1.merge(d2));
```
dcapwell: can we do `return ReducingFuture.reduce(futures, Data::merge);` instead?

accord-core/src/main/java/accord/utils/ReducingFuture.java
```diff
@@ -0,0 +1,56 @@
...
+            for (int i=1, mi=futures.size(); i<mi; i++)
+                result = reducer.apply(result, futures.get(i).getNow());
+
+            trySuccess(result);
+        }
+    }
+
+    public static <T> Future<T> reduce(List<? extends Future<T>> futures, BiFunction<T, T, T> reducer)
+    {
+        Preconditions.checkArgument(!futures.isEmpty());
```
dcapwell: I know Apache Cassandra doesn't provide messages often, but I find that makes life much harder... can we have a simple message like "future list is empty"?

accord-core/src/main/java/accord/utils/ReducingFuture.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.function.BiFunction;
+
+public class ReducingFuture<V> extends AsyncPromise<V>
+{
+    private static final AtomicIntegerFieldUpdater<ReducingFuture> PENDING_UPDATER = AtomicIntegerFieldUpdater.newUpdater(ReducingFuture.class, "pending");
+    private final List<? extends Future<V>> futures;
+    private final BiFunction<V, V, V> reducer;
+    private volatile int pending;
+
+    protected ReducingFuture(List<? extends Future<V>> futures, BiFunction<V, V, V> reducer)
```
dcapwell: can we make this private?

accord-core/src/main/java/accord/utils/ReducingFuture.java
dcapwell: The constructor is no longer used and `reduce` prevents empty, so we can remove the `trySuccess(null);`

dcapwell (changes requested):

accord-core/src/main/java/accord/local/Command.java
bdeggleston: no, it's still used on the c* side to populate it's read future

dcapwell (approved):

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -380,119 +402,155 @@ public void onChange(Command command)
...
 
     public BlockedBy blockedBy()
     {
         Command prev = this;
-        Command cur = directlyBlockedBy();
+        PartialCommand cur = directlyBlockedBy();
         if (cur == null)
             return null;
 
-        Command next;
```
belliottsmith, line 485: Removing this loop breaks the progress log. I can see why it has been done, but we'll have to do a bit more work here.

accord-core/src/main/java/accord/messages/Apply.java
belliottsmith: I may have to walk that back unfortunately. I'll do my best to minimise my interactions with this code, but it touches a lot of what I'm rebasing against, and I'm not sure how much I can leave it untouched in your absence.

bdeggleston: FWIW, temporarily converting to Future should accomplish 95% of the conversion to asyncchain, with just the future producer consumer needing any real changes, and the rest being a rename to asyncchain. If there’s something I’m. It thinking of that would complicate making that an intermediate step and you need to do some integration, that’s fine

bdeggleston: Tangentially, there may be something other than future support I’m missing, but if AsyncChain lives in accord at first and replaces future, we may also be able to eliminate the circular C* dependency, which would make discussions about incrementally merging a bit easier

accord-core/src/main/java/accord/local/Command.java
bdeggleston: This was the change i was trying to validate with the burn test, but it was already hanging on the progress log. Sorry if I forgot to explicitly call this change out, but yeah, we can’t pointer chase in C*

accord-core/src/main/java/accord/messages/Apply.java
belliottsmith: Yeah, I may be able to leave `AsyncChain` itself alone for now, but I'll be tweaking the control flow for processing of messages by command stores, as its very different with partial replication (with the concept of deferred execution when earlier messages haven't arrived, amongst other things). I had hoped to leave it all entirely untouched until we could collaborate, since it's all intertwined, but we can revisit it when you're back.

PR #11, opened 2022-10-24
Feedback/cassandra 17103

dcapwell:


accord-core/src/main/java/accord/local/Status.java
```diff
@@ -52,4 +52,16 @@ public boolean hasBeen(Status equalOrGreaterThan)
     {
         return compareTo(equalOrGreaterThan) >= 0;
     }
+
+    public boolean isInvalidated()
```
dcapwell, line 56: used here: https://github.com/apache/cassandra/pull/1948/files#diff-727e032963026a3afc0cac7963951ab882aadfaa1a73e2a666020c8e7329755eR243

bdeggleston (approved):
+1

PR #12, opened 2022-11-08
Add transaction expiration

belliottsmith:


accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -163,7 +157,17 @@ public long latestEpoch()
         }
 
         @Override
-        public Timestamp maxConflict(Keys keys)
+        public Timestamp preaccept(TxnId txnId, Keys keys)
+        {
+            Timestamp max = maxConflict(keys);
+            long epoch = latestEpoch();
+            if (txnId.compareTo(max) > 0 && txnId.epoch >= epoch && !agent.isExpired(txnId, time.now()))
```
aweisberg, line 164: Why return a new timestamp if is expired? I thought the expiration check is done on the coordinator now?

aweisberg (changes requested):
Only the one question. Wondering if that `isExpired` check in `InMemoryCommandStore` is left over from when expiration sent back the max ballot.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
belliottsmith: This permits a majority of remote replicas to also collectively reject an expired transaction received from a long running coordinator  - and also means that recovery doesn't accidentally register a preaccept that can be counted towards the original coordination. Though perhaps for that we should just explicitly refuse to use the original ballot for recovery.

This is all sort of arbitrary, so happy with any of these approaches.

aweisberg (approved):
+1 TY

PR #13, opened 2022-11-09
Refactor response tracking to improve efficiency and clarity

belliottsmith:


accord-core/src/main/java/accord/coordinate/CheckOn.java
```diff
@@ -94,23 +94,39 @@ protected AbstractRoute route()
...
+        KeyRanges rangesForNode = topologies().computeRangesForNode(from);
+        PartialRoute scope = this.route.slice(rangesForNode);
+        return isSufficient(scope, ok);
     }
 
     @Override
-    protected void onDone(Done done, Throwable failure)
+    protected boolean isSufficient(CheckStatusOk ok)
     {
+        return isSufficient(route, ok);
```
aweisberg, line 100: Should this basically be `UnsupportedOperationException` because `isSufficient` is now overridden?

accord-core/src/main/java/accord/coordinate/CheckOn.java
```diff
@@ -94,23 +94,39 @@ protected AbstractRoute route()
     @Override
     protected boolean isSufficient(Id from, CheckStatusOk ok)
     {
-        return ((CheckStatusOkFull)ok).sufficientFor(route()).compareTo(sufficient) >= 0;
+        KeyRanges rangesForNode = topologies().computeRangesForNode(from);
+        PartialRoute scope = this.route.slice(rangesForNode);
```
aweisberg, line 93: If we don't use the route that was provided to us to check sufficiency then doesn't the check that we have *alll* the information for the route no longer work?

I don't get how it works in general just because the `CheckStatusOK` we are checking is a fraction of the information we have collected and `CheckOn` maybe be trying to cover more shards so we might have sufficient information, but never notice with this approach to checking if responses up to this point were sufficient?

accord-core/src/main/java/accord/coordinate/CheckOn.java
```diff
@@ -94,23 +94,39 @@ protected AbstractRoute route()
...
             callback.accept(null, failure);
         }
-        else if (merged == null || merged.saveStatus == NotWitnessed)
-        {
-            callback.accept(CheckStatusOkFull.NOT_WITNESSED, null);
-        }
         else
         {
-            new OnDone().start();
+            if (success == Success.Success)
```
aweisberg, line 118: This wouldn't also be true with quorum?

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: No, we separately validate the final `merged` directly with this variant.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: The idea is that we validate that we have sufficient information _from each shard_. So, if the keys for the shards covered by the node whose response we have received is considered sufficient for its ranges, then once we receive such responses from every shard we should have sufficient information altogether. We validate this before invoking the final callback by invoking `isSufficient(merged)`.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: No, if anything the opposite. The point of `isSufficient` is to decide when we hit `Accept`. If we hit `AcceptQuorum` we should be `!isSufficient` here.

accord-core/src/main/java/accord/coordinate/CheckOn.java
aweisberg: Ok, so I think the thing I probably missed isSufficient is checked per shard. I have been living my life thinking we were checking for the whole thing.

accord-core/src/main/java/accord/coordinate/CheckOn.java
aweisberg: Oh right... AcceptQuorum is bubbling up here, but it's actually an indication of insufficient. I think that enum value could be better named.

Should it be `Insufficient` and `Sufficient` instead of `Accept` and `AcceptQuorum`?

belliottsmith: Perhaps `AcceptNow` and `AcceptQuorum`? It feels like the concept of sufficiency is fairly local, and probably shouldn't bubble up to the tracker - it only really knows if we're done _now_ or done when we hit a quorum. But perhaps we can standardise terminology between the two.

aweisberg: What makes it confusing is `Accept` since that is overloaded with unrelated consensus things.

Mentioning quorum seems unnecessary and maybe even misleading since a lot of this is about skipping waiting on the quorum. I think it would be clearer to lean into naming that focuses more on sufficiency or response collection.

Also seems like `ReadCoordinator` isn't necessarily `Reads`? Maybe that is an implementation detail of what is using it?

Maybe:
`AcceptQuorum` -> `NotSufficient`
`Accept` -> `Sufficient`
`recordQuorumReadSuccess` -> `recordInsufficientResponse`
`recordReadSuccess` -> `recordSufficientResponses`

I am kind of assuming that `isSufficient` can occur due to information collected across multiple responses and not just one.

It's pretty minor, the existing naming is ok it is really the `Accept` in `AcceptQuorum` and `Accept` that is confusing because `AcceptQuorum` sure sounds like it is done, but it doesn't mean that.

accord-core/src/main/java/accord/coordinate/CheckShards.java
```diff
@@ -44,12 +42,13 @@ private static Topologies topologyFor(Node node, RoutingKeys someKeys, long epoc
     }
 
     @Override
-    protected void contact(Set<Id> nodes)
+    public void contact(Id id)
```
aweisberg: Why public now?

accord-core/src/main/java/accord/coordinate/Coordinate.java
```diff
@@ -203,29 +113,6 @@ public void onCallbackFailure(Id from, Throwable failure)
         tryFailure(failure);
     }
 
-    // TODO (soon): I don't think we need to preaccept in later epochs? the Sync logic should take care of it for us,
-    //              since either we haven't synced between a majority and the earlier epochs are still involved for
-    //              later preaccepts, or they have been sync'd and the earlier transactions are known to the later epochs
-    //              (also, we aren't doing this on recovery, and everything works...)
-    private synchronized void onEpochUpdate()
```
aweisberg, line 210: Why is contacting nodes as topologies change no longer necessary here?

accord-core/src/main/java/accord/coordinate/Coordinate.java
```diff
@@ -296,7 +175,12 @@ private void onPreAccepted()
             }
             else
             {
-                node.withEpoch(executeAt.epoch, () -> Propose.propose(node, tracker.topologies(), Ballot.ZERO, txnId, txn, route, executeAt, deps, this));
+                node.withEpoch(executeAt.epoch, () -> {
+                    Topologies topologies = tracker.topologies();
+                    if (executeAt.epoch > txnId.epoch)
+                        topologies = node.topology().withUnsyncedEpochs(txn, txnId.epoch, executeAt.epoch);
```
aweisberg, line 181: We now make sure to update topologies `withUnynchedEpochs` before moving to `Propose`. Intuitively this because in the slow path they may not have gotten enough information, (with fast path accept  they will) so we need to include them in the propose?

Is this why we can remove the other code that would have included them in the preaccept round?

accord-core/src/main/java/accord/coordinate/InformHomeOfTxn.java
```diff
@@ -30,18 +30,21 @@
...
 
+import static accord.coordinate.tracking.AbstractTracker.ShardOutcomes.Fail;
+import static accord.coordinate.tracking.AbstractTracker.ShardOutcomes.Success;
+
 public class InformHomeOfTxn extends AsyncFuture<Void> implements Callback<SimpleReply>
 {
     final TxnId txnId;
     final RoutingKey homeKey;
-    final QuorumShardTracker tracker;
+    final QuorumTracker.QuorumShardTracker tracker;
```
aweisberg: Why remove the static import?

accord-core/src/main/java/accord/coordinate/tracking/QuorumTracker.java
```diff
@@ -18,27 +18,74 @@
...
+    {
+        protected int successes;
+        protected int failures;
+
+        public QuorumShardTracker(Shard shard)
+        {
+            super(shard);
+        }
+
+        public ShardOutcomes onSuccess(Object ignore)
```
aweisberg, line 39: Why does the ignore object need to be present? This isn't an override method stuck with an existing signature.

accord-core/src/main/java/accord/coordinate/Invalidate.java
```diff
@@ -55,8 +58,9 @@ public class Invalidate implements Callback<InvalidateReply>
     final BiConsumer<Outcome, Throwable> callback;
 
     boolean isDone;
+    boolean isBallotPromised;
     final List<InvalidateOk> invalidateOks = new ArrayList<>();
-    final QuorumShardTracker preacceptTracker;
+    final QuorumTracker.QuorumShardTracker tracker;
```
aweisberg: Bring back static import?

accord-core/src/main/java/accord/coordinate/Invalidate.java
```diff
@@ -68,15 +72,17 @@ private Invalidate(Node node, Shard shard, Ballot ballot, TxnId txnId, RoutingKe
...
 
-    public static Invalidate invalidateIfNotWitnessed(Node node, TxnId txnId, RoutingKeys informKeys, RoutingKey invalidateWithKey, BiConsumer<Outcome, Throwable> callback)
+    public static Invalidate invalidate(Node node, TxnId txnId, RoutingKeys informKeys, RoutingKey invalidateWithKey, BiConsumer<Outcome, Throwable> callback)
     {
         return invalidate(node, txnId, informKeys, invalidateWithKey, PreAccepted, callback);
     }
 
-    public static Invalidate invalidate(Node node, TxnId txnId, RoutingKeys informKeys, RoutingKey invalidateWithKey, BiConsumer<Outcome, Throwable> callback)
+    // TODO (now): (file separately) this is a bug, as it is possible for a preaccept to still be in-flight.
+    //             It's not even safe to do it after an initial Invalidate round unless we ensure that recovery never pre-accepts with txnId (which we should also enforce)
```
aweisberg, line 84: So post this patch it still doesn't work? Invalidate and preaccept racing like this, is it in the paper?

Invalidate is running the second round, why would it invalidate something that is preaccepted? Wouldn't enough electors refuse to accept the invalidation?

accord-core/src/main/java/accord/coordinate/Invalidate.java
```diff
@@ -113,33 +119,35 @@ public synchronized void onSuccess(Id from, InvalidateReply reply)
...
         InvalidateOk ok = (InvalidateOk) reply;
         invalidateOks.add(ok);
-        if (preacceptTracker.success(from))
+        if (tracker.onSuccess(from) == Success)
             invalidate();
     }
 
     private void invalidate()
     {
+        isBallotPromised = true;
```
aweisberg, line 129: This doesn't seem like quite the appropriate name. This seems like a basic we already started the invalidate so don't start it again?

Where as `isDone` is for the callback?

I am somewhat curious how `isDone` is needed for the callback. Why are we invoking the callback multiple times? Is it just defensive?

accord-core/src/main/java/accord/coordinate/Invalidate.java
```diff
@@ -113,33 +119,35 @@ public synchronized void onSuccess(Id from, InvalidateReply reply)
...
         // first look to see if it has already been
         {
-            Status maxStatus = invalidateOks.stream().map(ok -> ok.status).max(Comparable::compareTo).orElseThrow(IllegalStateException::new);
             Route route = InvalidateOk.findRoute(invalidateOks);
             RoutingKey homeKey = route != null ? route.homeKey : InvalidateOk.findHomeKey(invalidateOks);
+            SaveStatus maxStatus = invalidateOks.stream().map(ok -> ok.status).max(Comparable::compareTo).orElseThrow(IllegalStateException::new);
```
aweisberg, line 134: Doesn't look like this needed to move, diff would just be the added line maxKnown if it hadn't moved.

accord-core/src/main/java/accord/coordinate/Invalidate.java
```diff
@@ -211,10 +218,10 @@ else if (homeKey != null)
     @Override
     public void onFailure(Id from, Throwable failure)
```
aweisberg, line 225: `Callback` has `onFailure` and `onCallbackFailure` and doesn't document the difference, although yes it is obvious once you look at the usage.

`onCallbackFailure` bails out on `isDone`, but `isDone` is set to true in `onFailure`. The strategy here seems like it doesn't necessarily do what it is trying to do which is allow callbacks to handle their own errors. And then why? Shouldn't the callback's implement their own error handling by extending a wrapper class that does the try catch and then calls whatever is appropriate?

accord-core/src/main/java/accord/coordinate/MaybeRecover.java
```diff
@@ -74,18 +74,15 @@ public boolean hasMadeProgress(CheckStatusOk ok)
     }
 
     @Override
-    protected void onDone(Done done, Throwable fail)
+    protected void onDone(Success success, Throwable fail)
```
aweisberg, line 77: There are just sooo many callbacks with generic names like onSuccess, onFailure, onDone, and you have to make an extra hop to find out what they are the callback for. If we could avoid having generic names as much as possible and specify something like `onCheckShardsDone` even if it means declaring one off interfaces for each use case I think it would be an improvement.

accord-core/src/main/java/accord/coordinate/CheckShards.java
```diff
@@ -44,12 +42,13 @@ private static Topologies topologyFor(Node node, RoutingKeys someKeys, long epoc
...
     {
-        node.send(nodes, new CheckStatus(txnId, contactKeys, txnId.epoch, untilRemoteEpoch, includeInfo), this);
+        node.send(id, new CheckStatus(txnId, contactKeys, txnId.epoch, untilRemoteEpoch, includeInfo), this);
     }
 
-    protected abstract boolean isSufficient(Id from, CheckStatusOk ok);
+    protected boolean isSufficient(Id from, CheckStatusOk ok) { return isSufficient(ok); }
+    protected abstract boolean isSufficient(CheckStatusOk ok);
 
     protected Action check(Id from, CheckStatusOk ok)
```
aweisberg:
```suggestion
    protected Action checkSufficient(Id from, CheckStatusOk ok)
```
WDYT?

accord-core/src/main/java/accord/coordinate/ReadCoordinator.java
```diff
@@ -0,0 +1,157 @@
...
+        {
+            default: throw new AssertionError();
+            case NoChange:
+                break;
+            case Success:
+                Preconditions.checkState(!isDone);
+                isDone = true;
+                onDone(waitingOnData == 0 ? Success.Success : Success.ReachedQuorum, null);
+                break;
+            case Failed:
```
aweisberg, line 157: Near as I Can tell this is unreachable because `Failed` is used in quite a few checks but never actually returned?

accord-core/src/main/java/accord/coordinate/ReadCoordinator.java
```diff
@@ -0,0 +1,157 @@
...
+        this.failure = failure;
+        finishOnFailure();
+    }
+
+    protected void finishOnFailure()
+    {
+        Preconditions.checkState(!isDone);
+        isDone = true;
+        if (failure == null)
+            failure = new Exhausted(txnId, null);
```
aweisberg, line 141: How sure are we that exhaustion was the issue?

This is only effectively called from `onCallbackFailure` which is an exception in the callback and not a timeout?

accord-core/src/main/java/accord/topology/Topologies.java
```diff
@@ -122,13 +113,22 @@ static String toString(Topologies t)
 
     class Single implements Topologies
     {
+        private final TopologySorter sorter;
         private final Topology topology;
         private final boolean fastPathPermitted;
```
aweisberg: Unused now

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -149,7 +121,8 @@ public boolean hasMetFastPathCriteria()
     private boolean isDone;
 
     final List<RecoverOk> recoverOks = new ArrayList<>();
-    final FastPathTracker<ShardTracker> tracker;
+    final RecoveryTracker tracker;
```
aweisberg: Why so much package private stuff here instead of private?

accord-core/src/main/java/accord/coordinate/Invalidate.java
```diff
@@ -68,15 +72,17 @@ private Invalidate(Node node, Shard shard, Ballot ballot, TxnId txnId, RoutingKe
         this.informKeys = informKeys;
         this.invalidateWithKey = invalidateWithKey;
         this.recoverIfAtLeast = recoverIfAtLeast;
-        this.preacceptTracker = new QuorumShardTracker(shard);
+        this.tracker = new QuorumTracker.QuorumShardTracker(shard);
     }
 
-    public static Invalidate invalidateIfNotWitnessed(Node node, TxnId txnId, RoutingKeys informKeys, RoutingKey invalidateWithKey, BiConsumer<Outcome, Throwable> callback)
+    public static Invalidate invalidate(Node node, TxnId txnId, RoutingKeys informKeys, RoutingKey invalidateWithKey, BiConsumer<Outcome, Throwable> callback)
```
aweisberg, line 78: I have to admit I looked at where we invalidate vs invalidateIfNotWitnessed and I can't tell why one recovers if PreAccepted vs Accepted or why you renamed.

Why shouldn't it be `invalidateIfNotWitnessed` and `invalidateIfNotAccepted`?

Probably should document the methods.

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
```diff
@@ -0,0 +1,205 @@
...
+            Topology topology = topologies.get(i);
+            int size = topology.size();
+            for (int j = 0; j < size; ++j)
+                trackers[i * maxShards + j] = trackerFactory.apply(topology.get(j));
+            waitingOnShards += topologies.get(i).size();
+        }
+        this.waitingOnShards = waitingOnShards;
+    }
+
+    AbstractTracker(Topologies topologies, IntFunction<ST[]> arrayFactory)
```
aweisberg: Multiple constructors that is really just one?

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
```diff
@@ -0,0 +1,205 @@
...
+    {
+        Preconditions.checkArgument(topologies.totalShards() > 0);
+        int topologyCount = topologies.size();
+        int maxShards = topologies.get(0).size();
```
aweisberg: This duplicates `Topologies.maxShardsPerEpoch()`?

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
```diff
@@ -0,0 +1,205 @@
...
+
+    @VisibleForTesting
+    public ST unsafeGet(int topologyIdx, int shardIdx)
+    {
+        if (shardIdx >= maxShardsPerEpoch())
+            throw new IndexOutOfBoundsException();
+        return trackers[topologyOffset(topologyIdx) + shardIdx];
+    }
+
+    protected int maxShardsPerEpoch()
```
aweisberg, line 193: Store precomputed?

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
```diff
@@ -0,0 +1,205 @@
...
+
+    final Topologies topologies;
+    protected final ST[] trackers;
+    protected int waitingOnShards;
+
+    AbstractTracker(Topologies topologies, IntFunction<ST[]> arrayFactory, Function<Shard, ST> trackerFactory)
+    {
+        this(topologies, arrayFactory);
+        int topologyCount = topologies.size();
+        int maxShards = maxShardsPerEpoch();
```
aweisberg: Was this already computed in the other constructor?

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
```diff
@@ -0,0 +1,205 @@
...
+            minShardStatus = topologies.get(i).mapReduceOn(node, i * maxShards, AbstractTracker::apply, self, function, param, ShardOutcomes::min, minShardStatus);
+        }
+
+        switch (minShardStatus)
+        {
+            default: throw new AssertionError();
+            case SendMore:
+                return trySendMore();
+            case Success:
+                if (waitingOnShards == 0)
```
aweisberg, line 130: What does it mean if `waitingOnShards != 0` but map reducing on min. status is `Success`. Is that a bug?

accord-core/src/main/java/accord/coordinate/tracking/ReadTracker.java
```diff
@@ -18,232 +18,281 @@
...
+        public ShardOutcome<? super ReadTracker> recordSlowResponse(boolean ignore)
         {
-            ++contacted;
+            Preconditions.checkState(!hasFailed());
+            ++slow;
+
+            if (shouldRead() && canRead())
+                return SendMore;
+
+            return NoChange;
```
aweisberg, line 72: If we can't read isn't that failure?

Slow means we contacted someone without the info and if none of them have the info does it mean we are waiting on the progress logs to catch them up?

Will this read tracker terminate properly the operation properly if it receives response from everything, and they are all slow?

accord-core/src/main/java/accord/coordinate/tracking/ReadTracker.java
```diff
@@ -18,232 +18,281 @@
...
         }
 
-        public void recordSlowRead(Id node)
+        /**
+         * Have received a requested data payload with desired contents
+         */
+        public ShardOutcome<? super ReadTracker> recordReadSuccess(boolean isSlow)
```
aweisberg, line 78: This doesn't increment `acceptQuorum`, but it goes to the trouble of updating `inflight` and `slow`. None seem necessary since we return `DataSuccess` and if that `DataSuccess` isn't processed to completion nothing works anyways.

accord-core/src/main/java/accord/coordinate/tracking/ReadTracker.java
```diff
@@ -18,232 +18,281 @@
...
-     */
-    public Set<Id> computeMinimalReadSetAndMarkInflight()
+    public <T1> RequestStatus trySendMore(BiConsumer<T1, Id> contact, T1 with)
     {
-        Set<ReadShardTracker> toRead = foldl((tracker, accumulate) -> {
-            if (!tracker.shouldRead())
-                return accumulate;
+        ShardSelection toRead;
+        {
+            ShardSelection tmp = null;
```
aweisberg, line 245: Is `tmp` necessary?

accord-core/src/main/java/accord/impl/SizeOfIntersectionSorter.java
```diff
@@ -0,0 +1,51 @@
...
+            count2 += count(node2, shards, i * maxShardsPerEpoch, topology);
+        }
+
+        // sort more intersections later
+        return count1 - count2;
+    }
+
+    private static int count(Node.Id node, ShardSelection shards, int offset, Topology topology)
+    {
+        return topology.foldlIntOn(node, (j, s, v) -> s.get(j) ? v + 1 : v, shards, offset, 0, 0);
```
aweisberg: j, s, v is a little terse

accord-core/src/main/java/accord/topology/Topologies.java
```diff
@@ -298,10 +308,18 @@ public boolean contains(Id to)
             return false;
         }
 
+        @Override
+        public int estimateUniqueNodes()
```
aweisberg, line 309: This seems like a hefty over estimate if it is used allocate memory. Can't we precompute the precise answer and then memoize it and maybe combine with constructing the collection? Seems like in most cases we end up calling `Topologies.nodes()` and that calculates the precise set of things anyways so maybe just do it in advance?

accord-core/src/main/java/accord/topology/Topology.java
```diff
@@ -323,6 +321,35 @@ else if (a[ai] < b[bi])
         }
     }
 
+    public <P1, P2, P3, O> O mapReduceOn(Id on, int offset, IndexedTriFunction<? super P1, ? super P2, ? super P3, ? extends O> function, P1 p1, P2 p2, P3 p3, BiFunction<? super O, ? super O, ? extends O> reduce, O initialValue)
+    {
+        NodeInfo info = nodeLookup.get(on);
+        if (info == null)
+            return initialValue;
+        int[] a = supersetIndexes, b = info.supersetIndexes;
```
aweisberg, line 310: supersetIndexes seems to be the integer shard ids. For the entire topology it is just the sequence of numbers, but for `NodeInfo` is is the array of integers of shards it replicates.


Could the name change or add a comment?

aweisberg (changes requested):
I will say it was easy to understand everything even if it took a bit to read through.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: Perhaps simply `Success` and `SuccessIfQuorum`?

accord-core/src/main/java/accord/coordinate/Coordinate.java
belliottsmith: I believe I added a comment to explain it? If not I will do so. Basically it was never necessary - the reason to contact them (in the paper) was to seek promises, but we only need promises from the `txnId.epoch` shard - we can happily propose to the other shard with the promise we received from the original shard. And this simplifies things dramatically.

accord-core/src/main/java/accord/coordinate/Coordinate.java
belliottsmith: Essentially here we're just updating the topologies in the same way the deleted method would have done previously, to ensure we contact the new shards in the propose step.

accord-core/src/main/java/accord/coordinate/InformHomeOfTxn.java
belliottsmith: Probably IntelliJ

accord-core/src/main/java/accord/coordinate/tracking/QuorumTracker.java
belliottsmith: Method handles to provide to `AbstractTracker.recordResponse`

accord-core/src/main/java/accord/coordinate/Invalidate.java
belliottsmith: Take a look at my next patch. But the problem here is the `invalidateWithKey` which is only safe is we have prevented a pre-accept from taking place with our ballot, otherwise a pre-accept might have been inflight and reached there between when we started the invalidation and our BeginInvalidation arrives.

accord-core/src/main/java/accord/coordinate/Invalidate.java
belliottsmith: This is because we may have to _propose_ an invalidation, and we want to separately say we've finished the invalidation, and we've finished the preparation for taking our next step.

`isBallotPromised` should be redundant, but `isDone` might not be at least in the case of `onCallbackFailure`

accord-core/src/main/java/accord/coordinate/Invalidate.java
belliottsmith: This is something defensive to ensure it's considered by the implementor - which generally it wasn't being. I'm not sure what your concern is about `isDone` - the idea here is we don't want to invoke the callback multiple times...

accord-core/src/main/java/accord/coordinate/Invalidate.java
belliottsmith: This is related to the bug, and I have fixed it in a follow-up PR.

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
belliottsmith: Integer division has a latency of 10 cycles. Worth it? Perhaps, sure.

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
belliottsmith: Abandoned refactor, will merge

accord-core/src/main/java/accord/coordinate/Recover.java
belliottsmith: 🤷  - in a number of places it has been helpful recently, so it may have become my default by accident, but honestly I think I've fallen out of love with `private` except where I really want to enforce it strongly; it's just extra words (a bit like `final` for local variables). I'm king of my package, I can behave nicely. Don't feel strongly about it, happy to make private where relevant.

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
belliottsmith: This `Success` is per-shard, i.e. it means at least one shard returned success, and no shards returned failure. So if only one shard has returned `Success` but not all, we could absolutely still have `waitingOnShards != 0`

accord-core/src/main/java/accord/coordinate/tracking/ReadTracker.java
belliottsmith: Slow means we haven't received a prompt response? It means we still expect another response (either timeout or success or whatever).

Since this is not a terminal response, being unable to read isn't a failure, no.

accord-core/src/main/java/accord/coordinate/tracking/ReadTracker.java
belliottsmith: `acceptQuorum` is a separate concept - why would we increment it? We aren't accepting a quorum, we are accepting this single "sufficient" response.

I prefer to maintain my invariants wherever possible, even if only to assist testing and debugging.

accord-core/src/main/java/accord/coordinate/tracking/ReadTracker.java
belliottsmith: `toRead` must be final to be referenced in a lambda

accord-core/src/main/java/accord/topology/Topologies.java
belliottsmith: Is it a hefty calculation? It's _min_, so at e.g. rf=9 and 2 concurrent topologies it will allocate 11. But that should be 10 (it should be size() -1), I'll update that.

I don't mind memoizing, but this is not likely to be invoked more than once or so per lifetime of a `Topologies` and we'll probably allocate more memory calculating an accurate figure than we will guessing incorrectly.

accord-core/src/main/java/accord/topology/Topology.java
belliottsmith: supersetIndexes is not the shard ids, it is the indexes in which the shards occur. For a `Topologies` that hasn't been sliced it will be an array containing the integers [0..size) for simplicity. I'm not sure what you're asking to change?

accord-core/src/main/java/accord/impl/SizeOfIntersectionSorter.java
belliottsmith: This is a very short function definition, and I find it clearer when it can fit in a small space. I would be open to renaming `s` to `shard` - would that suffice?

belliottsmith: Also it looks like `j` could be `i` here, presumably it's been moved from somewhere there was a clash.

accord-core/src/main/java/accord/coordinate/ReadCoordinator.java
belliottsmith: This is called by `handle` on `Failed` - so if there is no exception then it's because we have received responses from everyone but `Reject`ed them.

accord-core/src/main/java/accord/coordinate/ReadCoordinator.java
belliottsmith: Why do you think `Failed` is never returned? `ReadShardTracker.recordReadFailure` invokes `ensureProgressOrFail` which will return `Fail` in the event it has exhausted its candidates. This is the min `ShardOutcomes` so will be used by `AbstractTracker` to propagate a `RequestStatus.Failed` response.

accord-core/src/main/java/accord/coordinate/MaybeRecover.java
belliottsmith: I think we have only one abstract `onDone` in the whole project, and only one `onSuccess`* - the latter is for `Callback` which we can't really escape, and the former is in `ReadTracker` - and it's a class, not an interface. So unless we want to introduce some wrapper methods just to translate the name I'm not 100% sure what's being suggested. Feel free to propose some changes in a patch?

\* We do have an `onSuccess` in `QuorumShardTracker` too, but I can't currently think of a better name for a generic success response.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: or `AcceptIfQuorum` or `AcceptAQuorum`?

belliottsmith: I had a look at a thesaurus, and another candidate is `Approve` and `ApproveIfQuorum`. I'll go with that for now.

accord-core/src/main/java/accord/coordinate/Coordinate.java
belliottsmith: Line 132-134. They may be insufficient, however,

accord-core/src/main/java/accord/coordinate/Invalidate.java
belliottsmith: We also change it to SaveStatus. Though I think I might revert that in the follow-up patch, which does more significant refactoring of Invalidation. I would not look too closely at this for this patch. OR we can pull that later patch into this PR if you prefer.

belliottsmith: I mostly split it up to accelerate review. Since they're both ready and we're reviewing, don't mind how we approach it.

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
belliottsmith: Actually this is the initial calculation that is required to permit the later calculation to work. But I have memoized this now.

belliottsmith:
I believe I have addressed everything outstanding, except those where I require further information.

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
aweisberg: Hmm sorry I may have left the comment in the wrong place. Agree it wasn't worth pre-computing.

accord-core/src/main/java/accord/topology/Topology.java
aweisberg: Hmm so why create an array of integers instead of an array of object reference to the actual shards? It's twice the space sure, but simpler.

A comment saying that `Superset indexes are indexes of the shards that are present in this topology (or node info) out of the shards array` would make it quick to understand.

accord-core/src/main/java/accord/topology/Topologies.java
aweisberg: Yeah seems fine. I think I thought `size()` was a deep size not number of topologies.

accord-core/src/main/java/accord/topology/Topology.java
belliottsmith: Because operating on integers is much more efficient, particularly intersecting two different arrays. It’s nothing to do with storage efficiency.

accord-core/src/main/java/accord/impl/SizeOfIntersectionSorter.java
aweisberg: It's mostly just my lack of comfort with left fold in general. Renaming to `shard` and `I` instead of `j` helps thanks.

accord-core/src/main/java/accord/coordinate/tracking/ReadTracker.java
aweisberg: I was thinking that a read that satisfies all the requirements technically also contributed to satisfying the quorum requirement, but tracking all those things doesn't seem to matter much because we got a sufficient response anyways.

So I was wondering why not also increment the fact this response contributed to a quorum.

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
aweisberg: Ah, I keep getting caught by that distinction when we do the min check across something that is over multiple shard and all it means is one shard has achieved that state.

accord-core/src/main/java/accord/coordinate/Recover.java
aweisberg: It's pretty low on the list of things I care about, but when I see something a package private I think it's a hint that it is intended to be part of the public (well package public) API of the class and part of how it operate.b

Having things that looks kinda like part of the API, but aren't part of it is the extra noise to me.

Pretty low priority just because most of the time what is meant to be externally used is pretty obvious without the modifiers.

accord-core/src/main/java/accord/coordinate/ReadCoordinator.java
aweisberg: When I searched I missed `AbstractTracker` returning it. Sorry!

accord-core/src/main/java/accord/coordinate/Recover.java
belliottsmith: Yeah, I understand that and it is sometimes super important for clarity. I’ve changed it here anyway as there was nothing lost with the greater specificity, just musing on why it happened - I think I have come to treat default visibility as acceptable particularly in narrowly-scoped packages, but I’m just reverse engineering (or post hoc rationalising) my actions.

accord-core/src/main/java/accord/coordinate/tracking/AbstractTracker.java
belliottsmith: No worries. Honestly I’ve vacillated over this a lot. I think it might be cleaner to return a global result from each shard’s AcceptOutcomes (since it can compute it) but I didn’t want to pollute the RequestOutcome with TryMore. I’m far from sold I got the balance right, I think it would be nicer to encapsulate the decision in AcceptOutcomes so the min is computed over the whole tracker. We could introduce an intermediate enum TrackerStatus that includes TryMore, I think that would be enough to clean this up a bit more.

accord-core/src/main/java/accord/coordinate/Invalidate.java
aweisberg: OK, I think I get what you mean. We don't want to give the callback a chance to process the error if the error came from the callback (the one passed to `Invalidate`).

The way it works now it just looks like there are a few cases where an error occurs for the first time and we don't do anything with it in `onCallbackFailure` because `isDone` what set to true somewhere in `onFailure` or `oSuccess`.

accord-core/src/main/java/accord/coordinate/CheckOn.java
aweisberg: Thanks that feels more intuitive to me.

aweisberg (approved):
+1 TY!

PR #14, opened 2022-11-17
Invalidation fixes/improvements

belliottsmith:


accord-core/src/main/java/accord/coordinate/CheckOn.java
```diff
@@ -161,25 +171,39 @@ void start()
...
+            switch (sufficientFor.propagate())
             {
                 default: throw new IllegalStateException();
-                case Invalidation:
+                case Invalidated:
                     command.commitInvalidate(safeStore);
                     break;
-                case Outcome:
+
+                case Applied:
```
aweisberg, line 187: Applied is not a return value of `Status.Known.propagate()`

accord-core/src/main/java/accord/coordinate/CheckOn.java
```diff
@@ -161,25 +171,39 @@ void start()
...
                     if (untilLocalEpoch >= full.executeAt.epoch)
                     {
                         confirm(command.commit(safeStore, maxRoute, progressKey, partialTxn, full.executeAt, partialDeps));
                         confirm(command.apply(safeStore, untilLocalEpoch, maxRoute, full.executeAt, partialDeps, full.writes, full.result));
                         break;
                     }
-                case ExecutionOrder:
+
+                case Committed:
+                case ReadyToExecute:
```
aweisberg, line 197: `ReadyToExecute` is not a return value

accord-core/src/main/java/accord/coordinate/CheckOn.java
```diff
@@ -161,25 +171,39 @@ void start()
...
                     confirm(command.commit(safeStore, maxRoute, progressKey, partialTxn, full.executeAt, partialDeps));
                     break;
-                case Definition:
+
+                case PreCommitted:
+                    command.precommit(safeStore, full.executeAt);
+                    if (!sufficientFor.definition.isKnown())
+                        break;
+
+                case Accepted:
```
aweisberg: `Accepted` is not a return value of propagate

accord-core/src/main/java/accord/coordinate/CheckOn.java
```diff
@@ -161,25 +171,39 @@ void start()
...
+                case PreAccepted:
                     command.preaccept(safeStore, partialTxn, maxRoute, progressKey);
-                case Nothing:
+                    break;
+
+                case AcceptedInvalidate:
+                case NotWitnessed:
```
aweisberg, line 210: These two aren't either. Should this use a more specific enum for this concept of what information should be propagated?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -88,14 +76,24 @@ public boolean hasBeen(Status status)
         return status().hasBeen(status);
     }
 
-    public boolean hasBeen(Known phase)
+    public boolean has(Known known)
```
aweisberg, line 82: It's not clear from the enum values which enum is in play especially now that there are quite a few. Should `has` be called something that does more heavy lifting?

accord-core/src/main/java/accord/coordinate/Coordinate.java
```diff
@@ -289,14 +168,22 @@ private void onPreAccepted()
                     }
                     else
                     {
-                        commitInvalidate(node, txnId, route, executeAt);
-                        accept(null, new Timeout(txnId, route.homeKey));
+                        node.withEpoch(executeAt.epoch, () -> {
```
aweisberg, line 171: Is it possible to somehow assert on whether we are missing these withEpoch calls?

Like we know the executeAtEpoch from inside `commitInvalidate`.

accord-core/src/main/java/accord/coordinate/FetchData.java
```diff
@@ -30,16 +33,16 @@ public static Object fetch(Known phase, Node node, TxnId txnId, RoutingKeys some
         else return fetchWithSomeKeys(phase, node, txnId, someKeys, untilLocalEpoch, callback);
     }
 
-    public static Object fetch(Known phase, Node node, TxnId txnId, AbstractRoute route, @Nullable Timestamp executeAt, long untilLocalEpoch, BiConsumer<Known, Throwable> callback)
+    public static Object fetch(Known fetch, Node node, TxnId txnId, AbstractRoute route, @Nullable Timestamp executeAt, long untilLocalEpoch, BiConsumer<Known, Throwable> callback)
```
aweisberg, line 37: Not consistent about renaming to fetch?

accord-core/src/main/java/accord/coordinate/FindRoute.java
```diff
@@ -32,7 +32,7 @@ public Result(Route route, Timestamp executeAt)
         public Result(CheckStatusOk ok)
         {
             this.route = (Route)ok.route;
-            this.executeAt = ok.saveStatus.status.compareTo(Status.Committed) >= 0 ? ok.executeAt : null;
+            this.executeAt = ok.saveStatus.status.compareTo(Status.PreCommitted) >= 0 ? ok.executeAt : null;
```
aweisberg, line 35: What exactly is this new precommitted state and what are we using it for?

accord-core/src/main/java/accord/coordinate/Invalidate.java
```diff
@@ -22,126 +22,114 @@
...
+            default: throw new AssertionError();
+            case Success:
+                invalidate();
+                break;
 
-            isDone = true;
-            callback.accept(null, new Preempted(txnId, null));
-            return;
-        }
+            case Failed:
```
aweisberg, line 130: This isn't failed so much as this is not ready to invalidate. Failure is a bit exceptional sounding. Maybe add a comment?
// This transaction is not able to be invalidated because reasons

accord-core/src/main/java/accord/coordinate/tracking/InvalidationTracker.java
```diff
@@ -0,0 +1,292 @@
...
+        {
+            return fastPathRejects > shard.fastPathElectorate.size() - shard.fastPathQuorumSize;
+        }
+
+        public boolean isPromised()
+        {
+            return promises >= shard.slowPathQuorumSize;
+        }
+
+        public boolean canPromise()
```
aweisberg: Unused

accord-core/src/main/java/accord/coordinate/Invalidate.java
```diff
@@ -151,12 +139,12 @@ private void invalidate()
                     {
                         RecoverWithRoute.recover(node, ballot, txnId, route, callback);
```
aweisberg: If recovery shouldn't loop back can we maybe have some kind of state check that the callback going into recovery isn't already a recovery callback?

accord-core/src/main/java/accord/coordinate/tracking/InvalidationTracker.java
```diff
@@ -0,0 +1,292 @@
...
+
+    private int promisedShard = -1;
+    private boolean rejectsFastPath;
+    private boolean hasFailedShard;
+    public InvalidationTracker(Topologies topologies)
+    {
+        super(topologies, InvalidationShardTracker[]::new, InvalidationShardTracker::new);
+    }
+
+    public int promisedShardIndex()
```
aweisberg: Unused?

accord-core/src/main/java/accord/coordinate/tracking/InvalidationTracker.java
```diff
@@ -0,0 +1,292 @@
...
+package accord.coordinate.tracking;
+
+import accord.api.RoutingKey;
+import accord.local.Node;
+import accord.topology.Shard;
+import accord.topology.Topologies;
+
+import static accord.coordinate.tracking.AbstractTracker.ShardOutcomes.*;
+
+public class InvalidationTracker extends AbstractTracker<InvalidationTracker.InvalidationShardTracker, Node.Id>
```
aweisberg, line 28: I reviewed this and I think it does what it is supposed to do which is figure out if there is a shard, any shard, that will reject the fast path and has given us sufficient promises to propose invalidation.

Also apparently collecting a quorum of promises from all the shards is good enough which I didn't understand.

accord-core/src/main/java/accord/coordinate/tracking/InvalidationTracker.java
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    public boolean isPromised()
+    {
+        return promisedShard >= 0;
+    }
+
+    public boolean isSafeToInvalidate(boolean contactedAllShards)
+    {
+        return rejectsFastPath || (contactedAllShards && !hasFailedShard);
```
aweisberg: If we contacted all shards and they all promised to invalidate how do we know there wasn't a fast path preaccept and actually we can't invalidate?

preacceptInvalidate checks nothing but the Ballot we provided?

accord-core/src/main/java/accord/coordinate/RecoverWithRoute.java
```diff
@@ -91,43 +101,47 @@ protected void onDone(Done done, Throwable failure)
...
-                assert merged.executeAt != null;
+
+            case OutcomeApplied:
+            case OutcomeKnown:
+                Preconditions.checkState(known.definition.isKnown());
+                Preconditions.checkState(known.order.hasExecuteAt());
                 // TODO: we might not be able to reconstitute Txn if we have GC'd on some shards
                 Txn txn = merged.partialTxn.reconstitute(route);
-                if (merged.committedDeps.covers(route))
+                if (known.order.hasDeps())
```
aweisberg: Is it really better not to assert this on the actual deps?

accord-core/src/main/java/accord/coordinate/tracking/RecoveryTracker.java
```diff
@@ -0,0 +1,75 @@
...
+
+import static accord.coordinate.tracking.AbstractTracker.ShardOutcomes.*;
+
+public class RecoveryTracker extends AbstractTracker<RecoveryTracker.RecoveryShardTracker, Node.Id>
+{
+    public static class RecoveryShardTracker extends QuorumShardTracker
+    {
+        protected int fastPathRejects = 0;
+
+        public RecoveryShardTracker(Shard shard)
```
aweisberg: Why did these 3 need to become public?

accord-core/src/main/java/accord/local/Status.java
```diff
@@ -131,8 +290,45 @@ public boolean hasBeen(Status equalOrGreaterThan)
...
+
+        return max;
+    }
+
+    public static <T> T max(T a, Status statusA, Ballot acceptedA, T b, Status statusB, Ballot acceptedB)
+    {
+        int c = statusA.phase.compareTo(statusB.phase);
+        if (c > 0) return a;
+        if (c < 0) return b;
+        if (statusA.phase != Phase.Accept || acceptedA.compareTo(acceptedB) >= 0)
```
aweisberg, line 409: What is this bit about A's phase not being accept mattering? Don't A and B have the same phase at this point?

accord-core/src/main/java/accord/local/Status.java
```diff
@@ -131,8 +290,45 @@ public boolean hasBeen(Status equalOrGreaterThan)
...
+        for (T item : list)
+        {
+            if (!filter.test(item))
+                continue;
+
+            Status status = getStatus.apply(item);
+            Ballot accepted = getAccepted.apply(item);
+            boolean update = max == null
+                          || maxStatus.phase.compareTo(status.phase) < 0
+                          || (status.phase.equals(Phase.Accept) && maxAccepted.compareTo(accepted) < 0);
```
aweisberg, line 391: Why is the `Ballot` from the `Accept` phase important vs a ballot from a phase later than accept?

accord-core/src/main/java/accord/topology/Topologies.java
```diff
@@ -298,10 +308,18 @@ public boolean contains(Id to)
             return false;
         }
 
+        @Override
+        public int estimateUniqueNodes()
```
aweisberg: Since `nodes()` calculates this precisely should we remember that value and then keep returning it?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -351,13 +369,14 @@ public enum CommitOutcome { Success, Redundant, Insufficient }
     // relies on mutual exclusion for each key
     public CommitOutcome commit(SafeCommandStore safeStore, AbstractRoute route, @Nullable RoutingKey progressKey, @Nullable PartialTxn partialTxn, Timestamp executeAt, PartialDeps partialDeps)
     {
-        if (hasBeen(Committed))
+        if (hasBeen(PreCommitted))
```
aweisberg, line 364: What is going on here with Precommitted vs Committed?

aweisberg (changes requested):
I still need to tie together a couple things in my understanding. Why PreCommitted needed to be introduced and how this is ensuring we don't enter recovery repeatedly.

I get you are plugging the hole where invalidate didn't try hard enough to invalidate in some cases I just need to understand the mechanics of why it failed.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: I decided to just make sure all were handled.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: We could introduce a dedicated enum, I just felt like we had enough enums.

accord-core/src/main/java/accord/local/Command.java
belliottsmith: Naming is always hard. I figure `has` is fine to overload here, it just means "do we have at least as much information as this thing" - happy to accept proposals for alternatives

accord-core/src/main/java/accord/coordinate/Coordinate.java
belliottsmith: Yes, I can add some asserts. Though we will fail if we don't have them, with or without asserts.

accord-core/src/main/java/accord/coordinate/FindRoute.java
belliottsmith: I believe I commented it, and perhaps there is a better name out there. It is the committed `executeAt` only, i.e. the `Committed` state but without the associated `Deps` necessary for execution

accord-core/src/main/java/accord/coordinate/Invalidate.java
belliottsmith: Is there a clean way to do that?

Also, it isn't true that we can never loop back, just that it cannot be infinite. We can:
1. attempt Invalidate with one random non-home shard
2. find enough data to attempt recovery on the home shard
3. find insufficient information on the home shard, so attempt invalidation of the home shard
4. race and now find sufficient information on the home shard to attempt recovery
5. find some other shard did not reach consensus, so go to invalidate - this time with the full route
6. invalidate across all shards

accord-core/src/main/java/accord/coordinate/tracking/InvalidationTracker.java
belliottsmith: > collecting a quorum of promises from all the shards is good enough which I didn't understand.

No, as discussed on Slack this is incorrect and I'm already working on undoing this.

accord-core/src/main/java/accord/coordinate/tracking/InvalidationTracker.java
belliottsmith: As discussed, this is a mistake - it should just check `rejectsFastPath`

accord-core/src/main/java/accord/coordinate/RecoverWithRoute.java
belliottsmith: We do, when we compute `known` above

accord-core/src/main/java/accord/coordinate/tracking/RecoveryTracker.java
belliottsmith: They probably did not.

accord-core/src/main/java/accord/local/Status.java
belliottsmith: If both are in the `Accept` phase we want the `Status` with the highest ballot.

accord-core/src/main/java/accord/local/Status.java
belliottsmith: Because only in the `Accept` phase does the `Ballot` determine precedence.

accord-core/src/main/java/accord/topology/Topologies.java
belliottsmith: I don't follow

accord-core/src/main/java/accord/local/Command.java
belliottsmith: If we've already been `PreCommitted` our `executeAt` should already be correct, so we just check it.

accord-core/src/main/java/accord/topology/Topologies.java
aweisberg: I am overthinking it. I think we always invoke `nodes` to do something useful with the `Topologies` and that calculates the precise number of unique nodes and we could then start returning it.

But's so rarely used doesn't seem worth the complexity.

accord-core/src/main/java/accord/local/Status.java
aweisberg: For my own understanding In phases after accept is it supposed to be an an invariant that the ballot doesn't change anymore?

I started asking questions about special casing because I was naively expecting to just blindly compare on ballot when phase is `==`.

accord-core/src/main/java/accord/topology/Topologies.java
belliottsmith: You’re right, though I think I’d be inclined to just precompute nodes and stash it somewhere instead in that case. I’d be ok with that.

accord-core/src/main/java/accord/local/Status.java
belliottsmith: Well technically no, you could still seek promises and accepts for newer ballots, the invariant is that they would be logically equivalent as the node doing that would always undertake the same action as any node with an earlier ballot that entered those states.

accord-core/src/main/java/accord/coordinate/FindRoute.java
aweisberg: I didn't see a comment in `Status`. It makes sense.

I was able to figure it out just from how it is only called in `CheckOn`. Is what is happening we find new states a transaction can be in and we keep adding representation and handling for them?

accord-core/src/main/java/accord/coordinate/Invalidate.java
aweisberg: From the description we should only ever invalidate on a random shard once, we should only attempt to invalidate on the home shard once, and then finally we might invalidate across all shards.

Maybe the invariant is a limit on not invalidating a txn using the same inputs twice? If we fail to invalidate it is because we switch to recovery or invalidating on a different set of shards.

Only if you think it would be useful at preventing future regressions.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: (I decided to just make sure all were handled.)

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: (I decided to just make sure all were handled.)

accord-core/src/main/java/accord/coordinate/Invalidate.java
belliottsmith: Yes, I think that would be a valid invariant. I'm not sure of a good way to enforce that, though? We could propagate the parameters most recently used for `Invalidate` (that are `@Nullable`) and ensure we are using strictly more information each time?

accord-core/src/main/java/accord/coordinate/tracking/InvalidationTracker.java
belliottsmith: This should be fixed

accord-core/src/main/java/accord/coordinate/tracking/InvalidationTracker.java
belliottsmith: This should also be fixed

accord-core/src/main/java/accord/coordinate/Coordinate.java
aweisberg: It's fine not to add asserts if it will fail immediately. Only reason to add asserts would be to make it fail when the wrong state/parameters are provided rather then later when they are used.

accord-core/src/main/java/accord/coordinate/Invalidate.java
aweisberg: FYI this is still present. I think it really is worth mentioning that failure here can be something exceptional or it can be something like the transaction has made to much progress to be invalidated.

accord-core/src/main/java/accord/coordinate/Invalidate.java
aweisberg: Yeah. As long as we catch it eventually it's fine if we repeat a few useless attempts and converge on the maximum amount of what we can provide to invalidate. It's mostly just to increase visibility when the logic is wrong.

accord-core/src/main/java/accord/coordinate/Invalidate.java
belliottsmith: I don't think it can be? Failed here means we literally had too many failures, or failed to get enough promises. I'll comment to this effect, anyway

accord-core/src/main/java/accord/coordinate/Invalidate.java
belliottsmith: Okay, in the process of addressing this I changed the semantics slightly. We now only reach `Success` in `InvalidationTracker` if we can reject the fast path OR have promises at every shard. This permits us to guarantee forward progress by only passing forwards the status we encountered in the first invalidation: if it was a recoverable status, then we fail if we cannot find enough information to recover. If it was not a recoverable status, we inform the recursive invalidation that it is safe to invalidate if it does not itself find a recoverable status, since its promises _combined with_ the failure to find enough information to recover _post_ these promises being made, guarantees that the original coordinator cannot reach a fast path decision (and also will abandon its coordination, as it has been preempted by us).

In principle we could weaken this, as _any_ shard promising _while also_ having insufficient information to recover would suffice. But it is simpler to just require that every shard we use for recovery is promised, so we know that if any shard we later contact has insufficient information then it's guaranteed to be safe to invalidate (unless we find a recoverable status, i.e. Accept or later, in which case we are guaranteed to find enough information to invoke recovery)

aweisberg: Sounds good.

aweisberg (approved):
+1 TY!

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -979,6 +985,9 @@ else if (existingRanges != additionalRanges && !route().covers(additionalRanges)
         if (!validate(ensurePartialTxn, existingRanges, additionalRanges, covers(partialTxn()), covers(partialTxn), "txn", partialTxn))
             return false;
 
+        if (partialTxn != null && kind() != null && !kind().equals(partialTxn.kind()))
+            throw new IllegalArgumentException("Transaction has different kind to the definition we previously received");
```
aweisberg, line 989: What was the bug here? How could kind change during the life of a txn?

accord-core/src/main/java/accord/impl/InMemoryCommandsForKey.java
```diff
@@ -79,8 +78,8 @@ public boolean isEmpty()
         public Stream<T> before(@Nonnull Timestamp timestamp, @Nonnull TestKind testKind, @Nonnull TestDep testDep, @Nullable TxnId depId, @Nonnull TestStatus testStatus, @Nullable Status status)
         {
             return commands.headMap(timestamp, false).values().stream()
-                    .filter(cmd -> testKind == RorWs || cmd.kind() == WRITE)
-                    .filter(cmd -> testDep == ANY_DEPS || cmd.known().deps == DepsUnknown || (cmd.partialDeps().contains(depId) ^ (testDep == WITHOUT)))
+                    .filter(cmd -> testKind == RorWs || cmd.kind() == WRITE) //
```
aweisberg: Comment isn't present?

accord-core/src/main/java/accord/impl/InMemoryCommandsForKey.java
```diff
@@ -79,8 +78,8 @@ public boolean isEmpty()
...
+                    .filter(cmd -> testDep == ANY_DEPS || (cmd.known().deps != DepsUnknown && (cmd.partialDeps().contains(depId) ^ (testDep == WITHOUT))))
```
aweisberg, line 83: Either they didn't care what the dependencies were, or they cared, and we won't show them a txn with unknown dependencies, or we have dependencies and we check as requested either it isn in the dependencies or is not in the dependencies).

When we say the dependencies for a txn are known and it's a partial dependencies we really mean we just know the part that we cover in some range of epochs?

accord-core/src/main/java/accord/local/Command.java
belliottsmith: Cassandra side tries to serialise with the kind, which may be known by Accept without the transaction definition being known, so we need to persist it separately. This is just an internal consistency check now we're storing it.

I would like to _very soon_ follow-up by refactoring `TxnId` to persist the `Kind` as this is much simpler and avoids any of these problems. But we have to draw a line somewhere for any given patch.

accord-core/src/main/java/accord/impl/InMemoryCommand.java
```diff
@@ -41,7 +41,7 @@ public class InMemoryCommand extends Command
     private PartialTxn partialTxn;
     private Ballot promised = Ballot.ZERO, accepted = Ballot.ZERO;
     private Timestamp executeAt;
-    private PartialDeps partialDeps = PartialDeps.NONE;
+    private @Nullable PartialDeps partialDeps = null;
```
aweisberg, line 46: So PartialDeps.NONE could be confused with something else? Like empty dependencies?

PR #15, opened 2022-11-17
Routables refactor

belliottsmith:


belliottsmith:
One issue with this refactor is that it exposes us potentially quite extensively to the "type pollution" bug in the JVM whereby instanceof and checked casts may be costly.

Now, at the very least in C* this will be somewhat mitigated by having `RoutingKey` and `Key` split their hierarchy - though I believe we currently implement that with a shared interface as well, we can perhaps migrate to a shared abstract class.
But that still leaves 2-3 interfaces we may regularly perform checked-casts on, most notably as part of generic type parameters, e.g. in Sortables, Routables, Route, etc.

MOST of these locations SHOULD be eliminated by HotSpot, and I will take a look at some of the call-sites via print assembly. But we are unlikely to readily eliminate them all.

```
1:	accord.impl.IntHashKey
Count:	426454
Types:
	accord.primitives.SortableKey
	accord.api.RoutingKey
	accord.api.Key
	accord.primitives.Routable
	accord.primitives.Seekable
Traces:
	accord.impl.IntHashKey.compareTo(IntHashKey.java:35)
		class: accord.primitives.SortableKey
		count: 156796
	accord.primitives.KeyRoute.<init>(KeyRoute.java:17)
		class: accord.api.RoutingKey
		count: 118564
	accord.impl.list.ListUpdate.slice(ListUpdate.java:57)
		class: accord.api.Key
		count: 38289
	accord.local.Node.trySelectProgressKey(Node.java:440)
		class: accord.primitives.Routable
		count: 34471
	accord.primitives.Deps.slice(Deps.java:466)
		class: accord.api.Key
		count: 29730
	accord.impl.list.ListUpdate.slice(ListUpdate.java:56)
		class: accord.primitives.SortableKey
		count: 13098
	accord.primitives.Range$EndInclusive.compareTo(Range.java:36)
		class: accord.primitives.SortableKey
		count: 11565
	accord.primitives.Deps.toRoutables(Deps.java:1001)
		class: accord.api.Key
		count: 6177
	accord.impl.list.ListQuery.compute(ListQuery.java:50)
		class: accord.api.Key
		count: 4598
	accord.burn.BurnTest.randomKey(BurnTest.java:108)
		class: accord.api.Key
		count: 3000
	accord.burn.BurnTest.lambda$burn$3(BurnTest.java:218)
		class: accord.api.Key
		count: 2238
	accord.local.CommandStore.hashIntersects(CommandStore.java:100)
		class: accord.primitives.SortableKey
		count: 1987
	accord.burn.BurnTest.generate(BurnTest.java:92)
		class: accord.api.Key
		count: 1929
	accord.impl.list.ListUpdate.apply(ListUpdate.java:46)
		class: accord.api.Key
		count: 1877
	accord.local.Node.trySelectHomeKey(Node.java:401)
		class: accord.primitives.Seekable
		count: 1645
	accord.impl.InMemoryCommandStore$State.lambda$commandsForKey$1(InMemoryCommandStore.java:102)
		class: accord.api.Key
		count: 458
	accord.coordinate.Invalidate.invalidate(Invalidate.java:167)
		class: accord.primitives.Routable
		count: 18
	accord.impl.SimpleProgressLog$BlockingState.invalidate(SimpleProgressLog.java:477)
		class: accord.primitives.Routable
		count: 14
```

belliottsmith:
Splitting `IntHashKey` into `Key` and `Hash` gives:

```
--------------------------
Type Pollution Statistics:
--------------------------
1:	accord.impl.IntHashKey$Hash
Count:	163599
Types:
	accord.api.RoutingKey
	accord.primitives.SortableKey
	accord.primitives.Routable
Traces:
	accord.primitives.KeyRoute.<init>(KeyRoute.java:17)
		class: accord.api.RoutingKey
		count: 71435
	accord.impl.IntHashKey$Hash.compareTo(IntHashKey.java:49)
		class: accord.primitives.SortableKey
		count: 62294
	accord.local.Node.trySelectProgressKey(Node.java:440)
		class: accord.primitives.Routable
		count: 28382
	accord.impl.IntHashKey$Key.compareTo(IntHashKey.java:36)
		class: accord.primitives.SortableKey
		count: 1454
	accord.coordinate.Invalidate.invalidate(Invalidate.java:167)
		class: accord.primitives.Routable
		count: 23
	accord.impl.SimpleProgressLog$BlockingState.invalidate(SimpleProgressLog.java:477)
		class: accord.primitives.Routable
		count: 9
	accord.primitives.Range$EndInclusive.compareTo(Range.java:36)
		class: accord.primitives.SortableKey
		count: 2
--------------------------
2:	accord.impl.IntHashKey$Key
Count:	140069
Types:
	accord.primitives.SortableKey
	accord.api.Key
	accord.primitives.Seekable
Traces:
	accord.impl.IntHashKey$Key.compareTo(IntHashKey.java:36)
		class: accord.primitives.SortableKey
		count: 61756
	accord.impl.list.ListUpdate.slice(ListUpdate.java:57)
		class: accord.api.Key
		count: 38958
	accord.primitives.Deps.slice(Deps.java:466)
		class: accord.api.Key
		count: 14798
	accord.primitives.Range$EndInclusive.compareTo(Range.java:36)
		class: accord.primitives.SortableKey
		count: 5518
	accord.impl.list.ListQuery.compute(ListQuery.java:50)
		class: accord.api.Key
		count: 4478
	accord.burn.BurnTest.randomKey(BurnTest.java:108)
		class: accord.api.Key
		count: 3009
	accord.burn.BurnTest.lambda$burn$3(BurnTest.java:218)
		class: accord.api.Key
		count: 2267
	accord.burn.BurnTest.generate(BurnTest.java:92)
		class: accord.api.Key
		count: 1969
	accord.impl.list.ListUpdate.slice(ListUpdate.java:56)
		class: accord.primitives.SortableKey
		count: 1943
	accord.impl.list.ListUpdate.apply(ListUpdate.java:46)
		class: accord.api.Key
		count: 1896
	accord.local.Node.trySelectHomeKey(Node.java:401)
		class: accord.primitives.Seekable
		count: 1495
	accord.primitives.Deps.toRoutables(Deps.java:1001)
		class: accord.api.Key
		count: 822
	accord.local.CommandStore.hashIntersects(CommandStore.java:100)
		class: accord.primitives.SortableKey
		count: 722
	accord.impl.InMemoryCommandStore$State.lambda$commandsForKey$1(InMemoryCommandStore.java:102)
		class: accord.api.Key
		count: 433
	accord.impl.IntHashKey$Hash.compareTo(IntHashKey.java:49)
		class: accord.primitives.SortableKey
		count: 5
```

belliottsmith:
This latest refactor has a significant impact on reducing checkcasts, but I'm not convinced that polluting the type hierarchy here is actually the best solution. I'm inclined to suggest we use byteweaving to replace the checkcasts in the synthetic bridge methods with checkcasts for the concrete type we expect.

```
--------------------------
Type Pollution Statistics:
--------------------------
1:	accord.impl.IntHashKey$Key
Count:	101098
Types:
	accord.primitives.SortableKey
	accord.api.Key
	accord.primitives.Seekable
Traces:
	accord.impl.list.ListUpdate.slice(ListUpdate.java:57)
		class: accord.api.Key
		count: 41227
	accord.impl.list.ListUpdate.slice(ListUpdate.java:56)
		class: accord.primitives.SortableKey
		count: 37146
	accord.primitives.Range$EndInclusive.compareTo(Range.java:36)
		class: accord.primitives.SortableKey
		count: 10105
	accord.primitives.Deps.slice(Deps.java:466)
		class: accord.api.Key
		count: 7013
	accord.local.CommandStore.hashIntersects(CommandStore.java:100)
		class: accord.primitives.SortableKey
		count: 2853
	accord.local.Node.trySelectHomeKey(Node.java:401)
		class: accord.primitives.Seekable
		count: 1465
	accord.impl.list.ListQuery.compute(ListQuery.java:50)
		class: accord.api.Key
		count: 968
	accord.primitives.Deps.toRoutables(Deps.java:1001)
		class: accord.api.Key
		count: 259
	accord.impl.InMemoryCommandStore$State.lambda$commandsForKey$1(InMemoryCommandStore.java:102)
		class: accord.api.Key
		count: 59
	accord.burn.BurnTest.randomKey(BurnTest.java:108)
		class: accord.api.Key
		count: 3
--------------------------
2:	accord.impl.IntHashKey$Hash
Count:	49963
Types:
	accord.api.RoutingKey
	accord.primitives.Routable
	accord.primitives.SortableKey
Traces:
	accord.primitives.KeyRoute.<init>(KeyRoute.java:17)
		class: accord.api.RoutingKey
		count: 24981
	accord.local.Node.trySelectProgressKey(Node.java:440)
		class: accord.primitives.Routable
		count: 24958
	accord.coordinate.Invalidate.invalidate(Invalidate.java:167)
		class: accord.primitives.Routable
		count: 14
	accord.impl.SimpleProgressLog$BlockingState.invalidate(SimpleProgressLog.java:477)
		class: accord.primitives.Routable
		count: 7
	accord.primitives.Range$EndInclusive.compareTo(Range.java:36)
		class: accord.primitives.SortableKey
		count: 3
```

belliottsmith:
I wonder if we should create a tool to compile a version of the lib that replaces all checkcast of each kind of interface with a matching concrete class from the implementation. It won't work for everything, as we have a split type hierarchy, but it would help resolve a significant number of these.

belliottsmith:
Hmm, in fairness, looking at the remaining traces for `Key`/`RoutingKey` after the final refactor, it looks like must of it would be avoidable by the implementation.

accord-core/src/main/java/accord/api/Key.java
```diff
@@ -18,10 +18,12 @@
 
 package accord.api;
 
+import accord.primitives.SortableKey;
+import accord.primitives.Seekable;
+
 /**
  * A routing key for determining which shards are involved in a transaction
```
aweisberg: Should this comment be updated?

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -171,7 +169,7 @@ public long all()
             return -1L >>> (64 - shards.length);
         }
 
-        public long shards(AbstractKeys<?, ?> keys, long minEpoch, long maxEpoch)
+        public long shards(AbstractRoutableKeys<?> keys, long minEpoch, long maxEpoch)
```
aweisberg: Unused?

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -181,17 +179,59 @@ public long shards(AbstractKeys<?, ?> keys, long minEpoch, long maxEpoch)
...
+                    {
+                        // include every shard if we match a range
+                        accumulate = Sortables.foldl((Ranges)keysOrRanges, ranges[i], (idx, k, p, a) -> p, terminalValue, accumulate, terminalValue);
+                    }
+                    return accumulate;
+                }
+            }
+        }
+
+        long shard(RoutingKey scope, long minEpoch, long maxEpoch)
```
aweisberg: Unused?

accord-core/src/main/java/accord/local/PreLoadContext.java
```diff
@@ -38,37 +39,47 @@
...
     }
 
-    static PreLoadContext contextFor(TxnId txnId, Iterable<Key> keys)
+    static PreLoadContext contextFor(TxnId txnId, Seekables<?, ?> keysOrRanges)
     {
-        return contextFor(Collections.singleton(txnId), keys);
+        switch (keysOrRanges.kindOfContents())
+        {
+            default: throw new AssertionError();
+            case Range: return contextFor(txnId); // TODO (soon): this won't work for actual range queries
```
aweisberg, line 61: Why not store the `Seekables` so they can be loaded?

accord-core/src/main/java/accord/primitives/AbstractKeys.java
```diff
@@ -2,24 +2,22 @@
...
-import accord.utils.SortedArrays;
+import accord.utils.*;
 import net.nicoulaj.compilecommand.annotations.Inline;
 
+import static accord.primitives.Sortable.Kind.Key;
+
 @SuppressWarnings("rawtypes")
 // TODO: check that foldl call-sites are inlined and optimised by HotSpot
-public abstract class AbstractKeys<K extends RoutingKey, KS extends AbstractKeys<K, KS>> implements Iterable<K>
+public abstract class AbstractKeys<K extends SortableKey, KS extends Sortables<K, ?>> implements Iterable<K>, Sortables<K, KS>
```
aweisberg:
```suggestion
// AbstractKeys should be avoided in favor of derived classes because it doesn't restrict to a single kind of key (Routable, Seekable)
public abstract class AbstractKeys<K extends SortableKey, KS extends Sortables<K, ?>> implements Iterable<K>, Sortables<K, KS>
```

Similar comment is also applicable to `Sortable` and `Sortables`.

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -50,9 +50,9 @@ CommandStore create(int id,
...
     public interface RangesForEpoch
     {
-        KeyRanges at(long epoch);
-        KeyRanges between(long fromInclusive, long toInclusive);
-        KeyRanges since(long epoch);
+        Ranges at(long epoch);
+        Ranges between(long fromInclusive, long toInclusive);
+        Ranges since(long epoch);
         boolean owns(long epoch, RoutingKey key);
         boolean intersects(long epoch, AbstractKeys<?, ?> keys);
```
aweisberg: Unused?

accord-core/src/main/java/accord/primitives/PartialTxn.java
```diff
@@ -8,31 +8,21 @@
...
-
-        return covering().containsAll(keys);
-    }
-
-    default boolean covers(AbstractRoute route)
-    {
-        if (query() == null && route.contains(route.homeKey))
-            return false;
-        return covering().containsAll(route);
+        if (query() == null)
```
aweisberg, line 19: I think this needs an explanation/comment. My take is that if this `PartialTxn` doesn't have the `query` we don't know if the `query` was never set for this `Txn` or if we just don't have it because the `Routable` doesn't contain the home key so we don't know what the home shard would know about the nullity of `query`.

accord-core/src/main/java/accord/primitives/Range.java
```diff
@@ -32,17 +31,17 @@
 /**
  * A range of keys
  */
-public abstract class KeyRange implements Comparable<RoutingKey>
+public abstract class Range implements Comparable<SortableKey>, Routable, Seekable
```
aweisberg: - [ ] For keys you seem to have quite successfully separated `Routable` and `Seekable`. `AbstractKeys` is minimally used outside of support code that doesn't care if it's `Routable` or `Seekable` so we get the static checking we want.


`Range` doesn't seem to have that. There is no `Seekable` key that can construct `Range` currently (and don't we need users to be able to express seekable key ranges?). Yet `Range` implements `Seekable`. So I went about removing `Seekable` from `Range` and everything compiled? 

I assume this is a WIP until range transactions exist for users?

accord-core/src/main/java/accord/primitives/Routable.java
```diff
@@ -0,0 +1,10 @@
+package accord.primitives;
+
+import accord.api.RoutingKey;
```
aweisberg, line 3: Unused import

accord-core/src/main/java/accord/primitives/Routables.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Either a Route or a collection of Routable
+ */
+public interface Routables<K extends Routable, U extends Routables<K, ?>> extends Iterable<K>, Sortables<K, U>
+{
+    enum RoutablesKind
+    {
+        RoutingKeys, PartialKeyRoute, FullKeyRoute, RoutingRanges, PartialRangeRoute, FullRangeRoute;
+        public boolean isRoute()
```
aweisberg: Why is `RoutingRanges` a `Route`?

accord-core/src/main/java/accord/primitives/Routables.java
```diff
@@ -0,0 +1,69 @@
...
+        {
+            switch (this)
+            {
+                default: throw new AssertionError();
+                case FullKeyRoute:
+                case FullRangeRoute:
+                    return true;
+                case RoutingKeys:
+                case PartialKeyRoute:
+                case RoutingRanges:
```
aweisberg: It's just `Ranges` as a class name. The differentiation between `Routable` and `Seekable` ranges doesn't exist yet.

accord-core/src/main/java/accord/primitives/Seekable.java
```diff
@@ -0,0 +1,5 @@
+package accord.primitives;
+
+public interface Seekable extends Sortable
```
aweisberg: Both `Seekable` and `SortableKey` extend `Sortable`. Maybe only one should?

accord-core/src/main/java/accord/primitives/Sortable.java
```diff
@@ -0,0 +1,14 @@
...
+import accord.api.RoutingKey;
+
+/**
+ * Either a RoutingKey, Key or a Range
+ */
+public interface Sortable
+{
+    enum Kind { Key, Range }
+    Kind kind();
+    Routable toRoutable();
```
aweisberg: So generally when trying to understand the relationship between routable and seekables, it's always possible to convert a `Seekable` to a `Routable`, and a `Routable` is always a `Routable` and I had the theory that the static typing is so consistent that we don't actually need `toRoutable` here and it only needs to be in `Seekable`.

This is after I already removed `Seekable` as an interface of `Range`. I did the move and cleaned things up and it compiles and it makes sense to me that we would only ever need to convert a `Seekable` to a `Routable` because in practice we always know if what we have is a `Seekable`s or `Routable`s.

accord-core/src/main/java/accord/primitives/Sortable.java
```diff
@@ -0,0 +1,14 @@
...
+    RoutingKey someIntersectingRoutingKey();
```
aweisberg: This doesn't necessarily seem like it belongs in `Sortable`, but I tried to find a way to move it or maybe have different versions for `Seekable` and `Sortable` with limited success.

accord-core/src/main/java/accord/primitives/Sortables.java
```diff
@@ -0,0 +1,257 @@
...
+    boolean isEmpty();
+    boolean intersects(AbstractRanges<?> ranges);
+    boolean contains(SortableKey key);
+    boolean containsAll(Sortables<?, ?> keysOrRanges);
+
+    U slice(Ranges ranges);
+    Sortables<K, U> union(U with);
+
+    /**
+     * @return an arbitrary key that relates to this Findable - this may or may not be contained by this Findable,
```
aweisberg: `Findable` isn't a thing anymore?

accord-core/src/main/java/accord/primitives/Sortables.java
```diff
@@ -0,0 +1,257 @@
...
+
+    @Inline
+    static <Input extends Sortable> long rangeFoldl(Sortables<Input, ?> inputs, AbstractKeys<?, ?> matching, IndexedRangeFoldToLong fold, long param, long initialValue, long terminalValue)
+    {
+        return Helper.rangeFoldl(Sortables::findNextIntersection, (ls, li, rs, ri) -> li + 1,
+                inputs, matching, fold, param, initialValue, terminalValue);
+    }
+
+    @Inline
+    static <Input extends Sortable> long foldlMissing(Sortables<Input, ?> inputs, Sortables<Input, ?> notMatching, IndexedFoldToLong<? super Input> fold, long param, long initialValue, long terminalValue)
```
aweisberg: These look newish, I know there are a lot of `foldl` implementations, is this worth adding unit tests?

accord-core/src/main/java/accord/topology/Topology.java
```diff
@@ -229,56 +237,92 @@ public Topology forKeys(AbstractKeys<?, ?> select, Collection<Id> nodes, Indexed
         return new Topology(epoch, shards, ranges, nodeLookup, rangeSubset, newSubset);
     }
 
-    private int[] subsetForKeys(AbstractKeys<?, ?> select, IndexedPredicate<Shard> predicate)
+    private int[] subsetFor(Routables<?, ?> select, IndexedPredicate<Shard> predicate)
```
aweisberg: Is this worth unit testing explicitly?

accord-core/src/main/java/accord/topology/Topology.java
```diff
@@ -229,56 +237,92 @@ public Topology forKeys(AbstractKeys<?, ?> select, Collection<Id> nodes, Indexed
...
+                bi = (int)(abi >>> 32);
+                if (ailim < (int)abi)
+                    throw new IllegalArgumentException("Range not found for " + select.get(ailim));
+
+                if (predicate.test(bi, shards[bi]))
+                    newSubset[count++] = bi;
+
+                ai = (int)abi;
+                ailim = as.findNext(ai + 1, bs.get(bi), FLOOR);
+                if (ailim < 0) ailim = -1 - ailim;
```
aweisberg, line 271: What is this trying to do here?

From the docs
```
     * If no entries match, similar to Arrays.binarySearch return either:
     *  FAST, CEIL: the entry following {@code find}, i.e. -1 - insertPos (== Arrays.binarySearch)
     *  FLOOR:      the entry preceding {@code find}, i.e. -2 - insertPos
```

accord-core/src/main/java/accord/primitives/AbstractRanges.java
```diff
@@ -0,0 +1,544 @@
...
+    @Override
+    public int findNext(int thisIndex, Range find, SortedArrays.Search search)
+    {
+        return SortedArrays.exponentialSearch(ranges, thisIndex, size(), find, Range::compareIntersecting, search);
+    }
+
+    @Override
+    public FullRoute<Range> toRoute(RoutingKey homeKey)
+    {
+        return new FullRangeRoute(homeKey, ranges);
```
aweisberg: Doesn't seem very safe since this may be a partial route? Similar unsafety on the key side. We seem to only do this once when a `Txn` is first started.

accord-core/src/main/java/accord/primitives/Route.java
```diff
@@ -3,50 +3,99 @@
...
+                return (FullKeyRoute) routables;
+            case FullRangeRoute:
+                return (FullRangeRoute) routables;
+        }
     }
 
-    @Override
-    public PartialRoute sliceStrict(KeyRanges ranges)
+    // this method exists solely to circumvent JDK bug with testing and casting interfaces
+    static PartialRoute<?> castToPartialRoute(@Nullable Routables<?, ?> routables)
```
aweisberg: Unused?

accord-core/src/main/java/accord/coordinate/FindRoute.java
```diff
@@ -20,26 +18,31 @@ public class FindRoute extends CheckShards
...
     }
 
     final BiConsumer<Result, Throwable> callback;
     FindRoute(Node node, TxnId txnId, RoutingKey homeKey, BiConsumer<Result, Throwable> callback)
     {
-        super(node, txnId, RoutingKeys.of(homeKey), txnId.epoch, IncludeInfo.Route);
+        this(node, txnId, RoutingKeys.of(homeKey), callback);
+    }
+
+    private FindRoute(Node node, TxnId txnId, RoutingKeys homeKey, BiConsumer<Result, Throwable> callback)
```
aweisberg: Why have the additional constructor?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -301,28 +302,27 @@ public AcceptOutcome accept(SafeCommandStore safeStore, Ballot ballot, PartialRo
         }
 
         TxnId txnId = txnId();
-        KeyRanges coordinateRanges = coordinateRanges(safeStore);
-        KeyRanges executeRanges = txnId.epoch == executeAt.epoch ? coordinateRanges : safeStore.ranges().at(executeAt.epoch);
-        KeyRanges acceptRanges = txnId.epoch == executeAt.epoch ? coordinateRanges : safeStore.ranges().between(txnId.epoch, executeAt.epoch);
+        Ranges coordinateRanges = coordinateRanges(safeStore);
+//        Ranges executeRanges = txnId.epoch == executeAt.epoch ? coordinateRanges : safeStore.ranges().at(executeAt.epoch);
```
aweisberg: Remove `executeRanges`? Why don't we use the ranges at the `executeAt.epoch` anymore?

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -91,12 +90,17 @@ public int generation()
         return generation;
     }
 
-    public boolean hashIntersects(RoutingKey key)
+    public boolean hashIntersects(SortableKey key)
```
aweisberg: Sortable is non-specific? This has to be either a Seekable or a Routable right? If it's the wrong kind then it won't intersect correctly?

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -91,12 +90,17 @@ public int generation()
...
         return ShardedRanges.keyIndex(key, numShards) == shardIndex;
     }
 
-    public boolean intersects(Keys keys, KeyRanges ranges)
+    public boolean hashIntersects(Sortable routable)
+    {
+        return routable instanceof Range || hashIntersects((SortableKey) routable);
+    }
+
+    public boolean intersects(Keys keys, Ranges ranges)
```
aweisberg: When do we have `Ranges` that are ranges of `Seekables` no `Routables` (and thus work with `Keys`? I checked and this seems unused and I think is also wrong.

Also maybe `AbstractKeys.any` shouldn't compile with this attempted usage?

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -91,12 +90,17 @@ public int generation()
...
     }
 
-    public boolean hashIntersects(RoutingKey key)
+    public boolean hashIntersects(SortableKey key)
     {
         return ShardedRanges.keyIndex(key, numShards) == shardIndex;
     }
 
-    public boolean intersects(Keys keys, KeyRanges ranges)
+    public boolean hashIntersects(Sortable routable)
```
aweisberg: Same thing here. Why `Sortable`? Can it be more specific?

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -181,17 +179,59 @@ public long shards(AbstractKeys<?, ?> keys, long minEpoch, long maxEpoch)
             return accumulate;
         }
 
-        KeyRanges currentRanges()
+        public <T extends Sortable> long shards(Sortables<T, ?> keysOrRanges, long minEpoch, long maxEpoch)
```
aweisberg: A lot more `Sortable` is it possible to be more specific like `Routable`?

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -181,17 +179,59 @@ public long shards(AbstractKeys<?, ?> keys, long minEpoch, long maxEpoch)
...
+            return result;
+        }
+
+        Ranges currentRanges()
         {
             return ranges[ranges.length - 1];
         }
 
-        static long keyIndex(RoutingKey key, long numShards)
+        static long keyIndex(SortableKey key, long numShards)
```
aweisberg: `RoutableKey`?

aweisberg (changes requested):
It almost seems like a lot of the abstract support classes used in primitives for things that are similar implementation wise should be package private so you must specify if you are using the `Seekable` or `Routable` version. `AbstractKeys`, `Sortables`, `Sortable` are the ones that stand out where they are used all over when a more concrete version could be specified.

There are enough instances that commenting on each one doesn't make sense, but I didn't a text search and found it's pretty common.

Also had some questions about the direction `Range` is going in because it's not used as a `Seekable` yet, and doesn't handle it well like `Key` does with concrete routing/non-routing versions.

I am still going through to look deeper  for changes I missed the first time through, but wanted to drop this off to start.

accord-core/src/main/java/accord/local/Status.java
```diff
@@ -213,6 +213,11 @@ public boolean isKnown()
         {
             return this == DefinitionKnown;
         }
+
+        public static Definition max(Definition a, Definition b)
```
aweisberg: Unused?

accord-core/src/main/java/accord/primitives/KeyRanges.java
```diff
@@ -1,494 +1,494 @@
...
-    {
-        return new KeyRanges(ranges);
-    }
-
-    public static KeyRanges single(KeyRange range)
-    {
-        return new KeyRanges(new KeyRange[]{range});
-    }
-}
+///*
```
aweisberg: Delete the file?

accord-core/src/main/java/accord/primitives/PartialKeyRoute.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Override
+    public boolean intersects(AbstractRanges<?> ranges)
+    {
+        // TODO (now): remove this in favour of parent implementation - ambiguous at present
+        return ranges.intersects(covering);
+    }
+
+    @Override
+    public AbstractRoutableKeys<?> with(RoutingKey withKey)
```
aweisberg, line 65: Should we use `AbstractRoutableKeys` or `Routables`? Should any of these abstract classes and `Sortable` + `Sortables` from `primitives`  be used outside the package?

belliottsmith:
> It almost seems like a lot of the abstract support classes used in primitives for things that are similar implementation wise should be package private so you must specify if you are using the Seekable or Routable version. AbstractKeys, Sortables, Sortable are the ones that stand out where they are used all over when a more concrete version could be specified.

You might well be right. I'll take a look and see if any are leftover at the end. It might be a vestige of the earlier state of affairs.

> Also had some questions about the direction Range is going in because it's not used as a Seekable yet, and doesn't handle it well like Key does with concrete routing/non-routing versions.

This is because logically a `Range` is `Seekable` even if it doesn't contain `Seekable` bounds - you can seek for all `Key` between an upper and lower `RoutingKey`. That said, I haven't fully developed this side of the hierarchy because there isn't yet enough concrete need for more development that I can be confident it would survive contact with a real use case. I expect to embellish this a little as I develop range transactions for bootstrap, and further still as and when we deliver user-level range transactions.

accord-core/src/main/java/accord/local/CommandStores.java
belliottsmith: We do operate over keys here for some operations, and I think it is fine in this case to operate over either, logically, without performing a transformation. CommandStores may be partitioned by RoutingKey but the main purpose of Routables is for coordination/inter-node sharing. I am open to reconsidering this, but it felt more reasonable (as well as more efficient) to permit `CommandStores` to operate over either (and to impose stricter requirements further up the API stack, as necessary)

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: We don't, but `Ranges` of `Routables` are comparable with `Key` - both `Seekable` and `Routable` are `Sortable`. This just asks, do any of the `Key` occur in any of the `RoutingKey` bounds of the `Ranges`

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: This intersection is consistent for both kinds of key. It's also hopefully not going to survive very long, as it's a source of inefficiency and complexity.

accord-core/src/main/java/accord/primitives/PartialTxn.java
belliottsmith: `query()` should never be `null` but only the home shard gets sent it for now. So if it's `null` we haven't got enough information to send to the home shard.

belliottsmith: (I will add a comment)

accord-core/src/main/java/accord/local/PreLoadContext.java
belliottsmith: Why?

aweisberg: I guess this gets into weird stuff like is Range over Routables or Seekables. Currently it's really just over Routables. I guess it doesn't matter since what is here can't be used to support seekable ranges yet.

accord-core/src/main/java/accord/local/CommandStore.java
aweisberg: If you intersect with a command store it's checking routable intersection not seekable intersection? I want to make sure I at least understand what is going on.

accord-core/src/main/java/accord/local/CommandStore.java
aweisberg: My confusion is that a Routable is something describing something over the partitioning schema and a Seekable is describing something over some logical range of values for a key.

When would it ever make sense to mix these in collections or compare them to one another?

accord-core/src/main/java/accord/local/CommandStores.java
aweisberg: Ah, I think the piece I am missing here is that if you use Seekables the seekables are automatically translated to Routable?

I think it is 100% fine to do that. My concern was that something accepts a Sortable but really only knows how to handle a routable or seekables, but not both.

accord-core/src/main/java/accord/local/Status.java
aweisberg: Still here?

accord-core/src/main/java/accord/primitives/Routables.java
```diff
@@ -0,0 +1,257 @@
...
+    boolean isEmpty();
+    boolean intersects(AbstractRanges<?> ranges);
+    boolean contains(RoutableKey key);
+    boolean containsAll(Routables<?, ?> keysOrRanges);
+
+    U slice(Ranges ranges);
+    Routables<K, U> union(U with);
+
+    /**
+     * @return an arbitrary key that relates to this Findable - this may or may not be contained by this Findable,
```
aweisberg: Findable is not a thing anymore?

accord-core/src/main/java/accord/primitives/Route.java
```diff
@@ -1,52 +1,100 @@
...
+                return (FullKeyRoute) routables;
+            case FullRangeRoute:
+                return (FullRangeRoute) routables;
+        }
     }
 
-    @Override
-    public PartialRoute sliceStrict(KeyRanges ranges)
+    // this method exists solely to circumvent JDK bug with testing and casting interfaces
+    static PartialRoute<?> castToPartialRoute(@Nullable Unseekables<?, ?> routables)
```
aweisberg: Still here an unused?

accord-core/src/main/java/accord/local/CommandStores.java
belliottsmith: As discussed offline, `CommandStores` is happy to accept either kind of key

accord-core/src/main/java/accord/local/CommandStores.java
belliottsmith: As discussed offline, we renamed Sortable to Routable, and Routable to Unseeable, to make clear that we could use either here.

accord-core/src/main/java/accord/primitives/AbstractKeys.java
belliottsmith: As we discussed offline, sometimes you are willing to accept either

accord-core/src/main/java/accord/primitives/Routables.java
belliottsmith: It isn't. Good spot.

accord-core/src/main/java/accord/primitives/Routables.java
belliottsmith: True. I guess the question is if we prefer to use enum terminology that is forward looking. I don't mind terribly.

accord-core/src/main/java/accord/topology/Topology.java
belliottsmith: HIGHER, i.e. we find the last overlapping item (if any) then take the index following it

accord-core/src/main/java/accord/primitives/AbstractRanges.java
belliottsmith: I've moved this to `Seekables`

accord-core/src/main/java/accord/coordinate/FindRoute.java
belliottsmith: Vestige of prior structure

accord-core/src/main/java/accord/local/Command.java
belliottsmith: We don't need to adopt any information about a proposed execution epoch, only if we actually decide to use that execution epoch. For recovery only the coordination epoch matters, so this was I think cleaning up a little the earlier changes we made to stop requiring preaccept prior to accept

accord-core/src/main/java/accord/topology/Topology.java
belliottsmith: Yes, though I would prefer to defer that and some other unit test improvements until later if possible, focusing effort where most valuable.

accord-core/src/main/java/accord/primitives/Sortables.java
belliottsmith: This was cut/paste from elsewhere, with some minor edits to generify it. I agree it could do with some randomised tests, but would prefer to focus elsewhere for now as time is rapidly getting away from us.

accord-core/src/main/java/accord/primitives/Routables.java
aweisberg: It makes a lot of sense now that `Sortable` is `Routable`

aweisberg (approved):
+1 LGTM. Thank you!

PR #17, opened 2022-12-16
Async chain

bdeggleston:
This PR relates to https://github.com/apache/cassandra/pull/2057


accord-core/src/main/java/accord/utils/async/AsyncResults.java
```diff
@@ -0,0 +1,295 @@
...
+            {
+                this.callback = callback;
+            }
+        }
+
+        private void notify(Listener<V> listener, Result<V> result)
+        {
+            while (listener != null)
+            {
+                listener.callback.accept(result.value, result.failure);
```
dcapwell, line 68: With `Future` if the user function throws an exception, are able to set that as the future's result.  This logic doesn't seem to handle exceptions so `boolean trySetResult(V result, Throwable failure)` may lead to a failure on the producer side.

dcapwell: Looking closer these `Listeners` are more closely to `org.apache.cassandra.utils.concurrent.ListenerList.CallbackLambdaListener#run` which isn't explicit about the handling, so in both cases we depend on `java.lang.Thread#getUncaughtExceptionHandler`

accord-core/src/main/java/accord/utils/async/AsyncResult.java
```diff
@@ -0,0 +1,96 @@
+package accord.utils.async;
+
+import java.util.concurrent.Executor;
+import java.util.function.BiConsumer;
+
+/**
+ * Handle for async computations that supports multiple listeners and registering
+ * listeners after the computation has started
+ */
+public interface AsyncResult<V>
```
dcapwell: We spoke in slack but I feel that we should add `map` and `flatMap` here as well, so you are not forced to use the chain for single hop cases (which are 100% of the cases in Accord).

IMO we should replace `.toChain().map(...).complete()` with `.map()` and replace `.flatMap(...).flatMap(...).map(...)` with `.toChain().flatMap(...).flatMap(...).map(...).complete()`; AKA when you know you are a "chain" of calls, you call `toChain` to lower the cost.

I made this change in a feedback branch I can post, but the map/flatMap methods are

```
// feedback branch made sure null exeuctor works
default <T> AsyncResult<T> map(Function<? super V, ? extends T> mapper, Executor executor)
    {
        AsyncResult.Settable<T> settable = AsyncResults.settable();
        listen((success, failure) -> {
            if (failure != null)
            {
                settable.setFailure(failure);
                return;
            }
            try
            {
                T result = mapper.apply(success);
                settable.setSuccess(result);
            }
            catch (Throwable t)
            {
                settable.setFailure(t);
            }
        }, executor);
        return settable;
    }

    default <T> AsyncResult<T> map(Function<? super V, ? extends T> mapper)
    {
        return map(mapper, null);
    }

    default <T> AsyncResult<T> flatMap(Function<? super V, ? extends AsyncResult<T>> mapper, Executor executor)
    {
        AsyncResult.Settable<T> settable = AsyncResults.settable();
        listen((success, failure) -> {
           if (failure != null)
           {
               settable.setFailure(failure);
               return;
           }
           try
           {
               AsyncResult<T> next = mapper.apply(success);
               next.listen((s2, f2) -> {
                   if (f2 != null)
                   {
                       settable.tryFailure(f2);
                       return;
                   }
                   settable.trySuccess(s2);
               });
           }
           catch (Throwable t)
           {
               settable.setFailure(t);
           }
        }, executor);
        return settable;
    }

    default <T> AsyncResult<T> flatMap(Function<? super V, ? extends AsyncResult<T>> mapper)
    {
        return flatMap(mapper, null);
    }
```

I am totally cool with the trampoline (`AsyncChain`) but since the majority of cases are single hop we add a slightly higher cost than the raw `map` and `flatMap` methods (you can't avoid listening CAS, and the output Result has the same cost, so only difference is the cost to run the chain) I don't think we should default to use chain until we have at least 2 hops

accord-core/src/main/java/accord/utils/async/AsyncResult.java
```diff
@@ -0,0 +1,96 @@
...
+        listen((unused, failure) -> {
+            if (failure == null) runnable.run();
+            else throw new RuntimeException(failure);
+        });
+    }
+
+    default void listen(Runnable runnable, Executor executor)
+    {
+        listen(AsyncCallbacks.inExecutor(runnable, executor));
+    }
```
dcapwell: consistency comment: can we rename to `addCallback`?  `AsyncChain` and C* `Future` use `addCallback`, so would be good be match

accord-core/src/main/java/accord/coordinate/FetchData.java
```diff
@@ -77,7 +77,7 @@ private static Object fetchInternal(Ranges ranges, Known target, Node node, TxnI
     {
         long srcEpoch = executeAt == null || target.epoch() == Coordination ? txnId.epoch : executeAt.epoch;
         if (!node.topology().hasEpoch(srcEpoch))
-            return node.topology().awaitEpoch(srcEpoch).map(ignore -> fetchInternal(ranges, target, node, txnId, route, executeAt, untilLocalEpoch, callback));
+            return node.topology().awaitEpoch(srcEpoch).toChain().map(ignore -> fetchInternal(ranges, target, node, txnId, route, executeAt, untilLocalEpoch, callback)).beginAsResult();
```
dcapwell: in my feedback branch I switched from `toChain().map` to `.map`

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -93,24 +91,24 @@ public void onCallbackFailure(Id from, Throwable failure)
         }
     }
 
-    Future<Object> awaitCommits(Node node, Deps waitOn)
+    AsyncResult<Object> awaitCommits(Node node, Deps waitOn)
     {
         AtomicInteger remaining = new AtomicInteger(waitOn.txnIdCount());
-        Promise<Object> future = new AsyncPromise<>();
+        AsyncResult.Settable<Object> notifier = AsyncResults.settable();
```
dcapwell: I personally like `Promise` as it hides `[try|set][Success|Failure]` from the consumers so only producers can do this... it would be nice to have the same interface so those methods don't leak to consumers.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -287,24 +288,32 @@ public SynchronizedState(NodeTimeService time, Agent agent, DataStore store, Pro
...
-                return promise;
+                    @Override
+                    public void begin(BiConsumer<? super T, Throwable> callback)
+                    {
+                        synchronized (SynchronizedState.this)
+                        {
+                            try
+                            {
+                                T result = function.apply(SynchronizedState.this);
+                                callback.accept(result, null);
```
dcapwell: could we remove this from the sync block?  Switching from `Future` to `AsyncChain` removes all the overhead as this is working with an existing value (`this`) rather than passing around a volatile reference, but since the callback is in the sync block the whole chain's execution retains the lock...  This doesn't match the old behavior as the promise was resolved before returning (its success or failure by the time we return)

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -0,0 +1,406 @@
...
+    {
+        Preconditions.checkArgument(!chains.isEmpty());
+        return new AsyncChainCombiner.All<>(chains);
+    }
+
+    public static <V> AsyncChain<V> reduce(List<AsyncChain<V>> chains, BiFunction<V, V, V> reducer)
+    {
+        Preconditions.checkArgument(!chains.isEmpty());
+        if (chains.size() == 1)
+            return chains.get(0);
```
dcapwell, line 342: this isn't correct behavior, you need a `zero` for this case, else should reject.  There was another bug in accord where `reduce` would change the desired behavior the `reducer` requested when size 1... `reduce` should only ever work with 2+ elements and needs a `zero` to handle single element case.

See `reduce` example in `Stream` `java.util.stream.Stream#reduce(T, java.util.function.BinaryOperator<T>)`; they support single element by returning `Optional` in this case

dcapwell: by fixing this case, you can make the return be a different type than `V`, so could have

```
public static <V, R> AsyncChain<R> reduce(V zero, List<AsyncChain<V>> chains, BiFunction<V, V, R> reducer)
```

Though, if you do go that route the below logic makes things complex as types change

```
    if (Reduce.canAppendTo(chains.get(0), reducer))
        {
            AsyncChainCombiner.Reduce<V> appendTo = (AsyncChainCombiner.Reduce<V>) chains.get(0);
            appendTo.addAll(chains.subList(1, chains.size()));
            return appendTo;
        }
```

dcapwell:
Here is my feedback branch https://github.com/dcapwell/cassandra-accord/tree/feedback/async-chain

accord-core/src/main/java/accord/utils/async/AsyncResult.java
belliottsmith: I asked to remove it, for two reasons: 1) it’s a lot of code duplication; 2) it encourages precisely what we’re trying to avoid. There’s no easy way to impose a “no more than one hop” restriction.

we could introduce some syntactic sugar for constructing a chain perhaps, that is then applied here, but I do not think these use cases warrant the escape route of dedicated map methods for the AsyncResult.

accord-core/src/main/java/accord/coordinate/Recover.java
belliottsmith: That’s the same here? The AsyncResult hides it, the Settable exposes it?

accord-core/src/main/java/accord/utils/async/AsyncChains.java
belliottsmith: This isn’t a bug, it’s a choice about semantic and I don’t believe there is anything faulty about it? This is sufficient to correctly represent many “reductions”.  It’s not incompatible with a second implementation that accepts a parameter, if we want. That said, I’m not opposed to augmenting this one now, but it’s also fine without AFAICT.

bdeggleston: Right, this is about the same as `Stream#reduce`, where the reducer is only called if there is more than one result to feed it. The `Optional` returned by `Stream#reduce` is to support empty streams. This is explicitly not supported by this reduce method as you can see by the not-empty check. Everywhere this is currently used, it would be an error to not have at least one chain to reduce. I wouldn't be opposed to adding a zero parameter if that changes though.

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -93,24 +91,24 @@ public void onCallbackFailure(Id from, Throwable failure)
...
         for (int i = 0 ; i < waitOn.txnIdCount() ; ++i)
         {
             TxnId txnId = waitOn.txnId(i);
             // TODO (now): this should perhaps use RouteFragment as we might need to handle txns that are range-only
             new AwaitCommit(node, txnId, waitOn.someRoutables(txnId)).addCallback((success, failure) -> {
-                if (future.isDone())
+                if (notifier.isDone())
```
belliottsmith: result?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -730,7 +734,7 @@ public void accept(SafeCommandStore safeStore)
                 if (cur == null)
                 {
                     // need to load; schedule execution for later
-                    safeStore.execute(this, this);
+                    safeStore.execute(this, this).begin(AsyncCallbacks.throwOnFailure());
```
belliottsmith: Since this is async, what does `throwOnFailure` really mean here?

bdeggleston: currently it means that calling the callback with a failure throws an exception, presumably to be handled by the threads uncaught exception handler. Having now put that in writing, that doesn't seem like the best approach, since there are other use cases for submitting a failure (ie: in the case of accord Timeouts).

Semantically, it means "start this chain, and I don't care about the result, exception or otherwise", I'll change the name to noop and have it log exceptions

accord-core/src/main/java/accord/utils/async/AsyncResult.java
dcapwell: > There’s no easy way to impose a “no more than one hop” restriction.

Based off usage, we don't have more than 1 hop... so are we trying to optimize for a case that we do not have?  I am looking at both accord-core and the C* integration branch, and every call to `toChain` yields a single hop.  We even have fun patterns such as this

```
private AsyncChain<Void> applyChain(SafeCommandStore safeStore, boolean canReschedule)
...
return applyWithCorrectScope(safeStore.commandStore()).toChain();

// usage
private AsyncResult<Void> applyWithCorrectScope(CommandStore unsafeStore)
...
AsyncChain<Void> chain = super.applyChain(safeStore);
result = AsyncResults.forChain(chain);
```

So, because of this, we go from `AsyncResult` to `AsyncChain` back to `AsyncResult`, costing more than just passing around the single `AsyncResult` (and require more cas/volatile overhead); this pattern happens several times...

Given we don't have multiple hops this argument doesn't hold for me; what are we really trying to solve here?

> it’s a lot of code duplication;

Is it?  Looking at my feedback branch and `AsyncChains` I really don't see duplicate code... they have duplicate signatures but not bodies.

belliottsmith: Wdym more cas/volatile overhead? There should be no more overhead in this regard. If there is, it’s a mistake and we should fix it.

Slightly different code performing the exact same work is code duplication?and not having multiple hops today says nothing about the future? 

If you want to do chaining of AsyncResult, it would probably be better to have Future implement AshncResult and convert to Future, since at this point you are not saving anything with the separate concept.

belliottsmith: Yep, there should only be two CAS, same as we would need to directly transform. One to register the callback on the old AsyncResult, and one to set the new AsyncResult outcome. We have two intermediate objects: the Head (which is hopefully stack allocated), the Map node, and then the AsyncResult itself. It looks like your branch does something very similar with its direct implementation?

belliottsmith: If there's anything I would change, it would be the `beginAsResult` method to avoid the extra object invoking `settingCallback()` and instead have a variant of `AsyncResult.Settable` that is itself a callback. I would also consider renaming `toChain` as I think this gives the impression of more expense than there really is. HotSpot _should_ stack allocate the `Head` object, so we literally devolve to the same behaviour.

accord-core/src/main/java/accord/local/Command.java
belliottsmith: This seems quite similar to the pattern of invoking with the `node.agent()`. I'm down with standardising this to some global singleton that handles propagating exceptions, but let's do it across the board? Perhaps we want to call it `logErrors` or something else, that makes it clearer what's happening? Though `noOp` is also probably fine, and I don't mind terribly what we name this.

accord-core/src/main/java/accord/utils/async/AsyncResult.java
dcapwell: > Wdym more cas/volatile overhead? There should be no more overhead in this regard. If there is, it’s a mistake and we should fix it.

We have a AsyncResult, we create a AsyncChain which listens to the result, we then produce a new AsyncResult which is a copy of the first one.  If we used the AsyncResult directly we would avoid the extra overhead (listen + copying result); see extra last comment going over this more clearly.

> Slightly different code performing the exact same work is code duplication?

In this case its the same semantic work, but that's a similar argument to saying `Iterator` shouldn't have a `forEach` because `Stream` has a `forEach`, or that `Routables.foldl` and `Topology.foldl` should not both exist.  Why is it a problem for `map` and `flatMap` to only exist in one type where as `forEach` and `fold` are allowed across multiple types?

> and not having multiple hops today says nothing about the future?

It does about the concern. If the concern is that we will have multiple hops, `AsyncChains` offers a way to lower this when we get this issue, but this has nothing to do with the majority of cases which only have a single hop.  Why choose to impact the majority of cases because one day we may have a case that could choose to use a different API call?

> If you want to do chaining of AsyncResult, it would probably be better to have Future implement AshncResult and convert to Future, since at this point you are not saving anything with the separate concept.

From a user/contributor to this project, `AsyncResult` to me is just a `Future`.  We are able to remove the cassandra dependency by taking ownership of our own future, which has its own benefit.  If we rename `AsyncResult` to `Future`, I am cool with this...

> Yep, there should only be two CAS, same as we would need to directly transform. One to register the callback on the old AsyncResult, and one to set the new AsyncResult outcome.

In the case of `map` they are equal, never attempted to argue that there are more CAS in this regard.  My comment about extra CAS was the new pattern of

```
private AsyncChain<Void> applyChain(SafeCommandStore safeStore, boolean canReschedule)
...
return applyWithCorrectScope(safeStore.commandStore()).toChain();

// usage
private AsyncResult<Void> applyWithCorrectScope(CommandStore unsafeStore)
...
AsyncChain<Void> chain = super.applyChain(safeStore);
result = AsyncResults.forChain(chain);
```

This pattern exists a few times in the Cassandra PR and are all the same:
1) We have a async operation, so have a `AsyncResult`
2) API forces `AsyncChain`, so "listen" to the result and return the chain
3) We want the result, so convert to `AsyncResult`

This pattern is where my comment comes from, this pattern "could" return the original result, but because we expose Chains we now have 2 extra CAS (1 to listen, 1 to store the result).

belliottsmith: From an efficiency perspective, I disagree this is material, as the bulk of cost can be resolved with an override to `beginAsResult`.

However, from an ergonomics perspective I agree this is suboptimal. I have pushed branches* that make `AsyncResult` an `AsyncChain` so that it can be used directly wherever an `AsyncChain` would be, and we avoid any conversion. The only semantic difference is that an `AsyncResult` may have an unlimited number of distinct `AsyncChain` started from them, so they create a kind of synchronisation point in the chain.

\*
https://github.com/belliottsmith/cassandra-accord/tree/async-chain-suggest
https://github.com/belliottsmith/cassandra/tree/async-chains-suggest

accord-core/src/main/java/accord/coordinate/Coordinate.java
```diff
@@ -34,19 +34,19 @@
 import accord.messages.PreAccept.PreAcceptReply;
 import com.google.common.base.Preconditions;
```
belliottsmith, line 35: We've migrated to `Invariants` mostly, to aid debugging (and also it's a better name)

accord-core/src/main/java/accord/utils/async/AsyncChainCombiner.java
```diff
@@ -0,0 +1,150 @@
+package accord.utils.async;
+
+import com.google.common.base.Preconditions;
```
belliottsmith, line 21: Also here

accord-core/src/main/java/accord/utils/async/AsyncResults.java
```diff
@@ -0,0 +1,309 @@
+package accord.utils.async;
+
+import com.google.common.base.Preconditions;
```
belliottsmith, line 21: ... and here

accord-core/src/main/java/accord/utils/async/AsyncResults.java
```diff
@@ -0,0 +1,309 @@
...
+        Preconditions.checkArgument(!results.isEmpty());
+        if (results.size() == 1)
+            return results.get(0).toChain();
+        return new AsyncChainCombiner.Reduce<>(toChains(results), reducer);
+    }
+
+    public static <V> V getBlocking(AsyncResult<V> asyncResult) throws InterruptedException, ExecutionException
+    {
+        AtomicReference<Result<V>> callbackResult = new AtomicReference<>();
+        CountDownLatch latch = new CountDownLatch(1);
```
belliottsmith, line 324: Hmm. We have a problem here which is that these aren't very compatible with the Simulator. We are meant to use `newCountDownLatch` in Cassandra for this.

accord-core/src/main/java/accord/utils/async/AsyncChains.java
dcapwell: > Right, this is about the same as Stream#reduce, where the reducer is only called if there is more than one result to feed it.

tested `Stream.of(0).reduce((i1, i2) -> i1 + i2).get();` and it does match your semantics... I think I am remembering scalaz too much here, I am sorry.

accord-core/src/main/java/accord/utils/async/AsyncResult.java
dcapwell: Benedict and I spoke in slack, and the summary is:

1) to simplify the usage, make `AsyncResult` implement `AsyncChain`, this has 2 benefits: no more `toChain`, and `AsyncChains.forChain` can avoid the copy of the result and just return the original result (avoiding 2 extra CAS; this case exists in C*)
2) AsyncChains.Head does not block multiple calls to `begin` which can allow multiple execution in Cassandra that were not possible before; we should block this.  This is localized to Head as `chain.map(...).beginAsResult()` would fail if begin was called multiple times

dcapwell:
In my feedback branch I pushed JMH benchmarks for result.map vs chain.map, I re-structured the output to be easier to read (group by chainSize rather than benchmark).

```
Benchmark                   (chainSize)  Mode  Cnt   Score   Error  Units
AsyncChainBenchmark.chain             1  avgt       19.267          ns/op
AsyncChainBenchmark.result            1  avgt       13.028          ns/op

AsyncChainBenchmark.chain             2  avgt       24.885          ns/op
AsyncChainBenchmark.result            2  avgt       24.677          ns/op

AsyncChainBenchmark.chain             3  avgt       26.092          ns/op
AsyncChainBenchmark.result            3  avgt       36.276          ns/op

AsyncChainBenchmark.chain             4  avgt       30.516          ns/op
AsyncChainBenchmark.result            4  avgt       47.993          ns/op

AsyncChainBenchmark.chain             5  avgt       36.291          ns/op
AsyncChainBenchmark.result            5  avgt       60.386          ns/op
```

`chain` starts to win after 3 hops

I did the same for `flatMap`, but made the assumption that the source of async will produce a `AsyncResult` (which is true in the Cassandra branch)

```
Benchmark                          (chainSize)  Mode  Cnt    Score   Error  Units
AsyncChainBenchmark.chainFlatMap             1  avgt        33.366          ns/op
AsyncChainBenchmark.resultFlatMap            1  avgt        25.701          ns/op

AsyncChainBenchmark.chainFlatMap             2  avgt        56.366          ns/op
AsyncChainBenchmark.resultFlatMap            2  avgt        51.788          ns/op

AsyncChainBenchmark.chainFlatMap             3  avgt        70.098          ns/op
AsyncChainBenchmark.resultFlatMap            3  avgt        78.100          ns/op

AsyncChainBenchmark.chainFlatMap             4  avgt        96.927          ns/op
AsyncChainBenchmark.resultFlatMap            4  avgt       104.856          ns/op

AsyncChainBenchmark.chainFlatMap             5  avgt       111.148          ns/op
AsyncChainBenchmark.resultFlatMap            5  avgt       133.539          ns/op
```

Given the assumption above, the fact that both models are the same makes sense, as you can not build a chain with `flatMap` (you are producing results with a single hop)

belliottsmith:
This benchmark isn't very representative, as it uses `Immediate` and immediately completed `AsyncResult`, whereas in normal async operation we can expect the chain to be completed prior to each step's result. It also did not incorporate my improvements to `AsyncChain<->AsyncResult` performance. Also, by using a for loop it likely distorts HotSpot's compilation. 

I have taken your branch and pulled in my improvements to `AsyncChain`. I have modified the benchmark to use `AsyncResult.Settable` for each step of the transformation, setting all of the `Settable` only after the chain is assembled. I have introduced separate methods for each level of nesting. Note that I still use the `AsyncResult->AsyncResult` structure for all of these tests, so the start and end object are still `AsyncResult`. I have introduced one additional variant that produces `AsyncChain` for the intermediate results in `flatMap` for comparison, this is `chainFlatMap`; `chainFlatMapResult` uses `AsyncResult.Settable` for each intermediate `flatMap`, like `resultFlatMap`.

On JDK11, this test* comes out with the chain methods faster for every variation.

```
AsyncChainBenchmark.chainFlatMap1                                     avgt         30.961           ns/op
AsyncChainBenchmark.chainFlatMapResult1                               avgt         46.037           ns/op
AsyncChainBenchmark.resultFlatMap1                                    avgt         46.371           ns/op

AsyncChainBenchmark.chainFlatMap2                                     avgt         34.907           ns/op
AsyncChainBenchmark.chainFlatMapResult2                               avgt         72.164           ns/op
AsyncChainBenchmark.resultFlatMap2                                    avgt         72.878           ns/op

AsyncChainBenchmark.chainFlatMap3                                     avgt         48.796           ns/op
AsyncChainBenchmark.chainFlatMapResult3                               avgt         98.623           ns/op
AsyncChainBenchmark.resultFlatMap3                                    avgt        120.239           ns/op

AsyncChainBenchmark.chainMap1                                         avgt         27.309           ns/op
AsyncChainBenchmark.resultMap1                                        avgt         29.192           ns/op

AsyncChainBenchmark.chainMap2                                         avgt         31.080           ns/op
AsyncChainBenchmark.resultMap2                                        avgt         37.800           ns/op

AsyncChainBenchmark.chainMap3                                         avgt         39.622           ns/op
AsyncChainBenchmark.resultMap3                                        avgt         64.441           ns/op
```

\* https://github.com/belliottsmith/cassandra-accord/tree/async-chain-bench

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -197,12 +202,12 @@ public void withEpoch(long epoch, Runnable runnable)
...
         {
             configService.fetchTopologyForEpoch(epoch);
-            topology.awaitEpoch(epoch).addListener(runnable);
+            topology.awaitEpoch(epoch).addCallback(runnable);
         }
     }
 
     @Inline
-    public <T> Future<T> withEpoch(long epoch, Supplier<Future<T>> supplier)
+    public <T> AsyncResult<T> withEpoch(long epoch, Supplier<AsyncResult<T>> supplier)
```
dcapwell, line 209: 2/3 usage store into a cache of some sort, but given the bias towards `AsyncChain` I think we should work with chain instead.  Assuming we include Benedicts change that `AsyncResult` implements `AsyncChain` then we can change to

```
public <T> AsyncChain<T> withEpoch(long epoch, Supplier<? extends AsyncChain<T>> supplier)
    {
        if (topology.hasEpoch(epoch))
        {
            return supplier.get();
        }
        else
        {
            return topology.awaitEpoch(epoch).toChain().flatMap(ignore -> supplier.get());
        }
    }
```

The side effect of this is that the callers get changed to call `beginAsResult`, but makes this not bias towards completing the chain in the middle.

accord-core/src/main/java/accord/utils/ThreadPoolScheduler.java
```diff
@@ -96,7 +95,7 @@ public void stop()
         }
         catch (InterruptedException e)
         {
-            throw new UncheckedInterruptedException(e);
+            throw new RuntimeException(e);
```
dcapwell, line 98: this should have a negative side effect with Simulator... we check that exceptions threw `UncheckedInterruptedException` and act differently, so this would change behavior there.

accord-core/src/main/java/accord/utils/async/AsyncCallbacks.java
```diff
@@ -0,0 +1,45 @@
...
+
+
+    public static <T> BiConsumer<? super T, Throwable> inExecutor(Runnable runnable, Executor executor)
+    {
+        return (result, throwable) -> {
+            if (throwable == null) executor.execute(runnable);
+            else throw new RuntimeException(throwable);
+        };
+    }
+}
```
dcapwell, line 48: should change to be

```
public static <T> BiConsumer<? super T, Throwable> inExecutor(Runnable runnable, Executor executor)
    {
        return inExecutor((result, throwable) -> {
            if (throwable != null) 
                throw new RuntimeException(throwable);
            runnable.run();
        }, executor);
    }
```

The other method handles the case that `executor.execute` throws (which it can), so simplifies.

accord-core/src/main/java/accord/utils/async/AsyncCallbacks.java
```diff
@@ -0,0 +1,45 @@
...
+        if (failure != null)
+            logger.error("Exception received by noop callback", failure);
+    };
+
+    public static <T> BiConsumer<? super T, Throwable> noop()
+    {
+        return NOOP;
+    }
+
+    public static <T> BiConsumer<? super T, Throwable> inExecutor(BiConsumer<? super T, Throwable> callback, Executor executor)
```
dcapwell, line 26: nit: should move `executor` first as it cleans up the usage when lambdas are used:

current:

```
inExecutor((result, throwable) -> {
            if (throwable != null)
                throw new RuntimeException(throwable);
            runnable.run();
        }, executor);
```

desired:

```
inExecutor(executor, (result, throwable) -> {
            if (throwable != null)
                throw new RuntimeException(throwable);
            runnable.run();
        });
```

dcapwell: I 100% acknowledge that this wouldn't be consistent with `org.apache.cassandra.utils.concurrent.Future`

accord-core/src/main/java/accord/utils/async/AsyncChain.java
```diff
@@ -0,0 +1,60 @@
...
+     * Support {@link com.google.common.util.concurrent.Futures#transform(ListenableFuture, com.google.common.base.Function, Executor)} natively
+     */
+    <T> AsyncChain<T> flatMap(Function<? super V, ? extends AsyncChain<T>> mapper);
+
+    /**
+     * Support {@link com.google.common.util.concurrent.Futures#addCallback} natively
+     */
+    AsyncChain<V> addCallback(BiConsumer<? super V, Throwable> callback);
+
+    void begin(BiConsumer<? super V, Throwable> callback);
```
dcapwell, line 44: IMO we should define and document expected semantics as this may be brittle to people familiar with `Future` (left this comment about `Head` as well)...

*should* `begin` allow multiple calls, or should this be rejected?  The current implementations (other than `Head`) reject multiple calls, which I feel is the desired behavior in all cases

accord-core/src/main/java/accord/utils/async/AsyncChain.java
```diff
@@ -0,0 +1,60 @@
...
+
+    default void begin(Runnable runnable)
+    {
+        begin((unused, failure) -> {
+            if (failure == null) runnable.run();
+            else throw new RuntimeException(failure);
+        });
```
dcapwell, line 51: made a comment in `accord.utils.async.AsyncCallbacks#inExecutor(java.lang.Runnable, java.util.concurrent.Executor)` that we should refactor to use `accord.utils.async.AsyncCallbacks#inExecutor(java.util.function.BiConsumer<? super T,java.lang.Throwable>, java.util.concurrent.Executor)`... If we do this, then we can have the following

```
// in AsyncCallacks
public static <T> BiConsumer<? super T, Throwable> inExecutor(Runnable runnable, Executor executor)
    {
        return inExecutor(wrap(runnable), executor);
    }

    public static <T> BiConsumer<T, Throwable> wrap(Runnable runnable)
    {
        return (result, throwable) -> {
            if (throwable != null)
                throw new RuntimeException(throwable);
            runnable.run();
        };
    }

// here
begin(AsyncCallbacks.wrap(runnable));
```

accord-core/src/main/java/accord/utils/async/AsyncChainCombiner.java
```diff
@@ -0,0 +1,150 @@
...
+    protected AsyncChainCombiner(List<AsyncChain<I>> inputs)
+    {
+        Preconditions.checkArgument(!inputs.isEmpty());
+        this.state = inputs;
+    }
+
+    private List<AsyncChain<? extends I>> inputs()
+    {
+        Object current = state;
+        Preconditions.checkState(current instanceof List);
```
dcapwell, line 45: nit: when this isn't true its great to know what the class is, so its easier to debug what happens

accord-core/src/main/java/accord/utils/async/AsyncChainCombiner.java
```diff
@@ -0,0 +1,150 @@
...
+        return (List<AsyncChain<? extends I>>) current;
+    }
+
+    private I[] results()
+    {
+        Object current = state;
+        Preconditions.checkState(current instanceof Object[]);
```
dcapwell, line 52: nit: when this isn't true its great to know what the class is, so its easier to debug what happens

accord-core/src/main/java/accord/utils/async/AsyncChainCombiner.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    int size()
+    {
+        Object current = state;
+        if (current instanceof List)
+            return ((List) current).size();
+        if (current instanceof Object[])
+            return ((Object[]) current).length;
+        throw new IllegalStateException();
```
dcapwell, line 73: nit: when this isn't true its great to know what the class is, so its easier to debug what happens

accord-core/src/main/java/accord/utils/async/AsyncChainCombiner.java
```diff
@@ -0,0 +1,150 @@
...
+
+    abstract void complete(I[] results, BiConsumer<? super O, Throwable> callback);
+
+    private void callback(int idx, I result, Throwable throwable)
+    {
+        int current = remaining;
+        if (current == 0)
+            return;
+
+        if (throwable != null && REMAINING.compareAndSet(this, current, 0))
```
dcapwell, line 84: shouldn't this be a CAS loop?  The following timing looks to do the wrong thing

```
T1: callback(?, null, NullPointerException), current = 4
T2: callback(?, "success", null), current = 4
T2: update results, `REMAINING.decrementAndGet(this)`
T1: throwable != null == true, REMAINING.compareAndSet(this, current, 0) == false
T1: updates results with a null <- unexpected behavior
```

dcapwell: Think it should be the following

```
if (throwable != null)
        {
            if (markFailed())
                callback.accept(null, throwable);

            return;
        }
...
private boolean markFailed()
    {
        int current;
        do
        {
            current = remaining;
            if (current == 0)
                return false;
        }
        while (!REMAINING.compareAndSet(this, current, 0));
        return true;
    }
```

This should also handle the case that multiple exceptions are seen at the same time.

accord-core/src/main/java/accord/utils/async/AsyncChainCombiner.java
```diff
@@ -0,0 +1,150 @@
...
+    {
+        All(List<AsyncChain<V>> asyncChains)
+        {
+            super(asyncChains);
+        }
+
+        @Override
+        void complete(V[] results, BiConsumer<? super List<V>, Throwable> callback)
+        {
+            List<V> result = Lists.newArrayList(results);
```
dcapwell, line 132: not in j8, can we use `Arrays.asList` or wrap?  

see https://docs.oracle.com/javase/8/docs/api/java/util/List.html

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -0,0 +1,432 @@
...
+        protected Head()
+        {
+            super(null);
+            next = this;
+        }
+
+        void begin()
+        {
+            begin(next);
+        }
```
dcapwell: believe Benedict fixes in his branch, but this impl allows calling `begin(BiConsumer)` multiple times, which could have bad side effects, we should guard against this

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -0,0 +1,432 @@
...
+        protected Link(Head<?> head)
+        {
+            super(head);
+        }
+
+        @Override
+        public void begin(BiConsumer<? super O, Throwable> callback)
+        {
+            Preconditions.checkArgument(!(callback instanceof AsyncChains.Head));
+            Preconditions.checkState(next instanceof AsyncChains.Head);
```
dcapwell, line 119: nit: would be great to have useful error messages

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -0,0 +1,432 @@
...
+    public AsyncChain<V> addCallback(BiConsumer<? super V, Throwable> callback)
+    {
+        return add(EncapsulatedCallback::new, callback);
+    }
+
+    // can be used by transformations that want efficiency, and can directly extend Link, FlatMap or Callback
+    // (or perhaps some additional helper implementations that permit us to simply implement apply for Map and FlatMap)
+    <O, T extends AsyncChain<O> & BiConsumer<? super V, Throwable>> AsyncChain<O> add(Function<Head<?>, T> factory)
+    {
+        Preconditions.checkState(next instanceof Head<?>);
```
dcapwell, line 252: nit: would be great to have useful error messages

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -0,0 +1,432 @@
...
+        Head<?> head = (Head<?>) next;
+        T result = factory.apply(head);
+        next = result;
+        return result;
+    }
+
+    <P, O, T extends AsyncChain<O> & BiConsumer<? super V, Throwable>> AsyncChain<O> add(BiFunction<Head<?>, P, T> factory, P param)
+    {
+        Preconditions.checkState(next instanceof Head<?>);
```
dcapwell, line 261: nit: would be great to have useful error messages

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -0,0 +1,432 @@
...
+        chain.begin((result, failure) -> {
+            callbackResult.set(new Result(result, failure));
+            latch.countDown();
+        });
+
+        latch.await();
+        Result result = callbackResult.get();
+        if (result.failure == null) return result.result;
+        else throw new ExecutionException(result.failure);
+    }
```
dcapwell, line 389: can simplify to be

```
CompletableFuture<V> future = new CompletableFuture<>();
        chain.begin((result, failure) -> {
            if (failure != null) future.completeExceptionally(failure);
            else future.complete(result);
        });
        return future.get();
```

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -0,0 +1,432 @@
...
+        chain.begin((result, failure) -> {
+            callbackResult.set(new Result(result, failure));
+            latch.countDown();
+        });
+
+        if (!latch.await(timeout, unit))
+            throw new TimeoutException();
+        Result result = callbackResult.get();
+        if (result.failure == null) return result.result;
+        else throw new ExecutionException(result.failure);
```
dcapwell, line 417: can simplify with

```
CompletableFuture<V> future = new CompletableFuture<>();
        chain.begin((result, failure) -> {
            if (failure != null) future.completeExceptionally(failure);
            else future.complete(result);
        });
        return future.get(timeout, unit);
```

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -0,0 +1,432 @@
...
+        catch (ExecutionException | InterruptedException e)
+        {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static <V> void awaitUninterruptibly(AsyncChain<V> chain)
+    {
+        getUninterruptibly(chain);
+    }
```
dcapwell, line 435: Might not be what people expect, I think it should be

```
public static <V> V getUninterruptibly(AsyncChain<V> chain) throws ExecutionException
    {
        boolean interrupted = false;
        try
        {
            while (true)
            {
                try
                {
                    return getBlocking(chain);
                }
                {
                    interrupted = true;
                }
            }
        }
        finally
        {
            if (interrupted)
                Thread.currentThread().interrupt();
        }
    }

    public static <V> void awaitUninterruptibly(AsyncChain<V> chain) throws ExecutionException
    {
        getUninterruptibly(chain);
    }
```

accord-core/src/main/java/accord/utils/async/AsyncResult.java
```diff
@@ -0,0 +1,86 @@
...
+public interface AsyncResult<V>
+{
+    void addCallback(BiConsumer<? super V, Throwable> callback);
+
+    default void addCallback(Runnable runnable)
+    {
+        addCallback((unused, failure) -> {
+            if (failure == null) runnable.run();
+            else throw new RuntimeException(failure);
+        });
```
dcapwell: should do the refactor I mention in `AsyncCallbacks`, this would let us do the following

```
default void addCallback(Runnable runnable)
{
    addCallback(AsyncCallbacks.wrap(runnable));
}
```

accord-core/src/main/java/accord/utils/async/AsyncResult.java
```diff
@@ -0,0 +1,86 @@
...
+
+    default void addListener(Runnable runnable)
+    {
+        addCallback(runnable);
+    }
+
+    default void addListener(Runnable runnable, Executor executor)
+    {
+        addCallback(runnable, executor);
+    }
```
dcapwell: can we remove?  these are duplicates of `addCallback`

accord-core/src/main/java/accord/utils/async/AsyncResults.java
```diff
@@ -0,0 +1,309 @@
...
+                runnable.run();
+                result.trySuccess(null);
+            }
+            catch (Exception e)
+            {
+                result.tryFailure(e);
+            }
+        });
+        return result;
+    }
```
dcapwell, line 300: wondering if we should remove in favor of `AsyncChains`.  This method is eager and returns a future of the result, where as chains is lazy and lets you keep adding to the chain until you are ready.  Users can use `AsyncChains.ofCallable(...).beginAsResult()` to get similar semantics.

This code is also dead code atm, so helps the argument to remove

accord-core/src/main/java/accord/utils/async/AsyncResults.java
dcapwell: I gave feedback in `AsyncChains` to switch to `CompletableFuture` to simplify the code, guess it would have the same issue with simulator...

dcapwell: @belliottsmith curious why?  I can look closer at our rewrite logic more (such as `org.apache.cassandra.simulator.asm.GlobalMethodTransformer#visitMethodInsn`) but I feel that if we are rewriting all C*/Accord classes we could always intercept `new` and not require the `new*` methods to be called?   I see in this case we don't extend java's and instead add a new `org.apache.cassandra.utils.concurrent.CountDownLatch` which appears to be replaced with a call to `org.apache.cassandra.simulator.systems.InterceptingGlobalMethods#newCountDownLatch` which returns a `org.apache.cassandra.simulator.systems.InterceptingAwaitable.InterceptingCountDownLatch`

Not sure how simulator deals with this, but the base types I see are `org.apache.cassandra.simulator.systems.InterceptingAwaitable` and `org.apache.cassandra.simulator.systems.InterceptedWait.TriggerListener`, so feel we could do the same for the base java concurrent utils.

dcapwell (commented):
finished reviewing and mostly ok assuming the patches from Benedicts patch are applied

accord-core/src/main/java/accord/utils/async/AsyncCallbacks.java
belliottsmith: Is this the right contract? We seem to be relying on the executor’s exception handling, and may be duplicating the exception handling since this is typically for a listener not the final consumer - should we just not submit to the executor if throwable is non-null?

In general I think composability should terminate with the exception, except for use cases that explicitly handle both…

accord-core/src/main/java/accord/utils/async/AsyncCallbacks.java
belliottsmith: Agreed

accord-core/src/main/java/accord/utils/async/AsyncChain.java
belliottsmith: Agreed, and addressed in my branch

accord-core/src/main/java/accord/utils/async/AsyncChains.java
belliottsmith: Can’t remember if this is simulator compatible or not - it might be, but less efficiently

accord-core/src/main/java/accord/utils/async/AsyncResults.java
belliottsmith: Agreed

accord-core/src/main/java/accord/utils/async/AsyncResults.java
belliottsmith: Now we use byte weaving we have a lot of power to take over things we didn’t initially. So yes in principle we can capture these. There’s two ways, 
1) by providing an overriding implementation and replacing allocations, as we do for ConcurrentHashMap - this isn’t ideal though and I think doesn’t work for all juc classes as I think not all can be fully overridden
2) via LockSupport intercept

(2) should theoretically already be supported, but is inefficient and not used extensively so might not be fully working - this patch does break simulation though, and I haven’t had time to fully dig into all of the issues, but I might actually be wrong about this being one of the reasons. I will have to dig in further.

accord-core/src/main/java/accord/utils/async/AsyncCallbacks.java
dcapwell: > should we just not submit to the executor if throwable is non-null?

I spent some time triggering failures at different locations to see that we are currently not consistent... so we should flesh out the semantics and push this to the type system...

Here are the cases I tested

```
AsyncChains.ofCallable(ignore -> {
                    throw new RejectedExecutionException();
                }, () -> 42)
                .map(i -> i + 1)
                .begin((success, failure) -> {
                    if (failure == null)
                        throw new IllegalStateException("Should see failure");
                });
```

`begin` throws, so caller needs to handle.

```
AsyncChains.ofCallable(ignore -> {
                    throw new RejectedExecutionException();
                }, () -> 42)
                .map(i -> i + 1)
                .beginAsResult()
                .addCallback((success, failure) -> {
                    if (failure == null)
                        throw new IllegalStateException("Expected to fail");
                });
```

`beginAsResult ` throws, so caller needs to handle.

```
AsyncChains.<Integer>ofCallable(fn -> fn.run(), () -> {
                    throw new RuntimeException("Unchecked");
                }).map(i -> i + 1).map(i -> i + 1)
                .begin((success, failure) -> {
                    if (failure == null)
                        throw new IllegalStateException("Should see failure");
                });
```

`begin` sees the exception rather than throws it, so caller doesn't see an exception

```
AsyncChains.<Integer>ofCallable(fn -> fn.run(), () -> {
                    throw new RuntimeException("Unchecked");
                }).map(i -> i + 1).map(i -> i + 1)
                .begin(() -> {
                });
```

`begin` throws, so caller needs to handle.

```
AsyncChains.ofCallable(fn -> fn.run(), () -> 42
                ).map(i -> i + 1)
                .map(ignore -> {
                    throw new RuntimeException("Unchecked");
                })
                .begin((success, failure) -> {
                    if (failure == null)
                        throw new IllegalStateException("Should see failure");
                });
```

`begin` does not fail here, the error is seen by the callback

```
AsyncChains.ofCallable(fn -> fn.run(), () -> 42).
                map(i -> i + 1)
                .map(ignore -> {
                    throw new RuntimeException("Unchecked");
                }).begin(() -> {
                });
```

`begin` throws, so caller needs to handle.



Small changes have big impacts to the callers of `begin`, which may not know about what happened to the other parts of the chain...

Personally I find this behavior brittle, and would propose the following

1) `begin(Runnable)` is removed or replaced with a `begin(Runnable, Consumer<Throwable>)`.  This is only used by tests, so safest to just remove.
2) we catch at all places and send the exception down the chain, `begin` should never throw

Main reason for #2 is that the call to `begin` is expected to handle exceptions async, so by having it also throw, we make the usage harder to do correctly... by centralizing the exceptions we avoid future issues.

accord-core/src/main/java/accord/utils/async/AsyncCallbacks.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.concurrent.Executor;
+import java.util.function.BiConsumer;
+
+public class AsyncCallbacks
+{
+    private static final Logger logger = LoggerFactory.getLogger(AsyncCallbacks.class);
+
+    private static final BiConsumer<Object, Throwable> NOOP = (unused, failure) -> {
+        if (failure != null)
+            logger.error("Exception received by noop callback", failure);
```
belliottsmith: This should be passed to some global hook for unexpected exceptions

accord-core/src/main/java/accord/utils/async/AsyncCallbacks.java
```diff
@@ -0,0 +1,45 @@
...
+            }
+        };
+    }
+
+
+    public static <T> BiConsumer<? super T, Throwable> inExecutor(Runnable runnable, Executor executor)
+    {
+        return (result, throwable) -> {
+            if (throwable == null) executor.execute(runnable);
+            else throw new RuntimeException(throwable);
```
belliottsmith, line 45: This still needs to be routed to some global sink/handler

accord-core/src/main/java/accord/utils/async/AsyncChainCombiner.java
belliottsmith: Yeah, this is a bug. I've made some suggested changes to this class to address this and also avoid allocating additional memory for callbacks (given the choice, I'd probably prefer an additional CAS to an allocation): https://github.com/belliottsmith/cassandra-accord/tree/async-chains-suggest2

accord-core/src/main/java/accord/utils/async/AsyncResult.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.concurrent.Executor;
+import java.util.function.BiConsumer;
+
+/**
+ * Handle for async computations that supports multiple listeners and registering
+ * listeners after the computation has started
+ */
+public interface AsyncResult<V> extends AsyncChain<V>
+{
+    AsyncResult<V> addCallback(BiConsumer<? super V, Throwable> callback);
```
dcapwell, line 30: now that this is also a `AsyncChain` this already exists, are you trying to retain the return type?  If so can we add `@Override`, else remove?

accord-core/src/main/java/accord/utils/async/AsyncResult.java
```diff
@@ -0,0 +1,107 @@
...
+
+    default AsyncResult<V> addCallback(Runnable runnable)
+    {
+        return addCallback((unused, failure) -> {
+            if (failure == null) runnable.run();
+            else throw new RuntimeException(failure);
```
dcapwell, line 36: I think this gets back to a comment I was making in another thread, error handling isn't consistent with AsyncChains and semantics are not clear as they different from `Future`.


```
AsyncChains.failure(new NullPointerException())
                .beginAsResult()
                .addCallback(() -> System.out.println("wont happen"))
                .begin((success, failure)-> {
                    if (failure != null) System.err.println("Begin saw a failure");
                    else System.out.println("Success");
                });
```

I would 100% expect `begin` to handle the exception, but the *caller* to `begin` handles it... 

Now, what happens with JDK futures?

```
CompletableFuture.<Integer>failedFuture(new NullPointerException())
                .whenComplete((s, f) -> {
                    // futures doesn't expose a similar API as errors should be handled
                    if (f == null) System.out.println("wont happen");
                })
                .thenAccept(System.out::println)
                .exceptionally(ignore -> null)
                .join();
```

The closest is the `accept` method, which no-op in the failure case.

We really shouldn't throw as this becomes unexpected.  Also chains isn't "handling" these types of throws, so they bubble up to the caller of `begin`

accord-core/src/main/java/accord/utils/async/AsyncResult.java
```diff
@@ -0,0 +1,107 @@
...
+    {
+        addCallback(runnable);
+        return this;
+    }
+
+    default AsyncResult<V> addListener(Runnable runnable, Executor executor)
+    {
+        addCallback(runnable, executor);
+        return this;
+    }
```
dcapwell, line 65: why mix "listener" and "addCallback"... this is just a rename; can we remove?

accord-core/src/main/java/accord/utils/async/AsyncResult.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Override
+    default void begin(BiConsumer<? super V, Throwable> callback)
+    {
+        addCallback(callback);
+    }
+
+    default AsyncResult<V> beginAsResult()
```
dcapwell, line 73: `@Override`?

accord-core/src/main/java/accord/utils/async/AsyncResults.java
```diff
@@ -0,0 +1,368 @@
...
+        boolean trySetResult(V result, Throwable failure)
+        {
+            return trySetResult(new Result<>(result, failure));
+        }
+
+        void setResult(Result<V> result)
+        {
+            if (!trySetResult(result))
+                throw new IllegalStateException("Result has already been set on " + this);
+        }
```
dcapwell, line 98: dead code, can we remove?

accord-core/src/main/java/accord/utils/async/AsyncResults.java
```diff
@@ -0,0 +1,368 @@
...
+
+    public static <V> V getUninterruptibly(AsyncResult<V> asyncResult)
+    {
+        try
+        {
+            return getBlocking(asyncResult);
+        }
+        catch (ExecutionException | InterruptedException e)
+        {
+            throw new RuntimeException(e);
```
dcapwell, line 360: when `ExecutionException` can you do `.getCause`?

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -455,25 +456,35 @@ public SynchronizedState(NodeTimeService time, Agent agent, DataStore store, Pro
...
             @Override
-            public Future<Void> execute(PreLoadContext context, Consumer<? super SafeCommandStore> consumer)
+            public AsyncChain<Void> execute(PreLoadContext context, Consumer<? super SafeCommandStore> consumer)
             {
                 return submit(context, i -> { consumer.accept(i); return null; });
             }
 
             @Override
-            public synchronized <T> Future<T> submit(PreLoadContext context, Function<? super SafeCommandStore, T> function)
+            public synchronized <T> AsyncChain<T> submit(PreLoadContext context, Function<? super SafeCommandStore, T> function)
```
dcapwell, line 465: don't think you need `synchronized` anymore

accord-core/src/main/java/accord/local/Node.java
dcapwell: Looking at the usage I still think this should be `Chain` and not `Result`.  `SimpleProgressLog` is one of the main callers from this patch and this adds 2 steps in the chain, so rather than turning to result eager we can do it at the final caller site, such as in `accord.impl.SimpleProgressLog.Instance.State.NonHomeState#run`; this could also call `begin` rather than `addCallback` as they have the same impact, but would be cheaper to do `begin`?

accord-core/src/main/java/accord/utils/ThreadPoolScheduler.java
dcapwell: I think we need to fork this to stay consistent with simulator

accord-core/src/main/java/accord/utils/async/AsyncCallbacks.java
belliottsmith: +1 to David's proposals here

dcapwell: copy/pasted these tests and ran them again, still seeing this behavior; `begin` is throwing rather than giving the exception to the callback

The only tests that do not fail are

```
AsyncChains.<Integer>ofCallable(fn -> fn.run(), () -> {
                    throw new RuntimeException("Unchecked");
                }).map(i -> i + 1).map(i -> i + 1)
                .begin((success, failure) -> {
                    if (failure == null)
                        throw new IllegalStateException("Should see failure");
                });

AsyncChains.ofCallable(fn -> fn.run(), () -> 42
                ).map(i -> i + 1)
                .map(ignore -> {
                    throw new RuntimeException("Unchecked");
                })
                .begin((success, failure) -> {
                    if (failure == null)
                        throw new IllegalStateException("Should see failure");
                });
```

accord-core/src/main/java/accord/utils/async/AsyncChains.java
dcapwell: This is still an issue, so created a test case to show it; `Head` still allows!

```
@Test
    void headRejectsSecondBegin()
    {
        AsyncChain<String> chain = new AsyncChains.Head<>() {
            @Override
            public void begin(BiConsumer<? super String, Throwable> callback) {
                callback.accept("success", null);
            }
        };

        chain.begin((i1, i2) -> {});
        assertThrows(() -> chain.begin((i1, i2) -> {}));
    }

    @Test
    void chainRejectsSecondBegin()
    {
        AsyncChain<String> chain = new AsyncChains.Head<>() {
            @Override
            public void begin(BiConsumer<? super String, Throwable> callback) {
                callback.accept("success", null);
            }
        };
        chain = chain.map(s -> s + " is true");
        chain.begin((i1, i2) -> {});
        AsyncChain<String> finalChain = chain;
        assertThrows(() -> finalChain.begin((i1, i2) -> {}));
    }

    private static void assertThrows(Runnable fn)
    {
        try
        {
            fn.run();
            Assertions.fail("Should have been rejected");
        }
        catch (AssertionError e)
        {
            if ("Should have been rejected".equals(e.getMessage())) throw e;
        }
        catch (Throwable t)
        {
            // expected
        }
    }
```

accord-core/src/main/java/accord/utils/async/AsyncCallbacks.java
dcapwell: yep, same issue if `begin`'s callback throws, what do we notify?

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
dcapwell: fixed in `immutable-state`

accord-core/src/main/java/accord/utils/async/AsyncResults.java
dcapwell: fixed in `immutable-state`

accord-core/src/main/java/accord/utils/async/AsyncResults.java
dcapwell: not dead code in `immutable-state`

accord-core/src/main/java/accord/utils/async/AsyncResult.java
dcapwell: fixed in immutable-state

accord-core/src/main/java/accord/utils/async/AsyncResult.java
dcapwell: I created a test case with the above example and `addCallback` returns `null` now...

```
@Test
    void test3()
    {
        AsyncChain<Object> chain = AsyncChains.failure(new NullPointerException("just kidding"));
        AsyncResult<Object> result = chain.beginAsResult();
        AsyncResult<Object> chain2 = result.addCallback(() -> System.out.println("wont happen"));
        chain2.begin((success, failure)-> { // chain2 is null!
                    if (failure != null) System.err.println("Begin saw a failure");
                    else System.out.println("Success");
                });
    }
```

dcapwell: found it, `accord.utils.async.AsyncResults.AbstractResult#addCallback` we return `null` when the chain is complete, but we should return `this`, fixing this causes the test to pass; below is the cleaned up version

```
@Test
    void test3()
    {
        AtomicReference<Boolean> sawFailure = new AtomicReference<>(null);
        AsyncChains.failure(new NullPointerException("just kidding"))
                .beginAsResult()
                .addCallback(() -> System.out.println("wont happen"))
                .begin((success, failure) -> {
                    if (failure != null) sawFailure.set(true);
                    else sawFailure.set(false);
                });
        Assertions.assertEquals(Boolean.TRUE, sawFailure.get());
    }
```

dcapwell: ok added an assert to make sure we don't trigger the callback; still passes

```
@Test
    void test3()
    {
        AtomicReference<Boolean> sawFailure = new AtomicReference<>(null);
        AtomicBoolean sawCallback = new AtomicBoolean(false);
        AsyncChains.failure(new NullPointerException("just kidding"))
                .beginAsResult()
                .addCallback(() -> sawCallback.set(true))
                .begin((success, failure) -> {
                    if (failure != null) sawFailure.set(true);
                    else sawFailure.set(false);
                });
        Assertions.assertEquals(Boolean.TRUE, sawFailure.get());
        Assertions.assertFalse(sawCallback.get());
    }
```

accord-core/src/main/java/accord/utils/async/AsyncChainCombiner.java
```diff
@@ -0,0 +1,168 @@
...
+        results()[idx] = result;
+        if (REMAINING.decrementAndGet(this) == 0)
+        {
+            try
+            {
+                complete(results(), callback);
+            }
+            catch (Throwable t)
+            {
+                callback.accept(null, t);
```
belliottsmith, line 99: I don't think this is safe, as we may be catching an exception thrown by the callback itself in `complete`. I think we need to say `complete` takes control of the callback, and invokes it exceptionally if it is unable to invoke it unexceptionally. But once invoked, control is handed to the callback, and we shouldn't invoke it _again_ with an exception.

belliottsmith:
This has been merged.

PR #22, opened 2023-01-06
TxnId should be more succinct and more useful

belliottsmith:
Specifically, TxnId should have no more than 20 bytes of payload, and should encode simple flags about a Txn, such as whether it is a read or write transaction, and whether it operates on keys or ranges.

### DIFF
--- a/accord-core/src/main/java/accord/impl/InMemoryCommandsForKey.java
+++ b/accord-core/src/main/java/accord/impl/InMemoryCommandsForKey.java
@@ -37,7 +37,6 @@ import static accord.local.CommandsForKey.CommandTimeseries.TestDep.*;
 import static accord.local.CommandsForKey.CommandTimeseries.TestKind.RorWs;
 import static accord.local.Status.KnownDeps.DepsUnknown;
 import static accord.local.Status.PreAccepted;
-import static accord.primitives.Txn.Kind.WRITE;
 
 public class InMemoryCommandsForKey extends CommandsForKey
 {
@@ -78,7 +77,7 @@ public class InMemoryCommandsForKey extends CommandsForKey
         public Stream<T> before(@Nonnull Timestamp timestamp, @Nonnull TestKind testKind, @Nonnull TestDep testDep, @Nullable TxnId depId, @Nonnull TestStatus testStatus, @Nullable Status status)
         {
             return commands.headMap(timestamp, false).values().stream()
-                    .filter(cmd -> testKind == RorWs || cmd.kind() == WRITE)
+                    .filter(cmd -> testKind == RorWs || cmd.txnId().isWrite())
                     // If we don't have any dependencies, we treat a dependency filter as a mismatch
                     .filter(cmd -> testDep == ANY_DEPS || (cmd.known().deps != DepsUnknown && (cmd.partialDeps().contains(depId) ^ (testDep == WITHOUT))))
                     .filter(cmd -> TestStatus.test(cmd.status(), testStatus, status))
@@ -89,7 +88,7 @@ public class InMemoryCommandsForKey extends CommandsForKey
         public Stream<T> after(@Nonnull Timestamp timestamp, @Nonnull TestKind testKind, @Nonnull TestDep testDep, @Nullable TxnId depId, @Nonnull TestStatus testStatus, @Nullable Status status)
         {
             return commands.tailMap(timestamp, false).values().stream()
-                    .filter(cmd -> testKind == RorWs || cmd.kind() == WRITE)
+                    .filter(cmd -> testKind == RorWs || cmd.txnId().isWrite())
                     // If we don't have any dependencies, we treat a dependency filter as a mismatch
                     .filter(cmd -> testDep == ANY_DEPS || (cmd.known().deps != DepsUnknown && (cmd.partialDeps().contains(depId) ^ (testDep == WITHOUT))))
                     .filter(cmd -> TestStatus.test(cmd.status(), testStatus, status))
--- a/accord-core/src/main/java/accord/local/Command.java
+++ b/accord-core/src/main/java/accord/local/Command.java
@@ -72,9 +72,6 @@ public abstract class Command implements CommandListener, BiConsumer<SafeCommand
 
     public abstract TxnId txnId();
 
-    public abstract Kind kind();
-    public abstract void setKind(Kind kind);
-
     // TODO (desirable, API consistency): should any of these calls be replaced by corresponding known() registers?
     public boolean hasBeen(Status status)
     {
@@ -295,12 +292,12 @@ public abstract class Command implements CommandListener, BiConsumer<SafeCommand
             return AcceptOutcome.Redundant;
         }
 
-        if (known().isDefinitionKnown() && !kind().equals(kind))
+        if (known().isDefinitionKnown() && txnId().rw() != kind)
             throw new IllegalArgumentException("Transaction kind is different to the definition we have already received");
 
         TxnId txnId = txnId();
         Ranges coordinateRanges = coordinateRanges(safeStore);
-        Ranges acceptRanges = txnId.epoch == executeAt.epoch ? coordinateRanges : safeStore.ranges().between(txnId.epoch, executeAt.epoch);
+        Ranges acceptRanges = txnId.epoch() == executeAt.epoch() ? coordinateRanges : safeStore.ranges().between(txnId.epoch(), executeAt.epoch());
         ProgressShard shard = progressShard(safeStore, route, progressKey, coordinateRanges);
 
         if (!validate(coordinateRanges, Ranges.EMPTY, shard, route, Ignore, null, Ignore, partialDeps, Set))
@@ -312,7 +309,6 @@ public abstract class Command implements CommandListener, BiConsumer<SafeCommand
         setExecuteAt(executeAt);
         setPromised(ballot);
         setAccepted(ballot);
-        setKind(kind);
         set(safeStore, coordinateRanges, Ranges.EMPTY, shard, route, null, Ignore, partialDeps, Set);
         switch (status())
         {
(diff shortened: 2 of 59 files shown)


accord-core/src/main/java/accord/primitives/RangeDeps.java
```diff
@@ -0,0 +1,5 @@
+package accord.primitives;
```
iamaleksey: Stray from adjacent work?

accord-core/src/main/java/accord/impl/InMemoryCommandsForKey.java
```diff
@@ -37,7 +37,7 @@
 import static accord.local.CommandsForKey.CommandTimeseries.TestKind.RorWs;
 import static accord.local.Status.KnownDeps.DepsUnknown;
 import static accord.local.Status.PreAccepted;
-import static accord.primitives.Txn.Kind.WRITE;
+import static accord.primitives.Txn.Kind.Write;
```
iamaleksey: Dead import

accord-core/src/main/java/accord/messages/BeginInvalidation.java
```diff
@@ -35,7 +35,7 @@ public BeginInvalidation(TxnId txnId, Unseekables<?, ?> someUnseekables, Ballot
```
iamaleksey, line 35: Pre-existing: missing `@Override`

accord-core/src/main/java/accord/messages/InformOfTxnId.java
```diff
@@ -25,7 +23,7 @@ public InformOfTxnId(TxnId txnId, RoutingKey homeKey)
     public void process()
```
iamaleksey, line 24: Pre-existing: missing `@Override`

accord-core/src/test/java/accord/txn/DepsTest.java
```diff
@@ -312,21 +312,21 @@ static Deps generate(Gen.Random random)
         {
             int epochRange = 3;
             int realRange = 50;
```
iamaleksey: Rename to hlcRange?

accord-core/src/main/java/accord/primitives/Ballot.java
```diff
@@ -33,4 +38,9 @@ public Ballot(long epoch, long real, int logical, Id node)
     {
```
iamaleksey, line 48: This constructor is a dirty liar.

accord-core/src/test/java/accord/coordinate/TopologyChangeTest.java
```diff
@@ -35,20 +32,18 @@
...
 import static accord.Utils.*;
 import static accord.impl.IntKey.keys;
 import static accord.impl.IntKey.range;
 import static accord.local.PreLoadContext.empty;
+import static accord.primitives.Routable.Domain.Key;
+import static accord.primitives.Txn.Kind.Write;
 
 public class TopologyChangeTest
 {
     private static TxnId coordinate(Node node, Keys keys) throws Throwable
```
iamaleksey: Unused method?

accord-core/src/main/java/accord/primitives/TxnId.java
```diff
@@ -22,13 +22,42 @@
 
 public class TxnId extends Timestamp
 {
-    public TxnId(Timestamp timestamp)
+    public static TxnId fromBits(long msb, long lsb, Id node)
     {
-        super(timestamp);
+        return new TxnId(msb, lsb, node);
+    }
```
iamaleksey, line 33: Nit: newline between methods?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -295,12 +292,12 @@ public AcceptOutcome accept(SafeCommandStore safeStore, Ballot ballot, Kind kind
             return AcceptOutcome.Redundant;
         }
 
-        if (known().isDefinitionKnown() && !kind().equals(kind))
+        if (known().isDefinitionKnown() && !txnId().rw().equals(kind))
```
iamaleksey: ==?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -977,7 +973,7 @@ else if (existingRanges != additionalRanges && !route().covers(additionalRanges)
         if (!validate(ensurePartialTxn, existingRanges, additionalRanges, covers(partialTxn()), covers(partialTxn), "txn", partialTxn))
             return false;
 
-        if (partialTxn != null && kind() != null && !kind().equals(partialTxn.kind()))
+        if (partialTxn != null && !txnId().rw().equals(partialTxn.kind()))
```
iamaleksey: ==?

accord-core/src/main/java/accord/primitives/TxnId.java
```diff
@@ -22,13 +22,42 @@
...
+    {
+        super(epoch, hlc, flags, node);
+    }
+
+    private TxnId(long msb, long lsb, Id node)
+    {
+        super(msb, lsb, node);
+    }
+
+    public Txn.Kind rw()
```
iamaleksey: I would consider adding `isWrite()` and `isRead()` to `TxnId` to further shorten call chains like `txnId.rw().isWrite()` in e.g. `InMemoryCommandsForKey` by one.

accord-core/src/main/java/accord/primitives/Ballot.java
```diff
@@ -22,15 +22,35 @@
...
         super(from);
     }
 
-    public Ballot(long epoch, long real, int logical, Id node)
+    Ballot(long epoch, long hlc, int flags, Id node)
+    {
+        super(epoch, hlc, flags, node);
+    }
+
+    private Ballot(long msb, long lsb, Id node)
```
iamaleksey: OCD: give both bottom two constructors same visibility? public, private, or package-private, doesn't matter.

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -19,62 +19,126 @@
...
-        return new Timestamp(epoch, real, logical + 1, node);
+        long lsb = this.lsb + HLC_INCR;
+        long msb = this.msb;
+        if (Long.compareUnsigned(this.lsb, lsb) > 0)
+            ++msb; // overflow of lsb
+        return new Timestamp(msb, lsb, node);
     }
 
     @Override
     public int compareTo(Timestamp that)
```
iamaleksey: `@NotNull that` to makе IDEA happy?

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -19,62 +19,126 @@
...
+    {
+        return (msb << 48) | (lsb >>> 16);
+    }
+
+    public int flags()
+    {
+        return (int) (lsb & MAX_FLAGS);
+    }
+
     public Timestamp withMinEpoch(long minEpoch)
```
iamaleksey: Maybe rename to `withEpochAtLeast()` or something along those lines? My initial understanding was: with min( epoch) out of {current, provided}, but it does the opposite of course.

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -19,62 +19,126 @@
...
 
     @Override
     public int compareTo(Timestamp that)
     {
         if (this == that) return 0;
-        int c = Long.compare(this.epoch, that.epoch);
-        if (c == 0) c = Long.compare(this.real, that.real);
-        if (c == 0) c = Integer.compare(this.logical, that.logical);
+        int c = Long.compareUnsigned(this.msb, that.msb);
+        if (c == 0) c = Long.compareUnsigned(this.lsb, that.lsb);
```
iamaleksey, line 139: Arbitrary as it is, don't we want to compare `Node.Id` before `flags`?

iamaleksey: To allow byte-order comparability (if ever relevant), when `Node.Id` becomes `int`, can spread it between `lsb` and the extra int, moving `flags` to the end of the logical byte-sequence.

iamaleksey (changes requested):
Basically LGTM. Some unresolved suggestions in Conversation section still. All minor, none binding, except for maybe enum comparisons using `equals()`.

accord-core/src/main/java/accord/primitives/Timestamp.java
belliottsmith: I've opted to ignore the flag bits entirely for identity purposes, to permit us to use them as carrier bits for things that do not relate 1:1 with the identity bits. We can move their position around later.

belliottsmith:
Thanks for the swift review. I think I've integrated every suggestion, but let me know if I missed anything.

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -19,62 +19,145 @@
...
+    }
+
+    public static Timestamp fromValues(long epoch, long hlc, int flags, Id node)
+    {
+        return new Timestamp(epoch, hlc, flags, node);
+    }
+
+    public static Timestamp maxForEpoch(long epoch)
+    {
+        return new Timestamp((epoch << 16) | 0xffff, Long.MAX_VALUE, Id.MAX);
```
iamaleksey: `epochMsb(epoch)` for consistency?

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -19,62 +19,145 @@
...
+    }
+
+    public static Timestamp minForEpoch(long epoch)
+    {
+        return new Timestamp(epoch << 16, 0, Id.NONE);
```
iamaleksey: `epochMsb(epoch)` for consistency?

accord-core/src/main/java/accord/primitives/Timestamp.java
iamaleksey: Also, `MAX_FLAGS`?

iamaleksey (approved):
LGTM

PR #23, opened 2023-01-06
Introduce range transactions, including burn test support for for read-only transactions

belliottsmith:


accord-core/src/main/java/accord/coordinate/CheckOn.java
```diff
@@ -136,21 +138,28 @@ class OnDone implements MapReduceConsume<SafeCommandStore, Void>
 
         public OnDone()
         {
-            Ranges localRanges = node.topology().localRangesForEpochs(txnId.epoch, untilLocalEpoch);
-            PartialRoute<?> selfRoute = route().slice(localRanges);
+            Ranges sliceRanges = node.topology().localRangesForEpochs(txnId.epoch(), untilLocalEpoch);
+            PartialRoute<?> selfRoute = route().slice(sliceRanges, Maximal);
```
aweisberg: Maximal is `/** Overlapping ranges are extended to the union of the overlaps */`

What does union of the overlaps mean? It seems like this behavior can cause the route to include ranges not in `sliceRanges`? Why is that useful vs just where they intersect?

accord-core/src/main/java/accord/coordinate/CheckOn.java
```diff
@@ -136,21 +138,28 @@ class OnDone implements MapReduceConsume<SafeCommandStore, Void>
...
+            if (selfRoute.domain().isRange())
+            {
+                // for range queries, we want to "cover" the entirety of every range
```
aweisberg: Why do we want to cover the entire thing if we don't replicate the data?

accord-core/src/main/java/accord/coordinate/ReadCoordinator.java
```diff
@@ -16,7 +16,7 @@
 
 abstract class ReadCoordinator<Reply extends accord.messages.Reply> extends ReadTracker implements Callback<Reply>
 {
-    private static final boolean DEBUG = false;
+    private static final boolean DEBUG = true;
```
aweisberg: Don't forget to turn this off

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -181,7 +197,14 @@ public NodeTimeService time()
 
         private Timestamp maxConflict(Seekables<?, ?> keysOrRanges, Ranges slice)
         {
-            return mapReduce(keysOrRanges, slice, CommandsForKey::max, (a, b) -> a.compareTo(b) >= 0 ? a : b, Timestamp.NONE);
+            Timestamp timestamp = mapReduceForKey(keysOrRanges, slice, (forKey, prev) -> Timestamp.max(forKey.max(), prev), Timestamp.NONE, null);
```
aweisberg, line 203: Some commands are indexed by key, some are indexed by range so we compare with every range command for now to find the max conflicting timestamp?

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -214,36 +237,176 @@ public void forCommittedInEpoch(Ranges ranges, long epoch, Consumer<Command> con
...
                 default:
                     throw new AssertionError();
                 case Key:
                     AbstractKeys<Key, ?> keys = (AbstractKeys<Key, ?>) keysOrRanges;
-                    return keys.stream()
-                            .filter(slice::contains)
-                            .map(this::commandsForKey)
-                            .map(map)
-                            .reduce(initialValue, reduce);
+                    for (Key key : keys)
```
aweisberg, line 385: De-streaming for performance?

accord-core/src/main/java/accord/impl/InMemoryCommandsForKey.java
```diff
@@ -74,25 +74,27 @@ public boolean isEmpty()
...
-            return commands.tailMap(timestamp, false).values().stream()
-                    .filter(cmd -> testKind == RorWs || cmd.txnId().isWrite())
+            for (Command cmd : (testTimestamp == TestTimestamp.BEFORE ? commands.headMap(timestamp, false) : commands.tailMap(timestamp, false)).values())
+            {
+                if (testKind == Ws && cmd.txnId().isRead()) continue;
                     // If we don't have any dependencies, we treat a dependency filter as a mismatch
-                    .filter(cmd -> testDep == ANY_DEPS || (cmd.known().deps != DepsUnknown && (cmd.partialDeps().contains(depId) ^ (testDep == WITHOUT))))
-                    .filter(cmd -> TestStatus.test(cmd.status(), testStatus, status))
-                    .map(map);
+                if (testDep != ANY_DEPS && (cmd.known().deps == DepsUnknown || (cmd.partialDeps().contains(depId) != (testDep == WITH))))
```
aweisberg: Subtle difference with how this is done in `InMemoryCommandStore` for range transactions is `hasProposedOrDecidedDeps` vs `deps == DepsUnkown` which will behave differently for `NoDeps`.

Should `hasProposedOrDecidedDeps` return true for `NoDeps`?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -292,30 +294,28 @@ public AcceptOutcome accept(SafeCommandStore safeStore, Ballot ballot, Kind kind
             return AcceptOutcome.Redundant;
         }
 
-        if (known().isDefinitionKnown() && txnId().rw() != kind)
```
aweisberg, line 295: Kind is not worth plumbing through everywhere for just this check?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -601,14 +603,18 @@ private boolean maybeExecute(SafeCommandStore safeStore, ProgressShard shard, bo
...
                 setStatus(ReadyToExecute);
                 logger.trace("{}: set to ReadyToExecute", txnId());
                 safeStore.progressLog().readyToExecute(this, shard);
                 notifyListeners(safeStore);
                 break;
 
             case PreApplied:
-                if (executeRanges(safeStore, executeAt()).intersects(writes().keys))
+                Ranges executeRanges = executeRanges(safeStore, executeAt());
+                boolean intersects = switchOnDomain(writes().keys, executeRanges, AbstractKeys::intersects, AbstractRanges::intersects);
```
aweisberg: Why is `switchOnDomain` necessary? Can intersection be done without having the caller use `switchOnDomain`?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -911,57 +917,50 @@ enum EnsureAction { Ignore, Check, Add, TrySet, Set }
...
-    private boolean validate(Ranges existingRanges, Ranges additionalRanges, ProgressShard shard,
-                             Route<?> route, EnsureAction ensureRoute,
+    private boolean validate(Ranges existingRanges, Ranges additionalRanges,
+                             ProgressShard shard, Route<?> route, EnsureAction ensureRoute,
+                             @Nullable PartialTxn partialTxn, EnsureAction ensurePartialTxn,
+                             @Nullable PartialDeps partialDeps, EnsureAction ensurePartialDeps)
+    {
+        return validate(existingRanges, additionalRanges, additionalRanges, shard, route, ensureRoute, partialTxn, ensurePartialTxn, partialDeps, ensurePartialDeps);
+    }
+    private boolean validate(Ranges existingRanges, Ranges additionalRouteRanges, Ranges additionalTxnRanges,
```
aweisberg: Additional txn ranges looks like it is always `Ranges.EMPTY` or the same as `acceptRanges`

It's also really not clear what it is. `validate` could use documentation of what its parameters are.

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -363,7 +363,7 @@ public TxnId nextTxnId(Txn.Kind rw, Domain domain)
 
     public Future<Result> coordinate(Txn txn)
     {
-        return coordinate(nextTxnId(txn.kind(), Key), txn);
+        return coordinate(nextTxnId(txn.kind(), txn.keys().domain()), txn);
```
aweisberg, line 369: Surprised to see domain here that is only used once in `SomeRoute`. The home key itself determines the domain so I don't see why the TxnId needs it?

accord-core/src/main/java/accord/local/PreLoadContext.java
```diff
@@ -33,11 +34,19 @@
 {
     /**
      * @return ids of the {@link Command} objects that need to be loaded into memory before this operation is run
+     *
+     * TODO (expected): this is used for Apply, NotifyWaitingOn and listenerContexts; others only use a single txnId
+     *  firstly, it would be nice to simply have that txnId as a single value.
+     *  In the case of Apply, we can likely avoid loading all
```
aweisberg: all transactions?

accord-core/src/main/java/accord/messages/Accept.java
```diff
@@ -39,51 +39,55 @@
...
         this.executeAt = executeAt;
         this.keys = keys;
         this.partialDeps = partialDeps;
-        this.kind = kind;
     }
 
     @Override
     public synchronized AcceptReply apply(SafeCommandStore safeStore)
     {
+        if (minEpoch < txnId.epoch())
```
aweisberg: How does `minEpoch` end up < `txnId.epoch` if we calculate `minEpoch` from a `Topologies` that was itself calculated using `TxnId.Epoch`?

accord-core/src/main/java/accord/messages/Accept.java
```diff
@@ -39,51 +39,55 @@
...
+        {
+            Ranges acceptRanges = safeStore.ranges().between(txnId.epoch(), executeAt.epoch());
+            if (!acceptRanges.intersects(scope))
+                return new AcceptReply(calculatePartialDeps(safeStore));
```
aweisberg, line 86: Why are the locally known deps relevant if none of the scope of the txn was replicated locally during the transactions epochs?

accord-core/src/main/java/accord/messages/BeginInvalidation.java
```diff
@@ -10,6 +10,7 @@
 import java.util.Collections;
 import java.util.List;
 
+import static accord.primitives.Routables.Slice.Overlapping;
```
aweisberg: Unused

accord-core/src/main/java/accord/messages/CheckStatus.java
```diff
@@ -37,6 +37,7 @@
 import static accord.local.Status.*;
 import static accord.local.Status.Durability.NotDurable;
 import static accord.messages.TxnRequest.computeScope;
+import static accord.primitives.Routables.Slice.Overlapping;
```
aweisberg: Unused

accord-core/src/main/java/accord/messages/Commit.java
```diff
@@ -36,6 +36,7 @@
 
 import static accord.local.Status.Committed;
 import static accord.local.Status.Known.DefinitionOnly;
+import static accord.primitives.Routables.Slice.Overlapping;
```
aweisberg: Unused

accord-core/src/main/java/accord/messages/GetDeps.java
```diff
@@ -11,6 +11,7 @@
 import java.util.Collections;
 
 import static accord.messages.PreAccept.calculatePartialDeps;
+import static accord.primitives.Routables.Slice.Overlapping;
```
aweisberg: Unused

accord-core/src/main/java/accord/messages/PreAccept.java
```diff
@@ -93,13 +92,21 @@ public PreAcceptReply apply(SafeCommandStore safeStore)
         // note: this diverges from the paper, in that instead of waiting for JoinShard,
         //       we PreAccept to both old and new topologies and require quorums in both.
         //       This necessitates sending to ALL replicas of old topology, not only electorate (as fast path may be unreachable).
+        if (minEpoch < txnId.epoch() && !safeStore.ranges().at(txnId.epoch()).intersects(scope))
+        {
+            // we only preaccept in the coordination epoch, but we might contact other epochs for dependencies
```
aweisberg, line 98: confused how there can be multiple epochs at preaccept since the `executeAt` epoch hasn't been determined yet? Will keep reading.

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -39,6 +39,7 @@
 import static accord.messages.ReadData.ReadNack.NotCommitted;
 import static accord.messages.ReadData.ReadNack.Redundant;
 import static accord.messages.TxnRequest.*;
+import static accord.primitives.Routables.Slice.Overlapping;
```
aweisberg: Unused

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -256,7 +257,7 @@ public Iterable<TxnId> txnIds()
     @Override
     public Seekables<?, ?> keys()
     {
-        return readScope;
+        return Keys.EMPTY;
```
aweisberg, line 259: Nothing to preload? Will keep reading

accord-core/src/main/java/accord/messages/TxnRequest.java
```diff
@@ -36,6 +36,7 @@
 import accord.topology.Topologies;
 import accord.topology.Topology;
 
+import static accord.primitives.Routables.Slice.Overlapping;
```
aweisberg: Unused

accord-core/src/main/java/accord/messages/WaitOnCommit.java
```diff
@@ -29,6 +29,7 @@
 import org.slf4j.LoggerFactory;
 
 import static accord.local.Status.Committed;
+import static accord.primitives.Routables.Slice.Overlapping;
```
aweisberg: Unused

accord-core/src/main/java/accord/primitives/AbstractRanges.java
```diff
@@ -210,9 +246,102 @@ static <RS extends AbstractRanges<?>, P> RS slice(Ranges covering, AbstractRange
...
+                Range l = covering.ranges[li], r = input.ranges[ri]; // l(eft), r(right)
+                RoutingKey ls = l.start(), rs = r.start(), le = l.end(), re = r.end(); // l(eft),r(ight) s(tart),e(nd)
+                int cs = rs.compareTo(ls), ce = re.compareTo(le); // c(ompare) s(tart),e(nd)
+                if (cs <= 0 && ce >= 0)
+                {
+                    buffer[bufferCount++] = r;
+                }
+                else
+                {
+                    buffer[bufferCount++] = r.subRange(cs <= 0 ? rs : ls, ce >= 0 ? re : le);
```
aweisberg: It was confusing that `subRange` constructs a new range, but it isn't actually a `subRange` of the range you invoked `subRange` on.

accord-core/src/main/java/accord/primitives/Deps.java
```diff
@@ -18,1311 +18,220 @@
...
         };
     }
 
-    private int startOffset(int keyIndex)
-    {
-        return keyIndex == 0 ? keys.size() : keyToTxnId[keyIndex - 1];
-    }
-
-    private int endOffset(int keyIndex)
+    public List<TxnId> txnIds(Seekable keyOrRange)
```
aweisberg: Could use Iterable instead of List and then not have to materialize

accord-core/src/main/java/accord/primitives/Deps.java
```diff
@@ -18,1311 +18,220 @@
...
     {
-        array[offset + index / 32] |= (1 << (index & 31));
+        return this == that || (that instanceof Deps && equals((Deps)that));
     }
 
-    private static boolean hasSetBit(int[] array, int offset, int index)
+    public boolean equals(Deps that)
     {
-        return (array[offset + index / 32] & (1 << (index & 31))) != 0;
+        return this.keyDeps.equals(that.keyDeps) && this.rangeDeps.equals(that.rangeDeps);
```
aweisberg, line 238: Add hash code throwing unsupported operation exception or the actual hash code?

accord-core/src/main/java/accord/primitives/Ranges.java
```diff
@@ -89,10 +96,29 @@ public Ranges union(Ranges that)
...
+    {
+        return with((AbstractRanges<?>) that);
+    }
+
+    public Ranges with(Ranges that)
+    {
+        return union(MERGE_OVERLAPPING, that);
+    }
+
+    public Ranges with(AbstractRanges<?> that)
```
aweisberg, line 98: Why have multiple overlapping versions of the same thing? `union` ends up being over` AbstractRanges`. Is it some kind of monomorphic call site thing?

accord-core/src/main/java/accord/primitives/Ranges.java
```diff
@@ -122,6 +150,14 @@ public Ranges union(UnionMode mode, Ranges that)
         });
     }
 
+    public Ranges union(UnionMode mode, AbstractRanges<?> that)
+    {
+        return union(mode, this, that, this, that, (left, right, ranges) -> {
+            if (ranges == left.ranges) return left;
```
aweisberg, line 144: This doesn't have `if (ranges == right.ranges) return right`

I assume this is some case where left and right are already the same contents?

Or is it because we only know `left` is a `Ranges`?

accord-core/src/main/java/accord/primitives/Routables.java
```diff
@@ -55,22 +81,61 @@
...
     @Inline
     static <Input extends Routable, T> T foldl(Routables<Input, ?> inputs, AbstractRanges<?> matching, IndexedFold<? super Input, T> fold, T initialValue)
     {
         return Helper.foldl(Routables::findNextIntersection, Helper::findLimit, inputs, matching, fold, initialValue);
     }
 
+    @Inline
+    static <T> T foldlMinimal(Seekables<?, ?> inputs, AbstractRanges<?> matching, IndexedFold<? super Seekable, T> fold, T initialValue)
+    {
+        return Helper.foldlMinimal(inputs, matching, fold, initialValue);
```
aweisberg, line 106: Could really use an explanation for what minimal means.

There are 8 different wrappers and 4 different implementations in Helper. At this point they need documentation (implementation at least) just to speed up knowing what they do.

accord-core/src/main/java/accord/primitives/Unseekables.java
```diff
@@ -45,7 +49,7 @@ public boolean isFullRoute()
                     return right;
 
                 // non-route types can always accept route types as input, so just call its union method on the other
-                return leftKind.isRoute() ? ((Unseekables)right).union(left) : ((Unseekables)left).union(right);
+                return left.with(right);
```
aweisberg, line 61: Up above where it says `// non-route types can always accept route types as input, so just call its union method on the other`

How has it established that `left` is not a route?

Really it looks like `with` just doesn't care. I'm not even clear on what the difference between `with` and `union` so maybe it needs a name change or documentation?

aweisberg: I just realized `union`, `with`, and `merge` are all the same thing?

Why `with` here and `union` at the end of the function?

accord-core/src/main/java/accord/impl/InMemoryCommandsForKey.java
```diff
@@ -108,15 +110,16 @@ public Stream<Command> all()
 
     // TODO (now): add validation that anything inserted into *committedBy* has everything prior in its dependencies
     private final Key key;
-    private final InMemoryCommandTimeseries<TxnIdWithExecuteAt> uncommitted = new InMemoryCommandTimeseries<>(cmd -> new TxnIdWithExecuteAt(cmd.txnId(), cmd.executeAt()));
```
aweisberg, line 111: I got to the end and never figured out why `uncommitted` could be replaced by `byId`. Does the `TestTimestamp` take care of that now?

aweisberg (commented):
Didn't quite finish or get to go back and clean up my comments/questions. I'll post them now so you can look at them during the time zone difference.

accord-core/src/main/java/accord/impl/InMemoryCommand.java
```diff
@@ -129,7 +131,7 @@ public Route<?> route()
     @Override
     protected void setRoute(Route<?> route)
     {
-        this.route = route;
+        this.route = Invariants.checkArgument(route, !route.isEmpty());
```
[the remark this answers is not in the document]
belliottsmith: I weakly prefer leaving `Invariants` in for invariant checks, but I suppose for `checkArgument` it's less necessary. But I like consistency too 🤷

accord-core/src/main/java/accord/coordinate/ReadCoordinator.java
belliottsmith: I think it's fine to leave it on for now in general tbh, it's a while until we go to prod. But, happy to flip it off if you prefer.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
belliottsmith: Basically, efficiency be damned for now yeah.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
belliottsmith: Just for clarity, as we short-circuit and also have two parameters to our map now.

accord-core/src/main/java/accord/impl/InMemoryCommandsForKey.java
belliottsmith: Good catch. This should also use `hasProposedOrDecidedDeps()`

accord-core/src/main/java/accord/local/Command.java
belliottsmith: It used to be necessary not for this check, but for correctness and to register with `CommandsForKey`.

accord-core/src/main/java/accord/local/Command.java
belliottsmith: You're right I think this is no longer necessary (used to be due to `hashIntersects`)

accord-core/src/main/java/accord/local/Node.java
belliottsmith: We don't necessarily know the `homeKey`

accord-core/src/main/java/accord/messages/Accept.java
belliottsmith: During range movements, we need to adopt dependencies on transactions from earlier epochs. This is part of what we discussed before about how we have diverged from the paper: in the paper new replicas do not participate in decisions until they have sufficient information to respond. However, today we instead seek "votes" from every epoch that has not fully migrated to the new replicas, from whom we only need the dependencies they know of.

accord-core/src/main/java/accord/messages/Accept.java
belliottsmith: Same reason: partially migrated ownership needs to seek info from earlier epochs. I would be more than open to modifying this to restore the paper's behaviour, to simplify this.

accord-core/src/main/java/accord/messages/PreAccept.java
belliottsmith: Same reason around epoch migration. Again, open to simplifying this by having new replicas wait until they can safely respond to process these messages, as that is my preferred approach also.

accord-core/src/main/java/accord/primitives/AbstractRanges.java
belliottsmith: Open to renaming this method. `newRange`?

accord-core/src/main/java/accord/primitives/Deps.java
belliottsmith: The list is materialised in C* side to keep in cache (for now) anyway. Might be something to look at in the near future as we optimise caching, as we shouldn't be keeping this collection in memory.

accord-core/src/main/java/accord/primitives/Ranges.java
belliottsmith: I think it was just to make the compiler happy without having to cast to `AbstractRanges` somewhere, I think because `Ranges` implements `Unseekables` while extending `AbstractRanges`

accord-core/src/main/java/accord/primitives/Ranges.java
belliottsmith: The latter. If it were we'd likely have called the other override.

accord-core/src/main/java/accord/primitives/Unseekables.java
belliottsmith: Not quite, I will document them more clearly. Though the distinction is only relevant for `Route`, for which `union` will return another `Route` whereas `with` will return an `Unseekables`. `union` definitely contains more information than either parameter, whereas `with` may downgrade from a `Route` in order to safely expand the `Unseekable` that are contained in either.

accord-core/src/main/java/accord/impl/InMemoryCommandsForKey.java
belliottsmith: and `minStatus`/`maxStatus`. Note this is all an implementation detail now, and perhaps in final implementation we will use just one collection so we can pack more into memory.

accord-core/src/main/java/accord/messages/ReadData.java
belliottsmith: We don't need the per-key data, it was for a C* impl reason that I determined was unnecessary - namely selecting a `nowInSeconds` that was consistent with the transaction's write `timestamp` (which is fine to be slightly stale, so can avoid bringing that info into memory)

belliottsmith: Note, I plan to add some stuff to this PR (or perhaps if you prefer a followup PR) to permit us to properly implement range transactions in the C* side. I had left it initially because range transactions are inherently more complicated for this element as they do not know which keys they want - they want ranges of keys, and this may be more info than we want/can bring into memory. So I have (almost) refactored this so that we do not need to declare the `keys` scope at all for `PreLoadContext` - we instead invoke an asynchronous method at the end of processing a message that provides the keys we need to operate over and it's up to the impl how it goes about that - which permits us to consistently operate over ranges as well.

accord-core/src/main/java/accord/primitives/Unseekables.java
belliottsmith: It might be that we can actually get rid of `merge` now and just use `with`. I am open to better names.

We could also get rid of `union` from `Routables`, `Unseekables` and `Seekables`, leaving it only for `Route`, and have only `with` in the other three. I will prepare this change and see how we like it.

belliottsmith: Ok, so I have kept all three, but modified the hierarchies and documented the differences. The idea behind `merge` is that it uses the strongest merge possible, so `union` if possible and `with` otherwise, as now defined.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: So this one is a little interesting, and I am open to suggestions. Essentially this is to make our invariants happy for ranges, but it could lead to fetching more data than we want. Basically I choose never to slice a `Range` that is one of the keys of a transaction to ensure identity is never broken. Each replica would slice it differently so we'd have a mess stitching them back together. But when we come to reconstitute a transaction, we don't fully cover the keys if we only partially intersect one of them. 

Perhaps we should modify the invariants of `reconstitutePartial` and the objects that make it up. I will take another look to refresh my memory.

accord-core/src/main/java/accord/local/Command.java
belliottsmith: You'll note the distinction only exists for `accept`, basically, where we prefer to also update our `Route` to include the ranges we would be involved in but we have no additional transaction data to replicate anyway. 

This isn't strictly necessary, as I have changed how we register ourselves with the `CommandStore` so that this `Route` isn't necessary for anything right now. Though it might help avoid future bugs in C* when forwarding data to other nodes, with this `CommandStore` having dependency data for keys not in the `Route`. That said, since we track dependency data by `Seekable/Key` maybe this isn't a concern either, and we should roll this change back and restore the prior documentation, and just not that in the case of `Accept` we might be registered against keys not implicitly covered by the `Route`.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: Ok, so this `Maximal` wasn't necessary, and in fact getting a `Maximal` _`Route`_ is probably nonsensical and dangerous so I have modified `slice` to return only a basic collection (non-Route) if a specific `Slice` is requested. I haven't removed `Maximal` as a feature for now as it might be easier to use this later inside of `PartialTxn` etc. when either slicing or reconstituting to define `covers`, though open to deleting it.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: We want `covers` to cover the bit we're slicing and the `Route` so that our invariants hold. We're only slicing here though, so not really doing any extra work.

Perhaps later we can look into slicing the `Route` to a `Minimal` one so we only seek information from the minimum number of replicas to answer our own local queries. I will add a big TODO.

belliottsmith:
I have added one large follow-up commit that helps support proper range transaction support inside of Cassandra. This required a bit of refactoring so that `calculateDeps` can be asynchronous, as plausibly a large range might require fetching state from disk that could not all readily be held in disk at-once. Also, presently, we don't have any way of knowing what ranges are in-memory and what ranges are not (though we will want to address that in performance work anyway).

Let me know if you would rather spin that addition into a separate patch.

accord-core/src/main/java/accord/coordinate/ReadCoordinator.java
aweisberg: Ah, no worries I just assumed you originally had it off and would want it back off.

accord-core/src/main/java/accord/impl/InMemoryCommand.java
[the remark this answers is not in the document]
aweisberg: For something as common as Invariants/Preconditions it seems like it adds a lot of indentation and visual noise, but doesn't bear much information.

One thing I notice about Invariants vs Preconditions is that Preconditions lets you include a string so the exception actually says what was wrong about the state or the null value without having to hunt down the exact line of code from the exact right version of the file.

I am not worked up about consistency, it's more potato potato for me.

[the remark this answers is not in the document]
belliottsmith: Well, we can add t whenever we want it

accord-core/src/main/java/accord/local/Node.java
aweisberg: Sure, but the one case where it is used is
```
        if (homeKey() != null)
            return switchOnDomain(txnId().domain(), homeKey(), PartialKeyRoute::empty, PartialRangeRoute::empty);
```
Since it is in flags sure it's basically free to have though.

accord-core/src/main/java/accord/primitives/Routables.java
```diff
@@ -55,62 +81,147 @@
     /**
      * Perform {@link SortedArrays#exponentialSearch} from {@code thisIndex} looking for {@code find} with behaviour of {@code search}
      */
-    int findNext(int thisIndex, K find, SortedArrays.Search search);
+    int findNext(int thisIndex, RoutableKey find, SortedArrays.Search search);
+
+    Domain domain();
+
+    @Inline
+    static <Param, Output> Output switchOnDomain(Routables<?, ?> inputs, Param param, BiFunction<AbstractKeys<?, ?>, Param, Output> ifKeys, BiFunction<AbstractRanges<?>, Param, Output> ifRanges)
```
aweisberg: This is unused now?

accord-core/src/main/java/accord/primitives/Routables.java
```diff
@@ -55,62 +81,147 @@
...
+        {
+            default: throw new AssertionError();
+            case Key: return ifKeys.apply((AbstractKeys<?, ?>) inputs, param);
+            case Range: return ifRanges.apply((AbstractRanges<?>) inputs, param);
+        }
+    }
 
-    Routable.Domain kindOfContents();
+    @Inline
+    static <Param, Output> Output switchOnDomain(Domain domain, Param param, Function<Param, Output> ifKeys, Function<Param, Output> ifRanges)
```
aweisberg: This is used once to produce either an empty `PartialKeyRoute` or `PartialRangeRoute`. Seems like it could just be a default method `toPartialRoute` in the `PartialRouteInterface`?

accord-core/src/main/java/accord/messages/ReadData.java
aweisberg: More smaller PRs is better for me.

Makes sense that we would want to be able to process incrementally for ranges and large transactions.

accord-core/src/main/java/accord/primitives/AbstractRanges.java
aweisberg: Yes please.

accord-core/src/main/java/accord/local/Command.java
belliottsmith: Let me know if you have an opinion here. Might be easier to just rollback, but don't have a strong opinion.

accord-core/src/main/java/accord/local/Node.java
belliottsmith: Hmm. I guess we keep `RangeDeps` and `KeyDeps` separate so this is actually implicit. So perhaps we don't need it.

accord-core/src/main/java/accord/messages/ReadData.java
belliottsmith: Ok, I'll revert those additional commits and file a new PR

accord-core/src/main/java/accord/primitives/Routables.java
belliottsmith: Have added javadoc

aweisberg:
I am a little mixed up on several things so to try and clarify. At each shard in terms of metadata about transaction execution over large numbers of transactions (as opposed to fetch this specific one) we need to materialize to schedule a transaction.

1. We need the max conflict to response with an `executeAt` during `PreAccept`
2. We need to calculate the dependency `TxnId`s that are < this `executeAt` during `PreAccept` or whenever trying to supply dependencies
3. Calculate the timestamp to use as `nowInSeconds` for TTLs
4. Calculate the timestamp to use when inserting as well as processing TTLs on the data we are inserting

3 and 4 maybe we don't need to look at other transactions to calculate? I had asked some questions about that earlier.

1 we can calculate pessimistically using an in-memory summary
2 we might actually need to materialize a large number of IDs in a range transaction?

We are going to materialize the entire set dependencies in memory at the coordinator to distribute to the replicas of each shard so there are some hard limits to how big those can practically get.

So are we trying to be able to handle large amounts of dependencies better (which is memory bound) or are we trying to better handle smaller amounts that aren't in memory and need to be read off of disk?

Even if we make this async isn't the shard blocked waiting on disk because of execution order constraints? Introducing a transaction with dependency data not in memory blocks subsequent transactions either way.

Or is what we are optimizing for here keeping the shard thread unblocked so it can continue to work on things that are ready to work on and not blocked by execution order constraints?

accord-core/src/main/java/accord/primitives/Unseekables.java
aweisberg: `UnionMode.mergeAdjacent` doesn't seem to be used.

At the moment it makes sense. There are a lot of methods for merging these things, but also a lot of them have little bits of specialization for doing things like avoiding constructing a new instance.

accord-core/src/main/java/accord/primitives/Routables.java
aweisberg: 3 of the folds that are marked unused in Accord also seem to be unused in Cassandra. For range folding documenting what an indexed fold is would be helpful.

One of the range folds also looks like it might be unused.

`foldlMissing` what is the missing aspect?

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -214,36 +248,253 @@ public void forCommittedInEpoch(Ranges ranges, long epoch, Consumer<Command> con
...
+                            if (accumulate.equals(terminalValue))
+                                return accumulate;
+                        }
+                    }
             }
+            return accumulate;
         }
 
-        public void forEach(Routables<?, ?> keysOrRanges, Ranges slice, Consumer<CommandsForKey> forEach)
+        public void forEach(Seekables<?, ?> keysOrRanges, Ranges slice, Consumer<CommandsForKey> forEach)
```
aweisberg: Should the `foreach` be private now?

belliottsmith:
Are we discussing here the refactor to make `calculateDeps` asynchronous? In large part that is to address the difficulty of caching for range transactions, as a range transaction would need to page in from disk the entire active "ForKey" (and equivalent for range) contents for all the covered ranges, and lock the records in cache.

Right now our cache is super inefficient so this is loads of data, though in fairness I hope our cache will end up much smaller - in which case perhaps it is fine to declare that the contents must be fully materialised and locked in cache. I figured it might anyway be better to permit paging through to avoid churn to the cache, but perhaps it does not justify the clarity penalty. I think our cache will need to be range-aware to be able to maintain simple summaries for answering (1), and it might be enough to maintain a small array of longs per key/range representing the TxnId and some minimal bookkeeping.

The refactor does have some other advantages, though, namely that for the majority of queries we may have to do less preparatory work, as we will expect the queries to be answered from cache anyway, and we will only go asynchronous when the cache is not sufficiently populated (and can resume the task as and when data is read from disk). This might simplify other elements of the code for cache management and scheduling, I'm not sure.

Note, that at least for _Recovery_ we will need to support asynchronous dependency calculations as we need to look at the dependencies of these records, which we _definitely_ do not want to keep in cache (even if we do currently).

For 3 and 4, I think we can track range-level info about whether we need to consult per-key level data, but while TTL can be ignored entirely I think, timestamp cannot as the Accord `Timestamp` is more than 8 bytes, and we could have two transactions execute with the same `hlc` value proposed by different nodes. So we do need some mechanism for resolving this. But this eventuality should be exceptionally rare, and a range map tracking regions or keys where this might be true should be sufficient even if the cache is paged out for that range.

> isn't the shard blocked waiting on disk because of execution order constraints

Why? Only transactions involving the `txnId` will be blocked, not the whole shard - and at least for the range transactions for topology movements no transactions will block execution on them anyway (they will be treated specially)

accord-core/src/main/java/accord/local/Command.java
aweisberg: I don't think I understand how registration changed beyond removing `uncommitted`? It looks basically the same except for the addition of `rangeCommands` to support ranges.

I also need to better understand what you mean by ranges we don't replicate, you mean ranges from older epochs?

Which route isn't necessary, the one passed to accept that is then passed to `Command.set`?

belliottsmith:
Another approach I wondered about would be creating pseudo command stores and permitting the message to be part processed for a slice of the command store that has been loaded. Perhaps that would be cleaner, if we want such a facility, and perhaps for now it would be better to punt on this until we are able to address caching, and for now just page/lock everything in as needed.

accord-core/src/main/java/accord/primitives/Routables.java
belliottsmith: I would prefer not to prune the fold functions for now, as the specific ones in use vary a great deal between patches, so we may end up just re-implementing and reviewing them.

I documented foldMissing at the public API, but I can copy the documentation over to the Helper methods too if you like.

accord-core/src/main/java/accord/primitives/Unseekables.java
belliottsmith: I would prefer to keep the facility there, as we may want to use it in future to minimise `Ranges`. I assume I did use it before.

I usually stick to the rule of "not using it, delete it" but with code still in flux I prefer to avoid repeating work in the near future. Some of these facilities can be cleaned up before we release if they remain unused.

accord-core/src/main/java/accord/primitives/Routables.java
belliottsmith: Do you mean a static method?

accord-core/src/main/java/accord/local/Node.java
belliottsmith: Ah, `homeKey` is presently always a `RoutingKey` so this information isn't retained in the `homeKey`.

Using a home range seems problematic, as the range may be owned by multiple shards.

belliottsmith:
Ok, I have for now removed the additional changes to make `calculateDeps` asynchronous

accord-core/src/main/java/accord/local/Command.java
belliottsmith: I could have been clearer (and perhaps I forgot this at the time), but I believe this is a vestigial change, as I had briefly erroneously used the `Route` to register during `Accept` in an earlier draft of the patch (which is non-sensical, as the types aren't correct). I left it because having a more complete `Route` might all-things-considered be better _but_ I don't foresee us actually persisting the `Route` separately per-store once Aleksey's work lands, permitting us to simply reference a commit log entry and the new command state resulting from processing it, so perhaps better to just roll this back.

belliottsmith: I have removed this change

accord-core/src/main/java/accord/local/Node.java
belliottsmith: Hmm, that said, we could probably record it whenever we save the `homeKey` for some other reason. Or, we could stop accepting a `homeKey` on its own and require that we receive a `Route`.

accord-core/src/main/java/accord/primitives/Routables.java
aweisberg: Ah, got it, it's non-intersection.

accord-core/src/main/java/accord/primitives/Routables.java
aweisberg: Oops yes

accord-core/src/main/java/accord/primitives/Routables.java
belliottsmith: Could call it foldComplement if that’s clearer? Or some other equivalent term

accord-core/src/main/java/accord/messages/Accept.java
aweisberg: This needs a brief comment, explaining it's to avoid wasting time tracking Accept state when the votes isn't needed only the dependency information since we only replicate this in an earlier unsynchronized epoch.

I also suggested `minEpoch` would be clearer as `minUnsychronizedEpoch` although you were mentioning there were other aspects potentially represented in `minEpoch`

accord-core/src/main/java/accord/coordinate/CheckOn.java
aweisberg: We talked about this and my take away is that this looks like it should be a no-op because self `sliceRanges` should be a superset of `selfRoute`'s ranges, because `selfRoute` is a `route()` sliced on `sliceRanges`, but I see in the original code `slice` on `route()` was called with `Maximal` and then `Maximal` was removed and it now calls `RangeRoute.slice` which calls `AbstractRanges.slice` with `Overlapping`.

So `selfRoute` slices with `Overlapping` includes every range in the route that overlaps with any of `sliceRanges`.

Where this seems odd is that the response being given doesn't actually cover the entirety of the range from the route?  The whole purpose of tracking coverage is to preserve that knowledge so we can see what is and isn't covered and this bypasses that.

accord-core/src/main/java/accord/local/Node.java
aweisberg: Didn't 100% follow. Home key is supposed to be some `Routable` that survives topology changes so why would we care about the domain of the transaction for the home key?

When a coordinator mints a transaction there are couple of immutable things about the transaction that are known either provided by the application or generated at the coordinator. Home key is one, txnid and its components, the txn description from the application (keys etc.).

Putting it in the txnid makes sense when someone will only have the txnid, but really needs to know the domain of the transaction?

accord-core/src/main/java/accord/messages/Accept.java
aweisberg: I think it adds some extra code for both correctness and performance and it doesn't seem necessary? I think we should simplify this if we can.

What are the benefits we get from having and tracking unsynchronized epochs? Does it grant more flexibility somehow?

accord-core/src/main/java/accord/local/Node.java
belliottsmith: The home key isn't a `Routable`, it's always a `RoutingKey`

What I meant, though, was that we could stop maintaining a separate `homeKey` register in `Command`, and require that a `Route` is provided by any method that wants to provide this sort of information. This `Route` would have enough information to specify the `Domain`

accord-core/src/main/java/accord/messages/Accept.java
belliottsmith: The only advantage I can think of is that, were you to make a dramatic change to the topology, coordination wouldn't have to stop until the new replicas were up to speed. @bdeggleston wdyt? I still have a preference to have new replicas (or fast path electorate members) not answer messages until they're up to speed, as I think it simplifies this sort of thing. I could probably make the change as part of bootstrap.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: Ok, so the issue here is that we do not truncate the ranges in the `Route` we pass to `reconstitutePartial`, so when we slice the `partialTxn` to `sliceRanges` before invoking  `reconstitutePartial`, `reconstitutePartial` says we don't cover the whole range.

But, I think this is a problem with how this is modelled. I will give it some thought, perhaps for range transactions we should not be providing a `Route` to reconstitute partial, or we should also be supplying the subset of ranges.

accord-core/src/main/java/accord/local/Command.java
belliottsmith: Ok, so this change might still be desirable, or something like it. I have added logic to fail if we try to set an empty Route, as we _shouldn't_ ordinarily permit it. However, if we only participate in the `executeAt.epoch` for  `accept` (not in the `txnId.epoch`), then without this change we try to set an empty `Route` as we _don't_ want to set any txn or deps contents. We only register for the accept ranges to ensure we are adopted as a dependency.

Perhaps I could just not use the `validate` and `set` logic for `accept`. I'll think about it.

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: Ok, I've addressed this by separately slicing the keys/key ranges and the ranges they cover.

accord-core/src/main/java/accord/local/Command.java
belliottsmith: Ok, what I've done for now is to just invoke `set` with `acceptRanges` also, as it's just a slight inefficiency and otherwise a clarity improvement, and we can further improve it later. A lot of this stuff will change following the immutable state refactor and related caching improvements.

accord-core/src/main/java/accord/local/Node.java
belliottsmith: I will leave a TODO to consider this later as we address serialization/caching improvements, as I expect we will want to reform how we persist additional information like the homeKey anyway.

accord-core/src/main/java/accord/messages/Accept.java
aweisberg: Does up to speed mean on metadata or data?

I don't know how bootstrap and range movements are going to work with Accord so it's not clear to me.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
aweisberg: Still public

belliottsmith: Sorry, must have lost it in one of the rebases

accord-core/src/main/java/accord/messages/Accept.java
belliottsmith: Transaction log data, ie the list of transactions that were in flight prior to their joining

accord-core/src/main/java/accord/primitives/PartialDeps.java
```diff
@@ -55,15 +55,15 @@ public Deps reconstitute(FullRoute<?> route)
     }
 
     // PartialRoute<?>might cover a wider set of ranges, some of which may have no involved keys
```
aweisberg: This comment is out of date with type change

accord-core/src/main/java/accord/messages/Accept.java
aweisberg: @bdeggleston WDYT?

aweisberg (approved):
+1 LGTM. Thanks!

belliottsmith:
Thanks!

PR #26, opened 2023-01-17
ExclusiveSyncPoint

belliottsmith:
Introduce a special kind of `ExclusiveSyncPoint` transaction that invalidates all `TxnId` lower than it that are not witnessed by it. This permits bootstrapping nodes to ignore all `TxnId` older than this when starting their log.

### DIFF
--- a/accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
+++ b/accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
@@ -22,6 +22,7 @@ import java.util.List;
 
 import accord.local.Node;
 import accord.messages.Apply;
+import accord.messages.Commit;
 import accord.messages.PreAccept.PreAcceptOk;
 import accord.primitives.Ballot;
 import accord.primitives.Deps;
@@ -32,6 +33,7 @@ import accord.primitives.Timestamp;
 import accord.primitives.Txn;
 import accord.primitives.Txn.Kind;
 import accord.primitives.TxnId;
+import accord.topology.Topologies;
 import accord.utils.async.AsyncResult;
 
 import static accord.coordinate.Propose.Invalidate.proposeAndCommitInvalidate;
@@ -91,8 +93,12 @@ public class CoordinateSyncPoint extends CoordinatePreAccept<SyncPoint>
                 @Override
                 void onAccepted()
                 {
-                    node.send(tracker.nodes(), id -> new Apply(id, tracker.topologies(), tracker.topologies(), txnId.epoch(), txnId, route, txn, txnId, deps, txn.execute(txnId, null), txn.result(txnId, null)));
-                    accept(new SyncPoint(txnId, deps), null);
+                    Topologies commit = node.topology().preciseEpochs(route, txnId.epoch(), executeAt.epoch());
+                    Topologies latest = commit.size() == 1 ? commit : node.topology().forEpoch(route, executeAt.epoch());
+                    if (latest != commit)
+                        node.send(commit.nodes(), id -> new Commit(Commit.Kind.Maximal, id, commit.forEpoch(txnId.epoch()), commit, txnId, txn, route, null, executeAt, deps, false));
+                    node.send(latest.nodes(), id -> new Apply(id, latest, latest, executeAt.epoch(), txnId, route, txn, executeAt, deps, txn.execute(txnId, null), txn.result(txnId, null)));
+                    accept(new SyncPoint(txnId, deps, route), null);
                 }
             }.start();
         }
--- a/accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
+++ b/accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package accord.coordinate;
+
+import java.util.List;
+
+import accord.local.Node;
+import accord.messages.Apply;
+import accord.messages.PreAccept.PreAcceptOk;
+import accord.primitives.Ballot;
+import accord.primitives.Deps;
+import accord.primitives.FullRoute;
+import accord.primitives.Seekables;
+import accord.primitives.SyncPoint;
+import accord.primitives.Timestamp;
+import accord.primitives.Txn;
+import accord.primitives.Txn.Kind;
+import accord.primitives.TxnId;
+import accord.utils.async.AsyncResult;
+
+import static accord.coordinate.Propose.Invalidate.proposeAndCommitInvalidate;
+import static accord.primitives.Timestamp.mergeMax;
+import static accord.primitives.Txn.Kind.ExclusiveSyncPoint;
+import static accord.utils.Functions.foldl;
+
+/**
+ * Perform initial rounds of PreAccept and Accept until we have reached agreement about when we should execute.
+ * If we are preempted by a recovery coordinator, we abort and let them complete (and notify us about the execution result)
+ *
+ * TODO (desired, testing): dedicated burn test to validate outcomes
+ */
+public class CoordinateSyncPoint extends CoordinatePreAccept<SyncPoint>
+{
+    private CoordinateSyncPoint(Node node, TxnId txnId, Txn txn, FullRoute<?> route)
+    {
+        super(node, txnId, txn, route);
+    }
+
+    public static AsyncResult<SyncPoint> exclusive(Node node, Seekables<?, ?> keysOrRanges)
+    {
+        return coordinate(ExclusiveSyncPoint, node, keysOrRanges);
+    }
+
+    public static AsyncResult<SyncPoint> inclusive(Node node, Seekables<?, ?> keysOrRanges)
+    {
+        return coordinate(Kind.SyncPoint, node, keysOrRanges);
+    }
+
+    private static AsyncResult<SyncPoint> coordinate(Kind kind, Node node, Seekables<?, ?> keysOrRanges)
+    {
+        TxnId txnId = node.nextTxnId(kind, keysOrRanges.domain());
+        FullRoute<?> route = node.computeRoute(txnId, keysOrRanges);
+        CoordinateSyncPoint coordinate = new CoordinateSyncPoint(node, txnId, node.agent().emptyTxn(kind, keysOrRanges), route);
+        coordinate.start();
+        return coordinate;
+    }
+
+    void onPreAccepted(List<PreAcceptOk> successes)
+    {
+        Deps deps = Deps.merge(successes, ok -> ok.deps);
+        Timestamp executeAt = foldl(successes, (ok, prev) -> mergeMax(ok.witnessedAt, prev), Timestamp.NONE);
+        if (executeAt.isRejected())
+        {
+            proposeAndCommitInvalidate(node, Ballot.ZERO, txnId, route.homeKey(), route, executeAt, this);
+        }
+        else
+        {
+            // SyncPoint transactions always propose their own txnId as their executeAt, as they are not really executed.
+            // They only create happens-after relationships wrt their dependencies, which represent all transactions
+            // that *may* execute before their txnId, so once these dependencies apply we can say that any action that
+            // awaits these dependencies applies after them. In the case of ExclusiveSyncPoint, we additionally guarantee
+            // that no lower TxnId can later apply.
+            new Propose<SyncPoint>(node, tracker.topologies(), Ballot.ZERO, txnId, txn, route, deps, txnId, this)
+            {
+                @Override
+                void onAccepted()
+                {
+                    node.send(tracker.nodes(), id -> new Apply(id, tracker.topologies(), tracker.topologies(), txnId.epoch(), txnId, route, txn, txnId, deps, txn.execute(txnId, null), txn.result(txnId, null)));
+                    accept(new SyncPoint(txnId, deps), null);
+                }
+            }.start();
+        }
+    }
+}
--- a/accord-core/src/main/java/accord/coordinate/Propose.java
+++ b/accord-core/src/main/java/accord/coordinate/Propose.java
@@ -24,26 +24,32 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.BiConsumer;
 
-import accord.api.Result;
 import accord.api.RoutingKey;
 import accord.coordinate.tracking.AbstractTracker.ShardOutcomes;
 import accord.coordinate.tracking.QuorumTracker;
 import accord.coordinate.tracking.QuorumTracker.QuorumShardTracker;
 import accord.coordinate.tracking.RequestStatus;
-import accord.messages.Callback;
-import accord.primitives.*;
-import accord.topology.Shard;
-import accord.topology.Topologies;
 import accord.local.Node;
 import accord.local.Node.Id;
 import accord.messages.Accept;
 import accord.messages.Accept.AcceptReply;
+import accord.messages.Callback;
+import accord.primitives.Ballot;
+import accord.primitives.Deps;
+import accord.primitives.FullRoute;
+import accord.primitives.Route;
+import accord.primitives.Timestamp;
+import accord.primitives.Txn;
+import accord.primitives.TxnId;
+import accord.topology.Shard;
+import accord.topology.Topologies;
 
 import static accord.coordinate.tracking.AbstractTracker.ShardOutcomes.Fail;
 import static accord.coordinate.tracking.RequestStatus.Failed;
+import static accord.messages.Commit.Invalidate.commitInvalidate;
 import static accord.utils.Invariants.debug;
 
-class Propose implements Callback<AcceptReply>
+abstract class Propose<R> implements Callback<AcceptReply>
 {
     final Node node;
     final Ballot ballot;
@@ -52,14 +58,14 @@ class Propose implements Callback<AcceptReply>
     final FullRoute<?> route;
     final Deps deps;
 
-    private final List<AcceptReply> acceptOks;
+    final List<AcceptReply> acceptOks;
     private final Map<Id, AcceptReply> debug = debug() ? new HashMap<>() : null;
-    private final Timestamp executeAt;
-    private final QuorumTracker acceptTracker;
-    private final BiConsumer<Result, Throwable> callback;
+    final Timestamp executeAt;
+    final QuorumTracker acceptTracker;
+    final BiConsumer<? super R, Throwable> callback;
     private boolean isDone;
 
-    Propose(Node node, Topologies topologies, Ballot ballot, TxnId txnId, Txn txn, FullRoute<?> route, Deps deps, Timestamp executeAt, BiConsumer<Result, Throwable> callback)
+    Propose(Node node, Topologies topologies, Ballot ballot, TxnId txnId, Txn txn, FullRoute<?> route, Deps deps, Timestamp executeAt, BiConsumer<? super R, Throwable> callback)
     {
         this.node = node;
         this.ballot = ballot;
@@ -73,18 +79,9 @@ class Propose implements Callback<AcceptReply>
         this.acceptTracker = new QuorumTracker(topologies);
     }
 
-    public static void propose(Node node, Ballot ballot, TxnId txnId, Txn txn, FullRoute<?> route,
-                               Timestamp executeAt, Deps deps, BiConsumer<Result, Throwable> callback)
+    void start()
     {
-        Topologies topologies = node.topology().withUnsyncedEpochs(route, txnId, executeAt);
-        propose(node, topologies, ballot, txnId, txn, route, executeAt, deps, callback);
-    }
-
-    public static void propose(Node node, Topologies topologies, Ballot ballot, TxnId txnId, Txn txn, FullRoute<?> route,
-                               Timestamp executeAt, Deps deps, BiConsumer<Result, Throwable> callback)
-    {
-        Propose propose = new Propose(node, topologies, ballot, txnId, txn, route, deps, executeAt, callback);
-        node.send(propose.acceptTracker.nodes(), to -> new Accept(to, topologies, ballot, txnId, route, executeAt, txn.keys(), deps), propose);
+        node.send(acceptTracker.nodes(), to -> new Accept(to, acceptTracker.topologies(), ballot, txnId, route, executeAt, txn.keys(), deps), this);
     }
 
     @Override
@@ -106,7 +103,10 @@ class Propose implements Callback<AcceptReply>
             case Success:
                 acceptOks.add(reply);
                 if (acceptTracker.recordSuccess(from) == RequestStatus.Success)
+                {
+                    isDone = true;
                     onAccepted();
+                }
         }
     }
 
@@ -127,12 +127,7 @@ class Propose implements Callback<AcceptReply>
         callback.accept(null, failure);
     }
 
-    private void onAccepted()
-    {
-        isDone = true;
-        Deps deps = Deps.merge(acceptOks, ok -> ok.deps);
-        Execute.execute(node, txnId, txn, route, executeAt, deps, callback);
-    }
+    abstract void onAccepted();
 
     // A special version for proposing the invalidation of a transaction; only needs to succeed on one shard
     static class Invalidate implements Callback<AcceptReply>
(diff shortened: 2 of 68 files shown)


accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -374,9 +528,9 @@ public void run()
             }
         }
 
-        class AsyncState extends State implements SafeCommandStore
+        class AsyncState extends InMemoryCommandStore.State implements SafeCommandStore
```
aweisberg: Is removing the import an improvement?

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -54,6 +63,36 @@ public int id()
...
 
+    protected void setRejectBefore(ReducingRangeMap<Timestamp> newRejectBefore)
+    {
+        this.rejectBefore = newRejectBefore;
+    }
+
+    Timestamp preaccept(TxnId txnId, Seekables<?, ?> keys, SafeCommandStore safeStore)
+    {
+        NodeTimeService time = safeStore.time();
+        boolean isExpired = agent().isExpired(txnId, safeStore.time().now());
```
aweisberg, line 104: Maybe rename `isExpired` to `shouldReject`?

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
```diff
@@ -0,0 +1,93 @@
...
+        CoordinateSyncPoint coordinate = new CoordinateSyncPoint(node, txnId, node.agent().emptyTxn(kind, keysOrRanges), route);
+        coordinate.start();
+        return coordinate;
+    }
+
+    void onPreAccepted(List<PreAcceptOk> successes)
+    {
+        Deps deps = Deps.merge(successes, ok -> ok.deps);
+        Timestamp executeAt = foldl(successes, (ok, prev) -> mergeMax(ok.witnessedAt, prev), Timestamp.NONE);
+        if (executeAt.isRejected())
```
aweisberg, line 80: Should we expire?

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
```diff
@@ -0,0 +1,93 @@
...
+            proposeAndCommitInvalidate(node, Ballot.ZERO, txnId, route.homeKey(), route, executeAt, this);
+        }
+        else
+        {
+            // SyncPoint transactions always propose their own txnId as their executeAt, as they are not really executed.
+            // They only create happens-after relationships wrt their dependencies, which represent all transactions
+            // that *may* execute before their txnId, so once these dependencies apply we can say that any action that
+            // awaits these dependencies applies after them. In the case of ExclusiveSyncPoint, we additionally guarantee
+            // that no lower TxnId can later apply.
+            new Propose<SyncPoint>(node, tracker.topologies(), Ballot.ZERO, txnId, txn, route, deps, txnId, this)
```
aweisberg, line 91: Anonymously subclassing `Propose` to override `onAccepted` is moderately weird, but sure. For `ProposeAndExecute` there is a class.

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
```diff
@@ -0,0 +1,93 @@
...
+            {
+                @Override
+                void onAccepted()
+                {
+                    node.send(tracker.nodes(), id -> new Apply(id, tracker.topologies(), tracker.topologies(), txnId.epoch(), txnId, route, txnId, deps, txn.execute(txnId, null), txn.result(txnId, null)));
```
aweisberg: We send the apply to avoid having this transaction block others, and we don't need a response because it's fine if this transaction needs recovery?

accord-core/src/main/java/accord/coordinate/Propose.java
```diff
@@ -74,18 +73,19 @@ class Propose implements Callback<AcceptReply>
         this.acceptTracker = new QuorumTracker(topologies);
     }
 
-    public static void propose(Node node, Ballot ballot, TxnId txnId, Txn txn, FullRoute<?> route,
-                               Timestamp executeAt, Deps deps, BiConsumer<Result, Throwable> callback)
+    public static void propose(Node node, Topologies topologies, Ballot ballot, TxnId txnId, Txn txn, FullRoute<?> route,
```
aweisberg: Unused now

accord-core/src/main/java/accord/coordinate/Propose.java
```diff
@@ -128,12 +131,7 @@ public void onCallbackFailure(Id from, Throwable failure)
...
-    private void onAccepted()
-    {
-        isDone = true;
-        Deps deps = Deps.merge(acceptOks, ok -> ok.deps);
-        Execute.execute(node, txnId, txn, route, executeAt, deps, callback);
-    }
+    abstract void onAccepted();
 
     // A special version for proposing the invalidation of a transaction; only needs to succeed on one shard
     static class Invalidate implements Callback<AcceptReply>
```
aweisberg, line 133: Should invalidate be pulled out into a separate class now?

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -212,56 +213,63 @@ private void recover()
...
-                            {
-                                // TODO (required, consider): when writes/result are partially replicated, need to confirm we have quorum of these
-                                Persist.persistAndCommit(node, txnId, route, txn, executeAt, deps, acceptOrCommit.writes, acceptOrCommit.result);
-                                accept(acceptOrCommit.result, null);
-                            }
-                        });
-                    });
+                {
+                    // must have gone through Accepted, so we must have witnessed >= Accepted for each shard
+                    Deps acceptedDeps = tryMergeAcceptedDeps();
```
aweisberg, line 223: We don't need to collect deps anymore? Does this mean we always enter `Recover` after reaching a quorum from each shard?

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -212,56 +213,63 @@ private void recover()
...
                     return;
+                }
 
                 case Accepted:
-                    // no need to preaccept the later round, as future operations always include every old epoch (until it is fully migrated)
-                    propose(acceptOrCommit.executeAt, mergeDeps());
-                    return;
+                {
+                    Deps deps;
+                    if (txnId.rw().proposesDeps()) deps = tryMergeAcceptedDeps();
```
aweisberg, line 252: Why do we have this distinction here for this transaction type? (needs comment)

Seems like this is the one case where we are ok with accepted deps being null, but for a regular transaction it's an invariant that it not be null?

So a `proposeDeps` transaction must arrive here with less information, but still want to proceed to invalidation? Why is that?

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -212,56 +213,63 @@ private void recover()
...
+                    else deps = mergeDeps();
+
+                    if (deps != null)
+                    {
+                        // TODO (desired, behaviour): if we didn't find Accepted in *every* shard, consider invalidating for consistency of behaviour
+                        propose(acceptOrCommit.executeAt, deps);
+                        return;
+                    }
+
+                    // if we propose deps, and cannot assemble a complete set, then we never reached consensus and can be invalidated
```
aweisberg, line 262: What does it mean to propose deps?

accord-core/src/main/java/accord/local/SafeCommandStore.java
```diff
@@ -66,7 +70,37 @@
         EXECUTES_AFTER
     }
     enum TestDep { WITH, WITHOUT, ANY_DEPS }
-    enum TestKind { Ws, RorWs }
+    enum TestKind implements Predicate<Txn.Kind>
```
aweisberg: So with this do sync points only conflict in one direction and basically not block other transactions? Nice.

accord-core/src/main/java/accord/messages/BeginRecovery.java
```diff
@@ -152,16 +164,18 @@ public RecoverReply reduce(RecoverReply r1, RecoverReply r2)
         if (!ok1.status.hasBeen(PreAccepted)) throw new IllegalStateException();
 
         PartialDeps deps = ok1.deps.with(ok2.deps);
+        PartialDeps acceptedDeps = ok1.deps == ok1.acceptedDeps && ok2.deps == ok2.acceptedDeps ? deps : ok1.acceptedDeps.with(ok2.acceptedDeps);
```
aweisberg, line 169: If one or more of them is `PartialDeps.NONE` why are we doing `with`? To make sure we pick whichever one might not be `NONE` or get double `NONE` merged?

accord-core/src/main/java/accord/messages/BeginRecovery.java
```diff
@@ -216,19 +230,21 @@ public static class RecoverOk extends RecoverReply
         public final Ballot accepted;
         public final Timestamp executeAt;
         public final PartialDeps deps;
+        public final PartialDeps acceptedDeps; // only those deps that have previously been proposed
```
aweisberg, line 245: Good place to comment on when/why the distinction matters, but it if it can/should be covered where this is used that is also fine

accord-core/src/main/java/accord/messages/BeginRecovery.java
```diff
@@ -297,7 +313,7 @@ private static Deps acceptedStartedBeforeWithoutWitnessing(SafeCommandStore comm
     {
         try (Deps.Builder builder = Deps.builder())
         {
-            commandStore.mapReduce(keys, ranges, RorWs, STARTED_BEFORE, startedBefore, WITHOUT, startedBefore, Accepted, PreCommitted,
+            commandStore.mapReduce(keys, ranges, Any, STARTED_BEFORE, startedBefore, WITHOUT, startedBefore, Accepted, PreCommitted,
```
aweisberg: Why use `Any` for recovery? What is the interaction you are looking for with sync transactions?

Does recovery force them to order with sync transactions, but during regular execution we don't?

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -19,12 +19,16 @@
...
 import accord.local.Node.Id;
-import accord.utils.Invariants;
+
+import static accord.utils.Invariants.checkArgument;
 
 import javax.annotation.Nonnull;
 
 public class Timestamp implements Comparable<Timestamp>
 {
+    private static final int REJECTED_FLAG = 0x8000;
```
aweisberg, line 30: Should `REJECTED_FLAG` be the same as `MERGE_FLAGS`?

accord-core/src/main/java/accord/primitives/Txn.java
```diff
@@ -53,6 +66,16 @@ public boolean isRead()
             return this == Read;
         }
 
+        /**
+         * Note that it is only possible to do this for transactions whose execution time is not dependent on
+         * others, i.e. where we may safely propose executeAt = txnId regardless of when it is witnessed by
```
aweisberg, line 91: How does a transaction end up being that way? What transaction could safely propose a txnId of 0 at any time?

Also what is "this"? :-)

It's proposing invalidations, don't those need to properly order with what already occurred?

accord-core/src/main/java/accord/primitives/Txn.java
```diff
@@ -32,13 +32,26 @@
...
     {
-        Read, Write;
+        Read,
+        Write,
+
+        /**
+         * A pseudo-transaction whose deps represent the complete set of transactions that may execute before it,
+         * without interfering with their execution.
+         *
+         * A SyncPoint is unique in that it agrees not only an executeAt but also a precise collection of dependencies,
```
aweisberg: Precise collection needs elaboration, what you mean is that after `Preaccept` we have a precise set because we are fine with extra dependencies?

Don't all transactions determine a precise set of dependencies in order to determine their `executeAt`?

accord-core/src/main/java/accord/primitives/Txn.java
```diff
@@ -32,13 +32,26 @@
...
+         * so that is effect on invalidation of earlier transactions is durable. This is most useful for ExclusiveSyncPoint.
```
aweisberg: This last half really confuses me, `SyncPoint` shouldn't invalidate?

aweisberg (changes requested):
Bunch of questions and nits

Still reviewing `ReducingIntervalMap` and `ReducingRangeMap`

accord-core/src/main/java/accord/utils/ReducingRangeMap.java
```diff
@@ -0,0 +1,131 @@
...
+
+            j = keys.findNext(j + 1, endKeys.get(i), FAST);
+            if (j < 0) j = -1 - j;
+            else if (inclusiveEnds) ++j;
+        }
+        return accumulator;
+    }
+
+    /**
+     * returns a copy of this LinearRangeMap limited to the ranges supplied, with all other ranges reporting Ballot.none()
```
aweisberg: It's not a ballot anymore

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: We use it on both replica and coordinator, and on coordinator that perhaps isn't quite right. Though we could have different criteria for each, or we could stop checking it on coordinator and just check if any of the votes were to reject it,

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
belliottsmith: I think this is just Java 8's compilation woes, sometimes we have to rejig imports until it decides things are fine again. Just noise.

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
belliottsmith: Do you mean should we check the `agent.isExpired` thing? Since this has no user visible impact, and we presumably want it for some process reason it didn't seem valuable.

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
belliottsmith: Happy to introduce an inner class. I did previously have a static method I could pass a callback to that would create an anonymous class to wrap the callback but that felt wrong.

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
belliottsmith: Basically, though in practice this is a stub for whatever we may want to do with our implementations that use this. Or, for some distributed cleanup for those nodes that don't need it - this is probably the one part of this patch that isn't "done" (but is close enough that we could probably commit as things are so we have a shared base to operate on)

accord-core/src/main/java/accord/coordinate/Propose.java
belliottsmith: I don't think so, but happy to be told otherwise. We have a pattern of ActionX and ActionX.Invalidate - though perhaps this isn't a good pattern anyway 🤷

accord-core/src/main/java/accord/coordinate/Recover.java
belliottsmith: Yes, we require a quorum from each side for recovery. There were TODOs about using the deps we have where possible, and for this work it made sense to just do that since we need it for the exclusive sync point - which *proposes* deps, i.e. expects *precisely* those deps to be re-proposed should it fail to durably commit/apply/whathaveyou.

accord-core/src/main/java/accord/coordinate/Recover.java
belliottsmith: So, this is something that is probably going to evolve. In practice we *don't* need to treat them differently today, because we do actually send deps with accept today, *and* because we are able to invalidate any transaction that hadn't reached a quorum of accepts (though we don't currently, we probably *should*, as it is consistent with our behaviour elsewhere of invalidating transactions we know were not executed by their coordinator). 

However (and this is documented, but evidently not well), standard transactions are happy to get deps that are superset of those they need, and then filter them. However, exclusive sync points want to _precisely_ exclude txnId they did not witness. We don't want extra txnId sneaking in next time around, as that wouldn't have been witnessed by the replicas that wanted to know for sure they had a complete set of txnId lower than that sync point. So for these transactions we _require_ that we re-propose precisely the same deps as the original coordinator _or_ we invalidate the transaction if they cannot be found.

belliottsmith: I will of course try to express this better than I have, and in more places.

accord-core/src/main/java/accord/coordinate/Recover.java
belliottsmith: Send `Accept` with a specific set of `Deps` we want to be durably recoverable before we use them. So for these transactions we're really agreeing `Deps` rather than `executeAt` (since they always use `txnId` for that, though this is arbitrary)

belliottsmith: I think this will become more confusing when we permit regular transactions to _not_ send deps, but it is already the case that we do not treat the `deps` register on replicas as durable/consistent for regular transactions. I will think about where best to document this - I will at least mention it in `Command`, and perhaps also in `RecoverOk` and `Accept`

accord-core/src/main/java/accord/local/SafeCommandStore.java
belliottsmith: Yep, that's the idea

accord-core/src/main/java/accord/messages/BeginRecovery.java
belliottsmith: If either are NONE then the operation bails out cheaply. We don't know if either of them are NONE though, they might both be full deps. Perhaps I'm misunderstanding. All I'm doing with the ternary is avoiding calling `with` twice if I can reuse the prior result.

accord-core/src/main/java/accord/messages/BeginRecovery.java
belliottsmith: Yep, agreed

accord-core/src/main/java/accord/messages/BeginRecovery.java
belliottsmith: These collections are used for _invalidation_, not for deps or execution. We're looking here for the transactions that _invalidate_ us by _not witnessing us_. For sync points (most importantly exclusive sync points) this is the whole purpose of durably agreeing Deps - so that we know for sure we have seen all `TxnId` that can _possibly_ be agreed to execute.

accord-core/src/main/java/accord/primitives/Timestamp.java
belliottsmith: I will comment. The idea here is that we probably don't want all flags to be mergeable, though today we only have one flag, and we do want it to propagate.

accord-core/src/main/java/accord/primitives/Txn.java
belliottsmith: It's only possible to "propose deps". I will clean up this comment, to contain similar text to the other proposed places to explain this distinction. In this case we're essentially saying that, because this transaction doesn't actually execute, it doesn't need to agree an `executeAt`, permitting it the option of instead agreeing some `deps` from the single `PreAccept` round (since the deps after `PreAccept` are valid for `executeAt=txnId`, but not for `executeAt>txnId`)

accord-core/src/main/java/accord/primitives/Txn.java
belliottsmith: Nope, most transactions do not _agree_ their deps, and indeed for those that agree with 1RT consensus there are no durable deps to recover (and we intend to stop supplying them for accept too). But for SyncPoint, for the reasons stated elsewhere, this is fine because we always have `executeAt=txnId` and this is fine because we don't "execute" them.

accord-core/src/main/java/accord/primitives/Txn.java
belliottsmith: An `ExclusiveSyncPoint` invalidates all `TxnId` lower than it that aren't witnessed by it.

aweisberg: Ah got it, mixed up that `SyncPointer` includes both.

accord-core/src/main/java/accord/primitives/Txn.java
aweisberg: I still don't follow something basic. What does agree mean?

Maybe we need to step up a level and explain how a sync point achieves its objective where when a result is returned all the dependencies of this txn have reached applied.

Maybe I am crazy, but it looks like the callback fires after `Propose` completes, and doesn't wait for the dependencies to actually apply?

I am going to reread again today and see if I can get to a better understanding.

aweisberg: The reason the dependencies need to be applied is that the use case is needing the side effects of the dependencies to already have occurred for migration.

accord-core/src/main/java/accord/messages/BeginRecovery.java
aweisberg: Ok, it took me a minute. What this is doing is creating the accepted deps only if it has to because it's not the same as `deps` when one or more has returned `NONE` for accepted deps.

accord-core/src/main/java/accord/primitives/Txn.java
belliottsmith: Agree means we have a quorum of accepted proposals. Ie we have reached a durable decision. I think this is quite normal terminology? At least that’s how I talk about consensus

accord-core/src/main/java/accord/coordinate/Propose.java
aweisberg: FYI still unused

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
aweisberg: I don't really care. It's not hard to understand I just noted we have 3 ways to get at different types of proposals (Invalidate, ProposeAndExecute, and this).

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
aweisberg: I mentioned this earlier, but for inclusive sync point I think we need to know when it is applied (because apply waits for dependencies right?) since we want the side effects of the dependencies.

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
aweisberg: Looks like we expire requests so if the transaction is stuck it will eventually generate an error so yeah seems fine to not align with a specific timeout just lack of progress.

I do wonder if it could get wedged in say the transaction logic, but then a retry wouldn't help either.

accord-core/src/main/java/accord/local/CommandStore.java
aweisberg: I am just reacting to `isExpired` being reused for the result of `rejectBefore`
If you want to call both of them expiration that's pretty reasonable.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
aweisberg: Yikes!

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
belliottsmith: I guess we could put `ProposeAndExecute` as an inner class. I don't think of this as three different kinds of Propose though, but as two different behaviours conducted after `Propose`, and one special kind of propose (invalidation), as this is behavioural different at the replica, rather than in action following. 

We could chain a layer of indirection with callbacks, but it seemed clearer to me this way. I don't have a super strong feeling, but would prefer to retain the symmetry of `Invalidate` with other classes

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
belliottsmith: Yep, my expectation is that the coordinator would register a listener for the application of the transaction locally. The apply at remote replicas is just a matter of cleaning up really. We do not need any remote replica to apply in order to achieve our goal, so we can return immediately upon reaching consensus and use our dependencies for our purposes.

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: Oh, I see. I thought you meant the method, but you mean the local variable. Makes perfect sense.

accord-core/src/main/java/accord/primitives/Txn.java
aweisberg: I am figuring through how regular transactions don't agree on their deps, so with 1RT yeah that makes sense.

There are no changes to how `Accept` handles deps so I was trying to understand what it means for these transactions to agree on deps, but regular transactions don't. So one different is we always run accept, and I guess in the future you mean that accept won't persist dependencies for regular transactions (recovery will go find them again?), but we will continue to persist them for barriers so when they recover the precise set of dependencies is returned.

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
aweisberg: OK, there are two versions of this though. One is locally listening (key migration), but in the other we actually want to know it was applied at a quorum remotely (range migration)?

Or can we locally listen for the state change to globally durable?

I wonder if we will have trouble with lost notifications here. Like we may not be involved in the dependency transactions (range we don't replicate) at all so why would we receive notification of it?

accord-core/src/main/java/accord/primitives/Txn.java
belliottsmith: Yes, but it’s slightly more than that: since an Accept round for a normal transaction means it has (likely) taken a later executeAt the dependencies aren’t enough for execution as they don’t include the full set of deps for the higher executeAt - for that we use the dependencies returned in the Acceot response

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
belliottsmith: True, for the global range migration version we might want to tweak this to support waiting for a certain level of persistence at all replicas. Is it better to add that at point of use, or would you like to add it here in this patch?

aweisberg: I can do it later. I think I will also need to add the local metadata only behavior of the sync point where we already know it has been satisfied.

belliottsmith:
Ok, have addressed the last nits. Merge on hold until immutable state lands.

belliottsmith:
I've rebased and back ported some relevant changes from the bootstrap PR (plus some changes not super tied to either). I've also brought in some changes to listener handling. I plan to merge this to cep-15-accord alongside bootstrap, to simplify things, but would be good to get this review signed off once you're ready.

PR #27, opened 2023-01-18
Support asynchronous CommandStore message processing

belliottsmith:
While messages are submitted to a CommandStore asynchronously, the processing of a message is then expected to be synchronous. However, it would be beneficial particularly for large range transactions, but also for recovery, to be able to defer some work that might involve IO to after the command store processing completes, e.g. for calculating dependencies.

### DIFF
--- a/accord-core/src/main/java/accord/impl/CommandsForKey.java
+++ b/accord-core/src/main/java/accord/impl/CommandsForKey.java
@@ -20,7 +20,7 @@ package accord.impl;
 
 import accord.api.Key;
 import accord.local.*;
-import accord.local.SafeCommandStore.CommandFunction;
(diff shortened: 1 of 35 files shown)


belliottsmith:
ping

aweisberg (changes requested):
Can you rename zSlowSearcher` and `SlowSearchFunction` to something like `IncrementalSearcher` or whatever seems more descriptive and add a comment that the motivation is to allow implementations to process results without caching all of them or blocking for a long time or to avoid bringing them into the cache?

Otherwise yeah I am basically +1

belliottsmith:
Oh, somehow I missed you +1'd this. I'll rebase this perhaps onto bootstrap.

PR #28, opened 2023-01-26
CASSANDRA-18203: CEP-15: (C*) Improve Burn Tests to include thread scheduling

dcapwell:


accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -576,15 +576,25 @@ public <T> Future<T> submit(PreLoadContext context, Function<? super SafeCommand
...
-            executor = Executors.newSingleThreadExecutor(r -> {
+            this(id, time, agent, store, progressLogFactory, rangesForEpoch, Executors.newSingleThreadExecutor(r -> {
                 Thread thread = new Thread(r);
                 thread.setName(CommandStore.class.getSimpleName() + '[' + time.id() + ']');
                 return thread;
-            });
+            }));
+        }
+
+        private SingleThread(int id, NodeTimeService time, Agent agent, DataStore store, ProgressLog.Factory progressLogFactory, RangesForEpochHolder rangesForEpoch, ExecutorService executor)
```
dcapwell: name may not make sense in this case, 100% cool making a base type such as `ExecutorCommandStore` and have `SingleThread` extend it and use the existing logic to build the single threaded executor

accord-core/src/test/java/accord/impl/basic/DelayedExecutorService.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private Task<?> newTaskFor(Runnable command) {
+        return command instanceof Task ? (Task<?>) command : newTaskFor(command, null);
+    }
+
+    @Override
+    public void execute(Runnable command) {
+        Task<?> task = newTaskFor(command);
+        int jitterMillis = THREAD_SCHEDULING_OVERHEAD_MILLIS + random.nextInt(1000);
```
belliottsmith: On average 505ms of latency for scheduling seems a bit steep. Might be nice to copy some of the simulator's behaviour here if we could, so that there's some patterns. Probably it should average tens to hundreds of micros, though, with some larger periods.

dcapwell: ill try to read the simulator scheduler and come back to this comment.

dcapwell: What I see is `org.apache.cassandra.simulator.Action#schedule` which adds the following jitter

```
@Override
    public long schedulerDelayNanos()
    {
        return (scheduler.delayChance.get(random) ? scheduler.longDelayNanos : scheduler.delayNanos).select(random);
    }

```

with these config values

```
// long delay chance
schedulerDelayChance = new ChanceRange(randomSource -> randomSource.qlog2uniformFloat(4), 0.01f, 0.1f),
schedulerDelayNanos = new LongRange(0, 50, MICROSECONDS, NANOSECONDS),
schedulerLongDelayNanos = new LongRange(50, 5000, MICROSECONDS, NANOSECONDS),
KindOfSequence kind = UNIFORM or UNIFORM_STEP or RANDOMWALK
```

So what I see is we randomly select the probability for the run between 1% and 10% and use that as a constant as the input into the `Decision` to select if we use "large" or not.

Large is 50 micro to 5ms.


I can def add similar logic to BurnTest, I didn't see an existing bias select function so went lazy with uniform; would be good to so the "large" range is rare.

Now, about the numbers I went with, these were pulled from a prod cluster for a different application where CFS avg 5ms to start a thread that had its condition met and max of 1s... these numbers were thread starting, but we are not, so the overhead really is just how long it takes for the thread to see the state...  So... lets test that!

On M1 Mac this is what I saw (in micro)

```
Summary of 10000000 Samples
Min: 6
Median: 4102
Avg: 4241
Max: 25729
Counts in Millis
0	910225	9.10%
1	982552	9.83%
2	977544	9.78%
3	1850840	18.51%
4	2811417	28.11%
5	1179405	11.79%
6	369632	3.70%
7	252550	2.53%
8	183516	1.84%
9	154046	1.54%
10	66330	0.66%
11	43829	0.44%
12	41320	0.41%
13	23590	0.24%
14	16804	0.17%
15	7367	0.07%
16	10912	0.11%
17	14986	0.15%
18	20019	0.20%
19	11307	0.11%
20	10720	0.11%
21	14148	0.14%
22	11867	0.12%
23	11820	0.12%
24	12201	0.12%
25	11053	0.11%
```

So yeah, I can work on similar logic, but I might tweak the ranges

dcapwell: if I avoid queue time (as another task on the queue would be running in that case) I get

```
Summary of 40000000 Samples
Min: 0
Median: 4
Avg: 4
Max: 2006
Counts in Millis
0	39999880	100.00%
1	80	0.00%
2	40	0.00%
```

dcapwell: pushed change

```
this.jitterInNano = new SegmentedIntRange(
                new IntRange(random, microToNanos(0), microToNanos(50)),
                new IntRange(random, microToNanos(50), msToNanos(5)),
                // this is different from Apache Cassandra Simulator as this is computed differently for each executor
                // rather than being a global config
                new Decision.FixedChance(random, new IntRange(random, 1, 10).getInt() / 100f)
        );
```

I tried zipf to bias to higher values in the range, but math3's version added a lot more cost, so went with simple uniform.

accord-core/src/test/java/accord/burn/random/RandomInt.java
```diff
@@ -0,0 +1,6 @@
+package accord.burn.random;
+
+public interface RandomInt
```
dcapwell: C* simulator passes random in, where as BurnTest defines it on init... Personally I am a fan of the C* simulator's version but felt best to match existing code

belliottsmith: I think ideally we would move towards as similar a style as possible, and share as much code as possible between the two. But no need to undertake that now if you don't want.

accord-core/src/test/java/accord/impl/basic/DelayedExecutorService.java
belliottsmith: I think it would be _preferable_ to use the same approach in both for randomising, as far as possible, so we can trend towards sharing codebases as much as possible.

accord-core/src/test/java/accord/burn/random/RandomInt.java
dcapwell: I don't actually mind the interfaces working that way, that doesn't force us to migrate 100% but does let us build in that direction.

Ill try it out, I think it only impacts 1 real class and that can hold `Random` for now

dcapwell: pushed

accord-core/src/test/java/accord/burn/random/SegmentedIntRange.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class SegmentedIntRange implements RandomInt
+{
+    private final RandomInt small, large;
+    private final Decision largeDecision;
+
+    public SegmentedIntRange(RandomInt small, RandomInt large, Decision largeDecision) {
+        this.small = Objects.requireNonNull(small);
+        this.large = Objects.requireNonNull(large);
+        this.largeDecision = Objects.requireNonNull(largeDecision);
```
belliottsmith: chooseLargeChance?

accord-core/src/test/java/accord/burn/random/SegmentedIntRange.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.Objects;
+import java.util.Random;
+
+public class SegmentedIntRange implements RandomInt
+{
+    private final RandomInt small, large;
+    private final Decision largeDecision;
+
+    public SegmentedIntRange(RandomInt small, RandomInt large, Decision largeDecision) {
```
belliottsmith: braces/newlines

accord-core/src/test/java/accord/burn/random/IntRange.java
```diff
@@ -0,0 +1,21 @@
+package accord.burn.random;
+
+import java.util.Random;
+
+public class IntRange implements RandomInt
+{
+    private final int min, upperBound;
```
belliottsmith: upperBound is confusing. Perhaps `maxDelta`, or `range`?

accord-core/src/test/java/accord/impl/basic/DelayedExecutorService.java
```diff
@@ -0,0 +1,107 @@
...
+    private final RandomInt jitterInNano;
+
+    public DelayedExecutorService(PendingQueue pending, Random random)
+    {
+        this.pending = pending;
+        this.random = random;
+        this.jitterInNano = new SegmentedIntRange(
+                new IntRange(microToNanos(0), microToNanos(50)),
+                new IntRange(microToNanos(50), msToNanos(5)),
+                // this is different from Apache Cassandra Simulator as this is computed differently for each executor
```
belliottsmith: Ideally it would probably be per-node for both, but not super important. I thought the C* simulator also used the random walk behaviour for jitter, but it looks like it doesn't. It would be good if both did, to simulate nodes becoming unhealthy for a period, then restoring behaviour. But not essential for this patch.

accord-core/src/test/java/accord/impl/basic/DelayedExecutorService.java
```diff
@@ -0,0 +1,107 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.AbstractExecutorService;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+public class DelayedExecutorService extends AbstractExecutorService
```
belliottsmith: Perhaps make clearer this isn't a real executor service? Perhaps `SimulatedExecutorService` or some other better name.

accord-core/src/test/java/accord/burn/random/SegmentedIntRange.java
dcapwell: I should try to figure out how to fix IntelliJ to match C*... I never pay attention when IntelliJ auto creates =(

accord-core/src/test/java/accord/burn/random/IntRange.java
dcapwell: I prefer `maxDelta` so went with that

accord-core/src/test/java/accord/impl/basic/DelayedExecutorService.java
dcapwell: > Ideally it would probably be per-node for both

makes sense, nodes coordinating to act badly at the same time isn't realistic.

> I thought the C* simulator also used the random walk behaviour for jitter, but it looks like it doesn't. It would be good if both did, to simulate nodes becoming unhealthy for a period

At least in this context I see `RandomWalkDecision` and tested and don't see such a behavior... Changed test to `RandomWalkPeriod` and see do see some logical temporal behavior but don't 100% grasp why yet; below are the results in my testing

```
# RandomWalkDecision for 10k samples
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 3 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 3 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 4 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
True seen 2 times in a row
```

```
# RandomWalkPeriod for 10k samples
Seq 82
Seq 136
Seq 18
Seq 4
Seq 2
Seq 7
Seq 3
Seq 8
Seq 3
Seq 10
Seq 34
Seq 5
Seq 15
Seq 214
Seq 13
Seq 2
Seq 6
Seq 216
Seq 243
```

The result are sequences seen where the result is > the mid point of the range, so with `new LongRange(100, 2000, MICROSECONDS, NANOSECONDS)` those are sequences seen greater than 1ms.

I do agree this type of behavior is desirable, so will take a look at this logic and see if I can get a working `Decision` that has a temporal behavior of returning `true`; ill time box it to a 1-3h though

accord-core/src/test/java/accord/impl/basic/DelayedExecutorService.java
dcapwell: used `SimulatedDelayedExecutorService` to denote both semantics; simulated + delayed

accord-core/src/test/java/accord/impl/basic/DelayedExecutorService.java
dcapwell: I created a `RandomWalkPeriodChance` that looks to be working in this manner, but the true happens 58% of the time where as we want around 1-10%... so trying to mess with this to get that involved

dcapwell: In better understanding `RandomWalkDecision`, the reason for the period being so low is that we are asking for between 1-10% chance, so we can't have long periods in order to respect that... for that range here is what I see

```
Number of trues: 4110(4.109999999999999%)
Number of false: 95890(95.89%)
True seq count 0 happened 91999 times
True seq count 1 happened 3689 times
True seq count 2 happened 187 times
True seq count 3 happened 14 times
True seq count 4 happened 1 times
```

switching between 10-20% I see

```
Number of trues: 15340(15.340000000000002%)
Number of false: 84660(84.66%)
True seq count 0 happened 71748 times
True seq count 1 happened 10884 times
True seq count 2 happened 1690 times
True seq count 3 happened 285 times
True seq count 4 happened 44 times
True seq count 5 happened 9 times
```

Using `FixedChance` I see

```
Number of trues: 4047(4.047%)
Number of false: 95953(95.953%)
True seq count 0 happened 92059 times
True seq count 1 happened 3750 times
True seq count 2 happened 135 times
True seq count 3 happened 9 times
```

dcapwell: Given this, there isn't a real change between `RandomWalkDecision` and `FixedChance` at this frequency, so will exclude from this patch.

belliottsmith: Could I take a look at your RandomWalkPeriodChance? What I would probably do is have some chance that the RandomWalkPeriod can exceed its normal min/max, and then the min/max are flipped to the expanded range until they return to their basic range. Or something like this. But, also, no need to agonise over this.

dcapwell: its in my git history, revert the last commit and you have a `main` method that I was using for my testing; edb8909f1fd8ff594f39ab1767e696d04a7d8d61 removes

belliottsmith: Here's my attempt: https://github.com/belliottsmith/cassandra-accord/tree/18203-suggest

This looks to produce some interesting patterns by switching between large and small ranges with some chance, and the length of a large/small period is used to determine how likely we are to switch. It appears to produce a correct distribution over a long enough sample period while still giving very lengthy periods of high jitter.

dcapwell: thanks, ill take a closer look.  at attempts I see it ranges between 5% and 13%, with the avg being around 7-8%.

saw some rather crazy cases like

```
103	3
	105	2
	111	2
	118	2
	144	2
	181	2
	208	2
	211	2
	295	2
	484	2
```

so many sequences of bad back and forth; in this case 484 long times!

Ill take a closer look Monday, but with these results could drop `SegmentedIntRange` in favor of `SegmentedRandomWalkPeriod` and drop `Decision` as well.

belliottsmith: If you run long enough it averages to around 6.2% for me which is 1/16 as requested. But you can get some very long runs of bad luck!

We should probably randomise the ratio for each test, or randomise the range of ratios per test and the specific ratio per host

dcapwell: I have been working on unit tests and finding this class is more complex than expected... I had to add a concept of "resample" so we keep adding more attempts to the loop when the target ratio isn't met... I start with 100k samples and in cases like the following, I need `21473837` resamples (`21473837 * 100 = 2,147,383,700` attempts needed).

```
int maxSmall = 78782771;
int maxLarge = 79722613;
int smallRatio = 14;
SegmentedRandomWalkPeriod period = new SegmentedRandomWalkPeriod(new IntRange(0, maxSmall), new IntRange(0, maxLarge), smallRatio, rs);
```

It seems that when it gets biased towards large it takes too many samples before getting back inline...  It also requires that `large` overlaps with `small` (though I think this is fixable), which makes this class less desirable I think; I personally feel that the existing `SegmentedIntRange` is better as it stays within the expected range.

If you want to take a look at the tests, they are in `accord.burn.random.SegmentedRandomRangeTest`.  `disjoint` tests `SegmentedIntRange` and `overlap` tests `Decision.SegmentedRandomWalkPeriod`.

belliottsmith: > It also requires that large overlaps with small (though I think this is fixable)

Why is this a problem?

> It seems that when it gets biased towards large it takes too many samples before getting back inline... 

I'm not convinced this is a major issue - we don't need to precisely hit the ratio every run, or even at all within a run (or even get close, we don't stipulate the variance, and the purpose here is to produce chaos). But, if you are concerned we could very simply bound the number of steps we permit to stay in the large side so we can call it a day.

I think we are probably overthinking this now, though, and should pick something and move to greener pastures. It is fine either way.

dcapwell: Created a new `FrequentLargeRange` which has a simple way to generate "runs" and makes sure it stays within the expected range.  This gives us the benefit that we can have large runs, that we can have disjoint small/large ranges, and that the ratio is respected in real time.

belliottsmith: Like I said, the range isn’t something we should be strict about. We’re just generating config to generate random values, and there’s no such thing as “staying within a range” for frequency of random values - that only makes sense over some time horizon and distribution, neither of which we configure.

a random walk is independently nice because it isn’t a uniform distribution but represents shifting patterns of behaviour, not just at the small/large boundary but within each scheme.

however, we’ve agonised over this for many times longer than we should have so 🤷

belliottsmith: (Your class could easily use RandomWalkPeriod for small and large if you want though - but really, I’ll stop suggesting things now and we’ll take whatever you’re happy with )

accord-core/src/test/java/accord/burn/TopologyUpdates.java
```diff
@@ -185,46 +186,47 @@ private static Collection<Node.Id> allNodesFor(Txn txn, Topology... topologies)
...
     {
         Map<TxnId, CheckStatusOk> syncMessages = new ConcurrentHashMap<>();
         Consumer<Command> commandConsumer = command -> syncMessages.merge(command.txnId(), new CheckStatusOk(node, command), CheckStatusOk::merge);
+        AsyncChain<Void> start;
         if (committedOnly)
-            getUninterruptibly(node.commandStores().forEach(commandStore -> InMemoryCommandStore.inMemory(commandStore).forCommittedInEpoch(ranges, srcEpoch, commandConsumer)));
+            start = node.commandStores().forEach(commandStore -> InMemoryCommandStore.inMemory(commandStore).forCommittedInEpoch(ranges, srcEpoch, commandConsumer));
         else
-            getUninterruptibly(node.commandStores().forEach(commandStore -> InMemoryCommandStore.inMemory(commandStore).forEpochCommands(ranges, srcEpoch, commandConsumer)));
+            start = node.commandStores().forEach(commandStore -> InMemoryCommandStore.inMemory(commandStore).forEpochCommands(ranges, srcEpoch, commandConsumer));
```
dcapwell, line 197: this caused a dead lock once the Executor was made the delayed one, so made it async

accord-core/src/test/java/accord/burn/TopologyUpdates.java
```diff
@@ -185,46 +186,47 @@ private static Collection<Node.Id> allNodesFor(Txn txn, Topology... topologies)
...
+        }));
     }
 
     private static final boolean PREACCEPTED = false;
     private static final boolean COMMITTED_ONLY = true;
 
-    /**
-     * Syncs all replicated commands. Overkill, but useful for confirming issues in optimizedSync
-     */
-    private static Stream<MessageTask> thoroughSync(Node node, long syncEpoch) throws ExecutionException
```
dcapwell, line 209: this was dead code, glad to remove or make async, so left it commented out just for now

accord-core/src/test/java/accord/impl/list/ListAgent.java
```diff
@@ -58,6 +58,7 @@ public void onInconsistentTimestamp(Command command, Timestamp prev, Timestamp n
     @Override
     public void onUncaughtException(Throwable t)
     {
+        if (t instanceof accord.coordinate.Timeout) return;
```
dcapwell, line 61: I don't like this but SPL sees it and now calls agent so had to do else we fail nearly every time... I feel like this is calling out that SPL should have different logic which would allow us to remove this

accord-core/src/main/java/accord/impl/SimpleProgressLog.java
```diff
@@ -500,6 +500,7 @@ void run(Command command)
                     Unseekables<?, ?> someKeys = unseekables(command);
 
                     BiConsumer<Known, Throwable> callback = (success, fail) -> {
+                        if (fail != null) node.agent().onUncaughtException(fail);
```
dcapwell, line 503: I have a lot of mixed feelings about this... right now we change our state to `Expected` and no-op if there is a failure, but since we do not handle I add this.  We see very frequent timeouts which caused BurnTest to fail 90% of the time, so I added a filter to ignore Timeout... which feels off...

The main reason I added this is that non-timeout exceptions were happening and we were ignoring them, so we were finding issues and by not calling `onUncaughtException` we wouldn't notice them.

dcapwell:
@belliottsmith rebased onto the change of Random to RandomSource

belliottsmith (approved):

PR #29, opened 2023-01-26
Cassandra 18204

dcapwell:


accord-core/build.gradle
```diff
@@ -56,16 +56,39 @@ task testJar(type: Jar, dependsOn: testClasses) {
...
+      // This is a hack to remove accord from the dependency list as this breaks Apache Cassandra
+      // This is a bug with Gradle that is fixed in 8.0 but not yet released
+      // see https://github.com/gradle/gradle/commit/6e596c638dff59dfa06032a7e7a3974275ddb3e1
+      pom.withXml {
+        asNode().dependencies.'*'.findAll() { it.artifactId.text() == accord_artifactId }.each() { it.parent().remove(it) }
       }
       from components.java
-    }
-
-    test(MavenPublication) {
```
dcapwell, line 66: this caused a conflict with `artifactId`, the actual jar wouldn't conflict but the `pom` would, so this always overrode  the pom of the previous task.  I didn't notice this till now due to all the dependencies/version already being present in C* build... cleaned this up so we just publish a new jar without overriding the `pom`

accord-core/build.gradle
```diff
@@ -56,16 +56,39 @@ task testJar(type: Jar, dependsOn: testClasses) {
...
+        scm {
+          connection = 'scm:https://gitbox.apache.org/repos/asf/cassandra-accord.git'
+          developerConnection = 'scm:https://gitbox.apache.org/repos/asf/cassandra-accord.git'
+          url = 'https://gitbox.apache.org/repos/asf?p=cassandra-accord.git;a=tree'
+        }
+      }
+      // This is a hack to remove accord from the dependency list as this breaks Apache Cassandra
+      // This is a bug with Gradle that is fixed in 8.0 but not yet released
+      // see https://github.com/gradle/gradle/commit/6e596c638dff59dfa06032a7e7a3974275ddb3e1
+      pom.withXml {
```
dcapwell, line 80: as note states, once Gradle 8 comes out and we switch, we can drop this logic (assuming the linked bug fix works)

accord-core/build.gradle
```diff
@@ -56,16 +56,39 @@ task testJar(type: Jar, dependsOn: testClasses) {
...
-    test(MavenPublication) {
-      artifactId = 'accord'
+      // don't need to add javadoc or sources as they were added already in the java block
       artifact testJar
+
+      versionMapping {
+        allVariants {
+          fromResolutionOf('testRuntimeClasspath')
+        }
+      }
```
dcapwell: I am testing how to add test dependencies... do we really need to is debatable (see it mostly as documentation given the fact we publish a -tests.jar)...  This doesn't actually work the way I want (remove and the same pom will be published); asked in SO to see if anyone can inform how https://stackoverflow.com/questions/75326910/how-to-get-test-dependencies-to-show-up-in-maven-pom-when-they-are-added-via-a-p

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -16,6 +16,6 @@
 
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-bin.zip
```
dcapwell: not related, just figured bumping to latest was prob a good idea...

buildSrc/src/main/groovy/accord.java-conventions.gradle
```diff
@@ -45,3 +45,17 @@ dependencies {
...
+}
+task copyTestDependencies(type: Copy) {
+    from configurations.testCompileClasspath - configurations.compileClasspath
+    into "${buildDir}/dependencies/test/libs"
+}
+
+task copyDependencies {
+  dependsOn(copyMainDependencies)
+  dependsOn(copyTestDependencies)
+}
```
dcapwell, line 61: this is for C* IDE support.  IntelliJ only knows are dependencies from directories, so to make sure accord works I need to dump dependencies (main/test) and let IntelliJ know about them.

maedhroz (approved):
Needs a quick rebase, but otherwise, LGTM

accord-core/build.gradle
```diff
@@ -56,15 +56,32 @@ task testJar(type: Jar, dependsOn: testClasses) {
 publishing {
   publications {
     mavenJava(MavenPublication) {
-      artifactId = 'accord'
+      artifactId = accord_artifactId
       pom {
+        name = 'Apache Cassandra Accord'
+        description = 'General purpose transactions library using a leaderless consensus protocol to have highly available transactions'
+        url = 'https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-15%3A+General+Purpose+Transactions'
```
maedhroz, line 63: I guess we don't have a README for https://github.com/apache/cassandra-accord, so this is probably the most useful link for now...

accord-core/build.gradle
maedhroz: Looks like it's in RC3...to risky to start using that now?

https://gradle.org/release-candidate

dcapwell: we don't have time to test it, so rather wait for the final release

accord-core/build.gradle
dcapwell: @belliottsmith any feedback on best URL to use?  This should be allowed to change over time, felt the CEP wiki was the best for now

accord-core/build.gradle
maedhroz: Released :)

`./gradlew wrapper --gradle-version 8.0`

accord-core/build.gradle
belliottsmith: Makes sense yeah

accord-core/build.gradle
dcapwell: :sigh: I still see it there

```
    <dependency>
      <groupId>accord</groupId>
      <artifactId>accord</artifactId>
      <version>1.0-SNAPSHOT</version>
      <scope>compile</scope>
      <optional>true</optional>
    </dependency>
```

PR #30, opened 2023-01-31
CEP-15: (Accord) Migrate Accord away from JDK random to a new interface RandomSource

dcapwell:


dcapwell:
There are a lot of unrelated changes, I can pull out into its own PR if desired.  I ran tests many times to make sure my changes were good and I found other issues... so kept making fixes until I was stable.

accord-core/src/main/java/accord/primitives/RangeDeps.java
```diff
@@ -612,14 +612,7 @@ protected RangeDeps build(TxnId[] txnIds, Range[] ranges, int[] txnIdsToRanges)
...
     {
-        return RangeDeps::compare;
-    }
-
-    public static int compare(Range left, Range right)
-    {
-        int c = left.start().compareTo(right.start());
-        if (c == 0) c = left.end().compareTo(right.end());
-        return c;
+        return Range::compare;
```
dcapwell, line 615: these are the same, so removed duplication of code.

accord-core/src/main/java/accord/utils/RelationMultiMap.java
```diff
@@ -221,7 +221,7 @@ public T build()
             {
                 sortedKeyIndexes = new int[keyCount];
                 sortedKeys = Arrays.copyOf(keys, keyCount);
-                Arrays.sort(sortedKeys);
+                Arrays.sort(sortedKeys, adapter.keyComparator());
```
dcapwell, line 224: I feel that there are more cases, these were just the cases needed to make the test stable...  I tried a patch to make this class typed to make sure we are the proper `Comparable` but that patch got very large, so punted for this work around... I do feel that range should be `Comparable<Range>` instead and just offer a `compareToKey`

accord-core/src/main/java/accord/utils/JDKRandomAPI.java
```diff
@@ -0,0 +1,87 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package accord.utils;
+
+import java.util.Random;
+
+public interface JDKRandomAPI
+{
```
bdeggleston: why is this a separate interface from RandomSource? It seems like they could be merged

accord-core/src/test/java/accord/burn/TopologyUpdates.java
```diff
@@ -185,46 +186,47 @@ private static Collection<Node.Id> allNodesFor(Txn txn, Topology... topologies)
...
+//        Function<CommandSync, Collection<Node.Id>> allNodes = cmd -> node.topology().withUnsyncedEpochs(cmd.route, syncEpoch + 1).nodes();
+//
+//        Ranges ranges = localTopology.ranges();
+//        Stream<MessageTask> messageStream = Stream.empty();
+//        for (long epoch=1; epoch<=syncEpoch; epoch++)
+//        {
+//            messageStream = Stream.concat(messageStream, syncEpochCommands(node, epoch, ranges, allNodes, syncEpoch, COMMITTED_ONLY));
+//        }
+//        return messageStream;
+//    }
```
bdeggleston: why is this commented out?

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -423,6 +424,16 @@ public static <V> AsyncChain<V> reduce(AsyncChain<V> a, AsyncChain<V> b, BiFunct
...
     }
 
+    public static <A, B> AsyncChain<B> reduce(List<? extends AsyncChain<? extends A>> chains, B zero, BiFunction<B, ? super A, B> reducer)
+    {
+        switch (chains.size())
+        {
+            case 0: return AsyncChains.success(zero);
+            case 1: return chains.get(0).map(a -> reducer.apply(zero, a));
+        }
+        return new ReduceWithZero<>(chains, zero, reducer);
```
bdeggleston: I don't think this is needed. The empty case is already handled above, and the only usage of this doesn't need a special initial value

accord-core/src/main/java/accord/utils/async/AsyncChainCombiner.java
```diff
@@ -165,4 +165,26 @@ static <V> boolean canAppendTo(AsyncChain<? extends V> chain, BiFunction<V, V, V
             return reduce.reducer == reducer;
         }
     }
+
+    static class ReduceWithZero<A, B> extends AsyncChainCombiner<A, B>
+    {
+        private final B zero;
+        private final BiFunction<B, ? super A, B> reducer;
+
+        protected ReduceWithZero(List<? extends AsyncChain<? extends A>> inputs, B zero, BiFunction<B, ? super A, B> reducer)
```
bdeggleston: assuming this isn't removed before commit, could we call this ReduceWithInitialValue, or maybe ReduceWithIdentiity?

bdeggleston (changes requested):

accord-core/src/main/java/accord/utils/JDKRandomAPI.java
dcapwell: they could, I did this at the start to keep things easier for me, but we don't touch this interface so can merge

accord-core/src/test/java/accord/burn/TopologyUpdates.java
dcapwell: left a comment about this, I did this and left GH comment; this is dead code and the related logic had a dead lock issue with this [patch](https://issues.apache.org/jira/browse/CASSANDRA-18203).  I am 100% fine making this async, but left it there to have that convo.

This is also from https://issues.apache.org/jira/browse/CASSANDRA-18203 and not this ticket, so you wouldn't see that comment here.

accord-core/src/main/java/accord/utils/async/AsyncChains.java
dcapwell: fair, we know it has 2, I think I did this trying to find a way to merge `Reduce` to handle both cases but after awhile I migrated to use its own class; ill simplify

dcapwell: trying to resolve this comment and I remember why it is needed; `zero` is type `B` but the chains are of type `A`, so can not reduce without keeping the `zero`

accord-core/src/main/java/accord/utils/async/AsyncChainCombiner.java
dcapwell: renamed and updated variables

accord-core/src/test/java/accord/burn/TopologyUpdates.java
dcapwell: migrated to async even though dead-code

accord-core/src/main/java/accord/utils/JDKRandomAPI.java
dcapwell: removed and restructured `RandomSource` so the new methods and the existing ones are next to their related methods

accord-core/src/test/java/accord/burn/TopologyUpdates.java
bdeggleston: Got it, yeah I didn't see the comment. I'll be a little pedantic and say that although the code is unreachable, it does provide value to the project as a tool when debugging some burn test failures, and is therefore not "dead". However, it also isn't run and is therefore quickly stops working if not being used diagnosing failures. Having said that, I'm not sure if the value it provides is still more than the cost to maintain it.

dcapwell: in this context the maintain was super simple, so cost is low.  Now does this function "work" is unknown...  I don't mind keeping it now that it compiles

bdeggleston (approved):
+1

belliottsmith:
I believe this has also been merged now, effectively? Feel free to re-open if I'm mistaken.

PR #33, opened 2023-02-02
Immutable state

bdeggleston:


accord-core/src/main/java/accord/primitives/TxnId.java
```diff
@@ -27,6 +28,9 @@
 
 public class TxnId extends Timestamp
 {
+//    public static final TxnId TARGET = new TxnId(1, 2, 3, new Node.Id(4));
+    public static final TxnId TARGET = new TxnId(1, 4, 3, new Node.Id(2));
```
dcapwell: should we revert?  This is dead code

accord-core/src/main/java/accord/utils/SortedArrays.java
```diff
@@ -21,6 +21,7 @@
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.function.IntFunction;
+import java.util.function.Predicate;
```
dcapwell: unused import

accord-core/src/main/java/accord/local/AsyncCommandStores.java
```diff
@@ -1,92 +0,0 @@
-/*
```
belliottsmith, line 1: Why erase these classes?

accord-core/src/main/java/accord/local/StateUpdate.java
```diff
@@ -0,0 +1,95 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package accord.local;
+
+public class StateUpdate<V, U extends StateUpdate<V, U>>
```
belliottsmith: unused?

accord-core/src/main/java/accord/local/ImmutableState.java
```diff
@@ -0,0 +1,202 @@
...
+ * between when the safe store for it's current operation is created, and closed, and is otherwise marked DORMANT.
+ * While DORMANT, it cannot be read from or updated. This is to help detect leaked instances being read from or
+ * updated when they shouldn't be.
+ *
+ * Once an update has created a superseding instance, the original instance is marked SUPERSEDED. It can then be
+ * marked CLEANING_UP, during which time its fields can be read to assist in determining what has been updated.
+ * From CLEANING_UP, an instance can be marked INVALIDATED to indicate that any access is illegal. An instance
+ * marked SUPERSEDED can also go directly to INVALIDATED, in the case of intermediate updates.
+ */
+public abstract class ImmutableState
```
belliottsmith: It seems like most of this stuff is to debug, primarily during development, but also is potentially easily forgotten to add to subclasses. I wonder if this would be better as something partially added with byte-weaving, or aspect-oriented programming when debugging is enabled? We could automatically add the checks to each sub-method, and also byte-weave in the state field, so we have no runtime cost while also making sure to capture every method, and also avoid polluting the extending classes.

accord-core/src/main/java/accord/messages/Accept.java
```diff
@@ -240,16 +242,15 @@ public void process()
         @Override
         public AcceptReply apply(SafeCommandStore safeStore)
         {
-            Command command = safeStore.command(txnId);
-            switch (command.acceptInvalidate(safeStore, ballot))
+            switch (Commands.acceptInvalidate(safeStore, txnId, ballot))
```
belliottsmith: Maybe with the new style it's better to just return the new `Command` and check if it has our ballot accepted? We could have a helper method to check for redundancy.

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -0,0 +1,1073 @@
...
+        return command.writes().apply(safeStore).flatMap(unused -> unsafeStore.submit(context, callPostApply(txnId)));
+    }
+
+    private static void apply(SafeCommandStore safeStore, Command.Executed command)
+    {
+        applyChain(safeStore, command).begin(safeStore.agent());
+    }
+
+    // TODO (expected, API consistency): maybe split into maybeExecute and maybeApply?
+    private static boolean maybeExecute(SafeCommandStore safeStore, Command command, ProgressShard shard, boolean alwaysNotifyListeners, boolean notifyWaitingOn)
```
belliottsmith: maybe accept `Committed`?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -18,1178 +18,1435 @@
...
     }
 
-    public AcceptOutcome accept(SafeCommandStore safeStore, Ballot ballot, PartialRoute<?> route, Seekables<?, ?> keys, @Nullable RoutingKey progressKey, Timestamp executeAt, PartialDeps partialDeps)
+    private abstract static class AbstractCommand extends Command
     {
-        if (this.promised().compareTo(ballot) > 0)
+        private final TxnId txnId;
+        private final SaveStatus status;
+        private final Status.Durability durability;
+        private final RoutingKey homeKey;
```
belliottsmith, line 172: We probably don't need this except for when `NotWitnessed`, (and perhaps each of the invalidated states) - perhaps we should have a dedicated `NotWitnessed` variant that does not extend this class, and is permitted also for each invalidated state where the definition is unknown? Perhaps `Empty` and `NotWitnessed` could also share a common ancestor?

accord-core/src/main/java/accord/impl/CommandsForKey.java
```diff
@@ -20,97 +20,550 @@
...
+            if (original != null)
+                original.markSuperseded();
+            updated.markActive();
+            safeStore.completeUpdate(this, original, updated);
+            completed = true;
+            return updated;
+        }
+    }
+
+    public static CommandsForKey updateLastExecutionTimestamps(CommandsForKey current, SafeCommandStore safeStore, Timestamp executeAt, boolean isForWriteTxn)
```
belliottsmith: unused?

accord-core/src/main/java/accord/local/SafeCommandStore.java
```diff
@@ -47,11 +48,27 @@
     Command ifLoaded(TxnId txnId);
     Command command(TxnId txnId);
 
+    CommandsForKey ifLoaded(RoutableKey key);
```
belliottsmith: Should we be exposing `CommandsForKey` here and elsewhere? We had gone to some lengths to relegate `CommandsForKey` to an implementation detail, to ensure we can treat key and range transactions consistently (and, also, because it can then more easily be dismantled into dedicated cache structures).

Looking at the patch more generally, it seems `CommandsForKey` has been promoted to a very central concept from an entirely hidden one, and I don't think this is necessary, and is probably counterproductive for both consistency with range transactions and for optimising the behaviour currently encoded by `CommandsForKey`.

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -18,1178 +18,1435 @@
...
+                Invariants.checkState(command.status() == Status.Accepted);
+                return (Preaccepted) complete(Command.updateAttributes(command, this));
+            }
+        }
 
-        return null;
-    }
+        public Accepted markDefined(Ballot promised)
+        {
+            Invariants.checkState(command.hasBeen(Status.AcceptedInvalidate));
```
belliottsmith: PreCommitted state isn't correctly handled. I have addressed this in my suggestion patch.

accord-core/src/main/java/accord/local/SafeCommandStore.java
```diff
@@ -90,6 +107,37 @@ <T> T mapReduce(Seekables<?, ?> keys, Ranges slice,
...
+        return beginUpdate(command(txnId));
+    }
+
+    void completeUpdate(Command.Update update, Command current, Command updated);
+
+    CommandsForKey.Update beginUpdate(CommandsForKey commandsForKey);
+
+    void completeUpdate(CommandsForKey.Update update, CommandsForKey current, CommandsForKey updated);
+
+    PostExecuteContext complete();
```
belliottsmith: This method and some of the others seem to be an implementation detail, and perhaps better not to be on the interface?

accord-core/src/main/java/accord/local/ImmutableState.java
belliottsmith: So, it looks like this has been introduced to grapple with multiple versions of a `Command`. But it seems to me that this isn't sufficient (there's a related just-found bug with range transactions as testimony), but also that this materially harms readability in many unrelated parts of the system. The checks are also not easily maintained, future maintenance will likely cause them to atrophy.

I think we should instead be working to forbid having more than one version accessible to the `CommandStore`, without at least knowing (with strong typing) that one may be stale.

Firstly, I wonder if a cleaner concept mightn't be to have a concept of `SafeCommand` or `LiveCommand` that is an indirect reference to the store's latest version of the `Command` (or has such a pointer for validation purposes, and fails if the local reference is different). In other cases a command should be treated as stale, or we could have a dedicated `StaleCommand` interface.

I think we should also make sure that, as far as possible, we do not keep more than one command around internally. For instance, our `SafeCommandStore` doesn't need to have its own separate copy, it should reference the global map directly, and publish updates to the global map directly. This should be easier without the builder control flow.

WDYT?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -18,1178 +18,1435 @@
...
-     * Note that for ProgressLog purposes the "home shard" is the shard as of txnId.epoch.
-     * For recovery purposes the "home shard" is as of txnId.epoch until Committed, and executeAt.epoch once Executed
-     */
-    public final void homeKey(RoutingKey homeKey)
-    {
-        RoutingKey current = homeKey();
-        if (current == null) setHomeKey(homeKey);
-        else if (!current.equals(homeKey)) throw new AssertionError();
-    }
+        private static class Factory
```
belliottsmith: Do we need to use factories here? These objects are immutable and we don't accept any user-provided objects I don't think, so it is probably cleaner to just have either static or instance methods that advance the state machine? They could be public or package private.

This also avoids the need to unpack sets into mutable ones, then repack them into immutable ones, halving the cost (which is a very large percentage of time spent on the trace that David shared with us privately)

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -18,41 +18,113 @@
...
+            throw new RuntimeException(e);
+        }
+        finally
+        {
+            state.checkIsActive();
+            state.markDormant();
+        }
+    }
+
+    public static <State extends ImmutableState, V> V ensureActiveState(State state, Callable<V> callable)
```
belliottsmith: This method really makes me think we need to find another solution to the stateful lifecycle of the immutable Commands. There's a *lot* of additional boiler-plate around the codebase now with these states, and they make the code harder to read and reason about.

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -18,1178 +18,1435 @@
...
-    protected abstract void setSaveStatus(SaveStatus status);
-
-    public Status status() { return saveStatus().status; }
-    protected void setStatus(Status status) { setSaveStatus(SaveStatus.get(status, known())); }
-
-    public Known known() { return saveStatus().known; }
-
-    public abstract Durability durability();
-    public abstract void setDurability(Durability v);
+        public static Command.Listener listener(TxnId txnId)
```
belliottsmith: Should this be next to Listener?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -18,1178 +18,1435 @@
...
+            case Applied:
+            case Invalidated:
+                return validateCommandClass(status, Executed.class, klass);
+            default:
+                throw new IllegalStateException("Unhandled status " + status);
+        }
     }
 
-    public void setDurability(SafeCommandStore safeStore, Durability durability, RoutingKey homeKey, @Nullable Timestamp executeAt)
+    public static Command addListener(SafeCommandStore safeStore, Command command, CommandListener listener)
```
belliottsmith: Can't help but feel that for add/remove listener and `updateWaitingOn` the balance of cost:benefit isn't necessarily in favour of immutability. We could be selective in our level of immutability, to make the lifecycle a bit easier to manage. But perhaps we can look at this later, once we have deterministic processing and can simply evict the cache directly to memtable (so we won't expect to have more than one live copy, and the immutability is just about clarity).

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -0,0 +1,1073 @@
...
+import java.util.function.Function;
+
+import static accord.api.ProgressLog.ProgressShard.*;
+import static accord.api.ProgressLog.ProgressShard.Local;
+import static accord.local.Commands.EnsureAction.*;
+import static accord.local.Status.*;
+import static accord.local.Status.Known.ExecuteAtOnly;
+import static accord.primitives.Route.isFullRoute;
+
+public class Commands
```
belliottsmith, line 47: Is it definitely clearer to have this in a separate class to Command?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -18,1178 +18,1435 @@
...
+    private static Status.Durability durability(Status.Durability durability, SaveStatus status)
     {
-        return known().definition.compareTo(definition) >= 0;
+        if (status.compareTo(SaveStatus.PreApplied) >= 0 && durability == NotDurable)
+            return Local; // not necessary anywhere, but helps for logical consistency
+        return durability;
     }
 
-    public boolean has(Outcome outcome)
+    public interface CommonAttributes
```
belliottsmith: Should this extend `Command`, or to extract a common interface?

accord-core/src/main/java/accord/local/Command.java
belliottsmith: Since preparing this feedback, I've seen David's traces that show this alone accounting for 7% of cpu time, where as much as 24% of cpu time is also spent on logging (so if we exclude that, we're at around 10% of cpu time). So I think we need to address this, but perhaps it can be in follow-up work, and perhaps this will be ameliorated once we elide transitive dependencies. We might also be able to move listener registration to `NotifyWaitingOn` so that we only register them on demand (saving some wasted work), or else elide transitive dependencies at registration time. Or all of the above.

belliottsmith (changes requested):
I've done a very rough initial pass at review, and the patch looks like it moves things in a good direction. The integration patch in particular looks to be cleaned up by these changes - I enjoyed seeing the negative diff removing all of the StoredX classes (though I have yet to review it properly). 

I do have a few suggestions, and some general concerns about how to make the most of these refactors. I expect I’ll have some more cosmetic suggestions in time as well.

1. I don't think we need to use builders, and think that direct transitions between states will be clearer and more efficient.
2. `CommandsForKey` has been woven throughout the API. This was excised for range transactions, and while I'm not sure if range transactions work right now, it is inconsistent with how their persistence is managed. I think this is quite easy to avoid, and permits us the flexibility to avoid using `CommandsForKey` when we have suitable caches.
3. I am a bit concerned about the statefulness of the immutable objects, their associated boiler-plate and the problems they try to address. The main benefit of this PR is improving developer ergonomics, so I hope we can figure out how to simplify this, and make the semantics more robust.

I have a (somewhat rushed) branch with some suggestions that partially address (1) and (2) here: https://github.com/belliottsmith/cassandra-accord/tree/immutable-state, by moving to direct transitions for `Command` objects (incompletely implemented), and removes `CommandsForKey` from the API, but otherwise keeps its implementation details largely the same.

bdeggleston:
What about making Command more of a state container? Something like AtomicReference, where there is no more than one instance in memory for a given Command/CommandStore, and the ‘actual’ state it contains is immutable and updated atomically. That would mostly eliminate the concerns and additional state required to protect against making decisions with stale state.

accord-core/src/main/java/accord/local/ImmutableState.java
bdeggleston: I just suggested something similar in a response to your main comment. The only scenario where I think a stale command would be useful is in the construction of an update

accord-core/src/main/java/accord/local/Commands.java
bdeggleston: I think so, but it probably depends on who you ask :)

With Command defining a hierarchy of immutable classes, separating the transformations applied to them into another class makes sense to me.

accord-core/src/main/java/accord/local/ImmutableState.java
belliottsmith: > The only scenario where I think a stale command would be useful is in the construction of an update

A stale command _could_ also be useful in listeners when deciding what action to take because of a delta, though I think today we probably don't require it (I do something similar in my suggestion patch for `CommandsForKey` but only take the prior `SaveStatus`). We will probably need to separately track/expose the state machine types anyway, though, as they can change within the same reference. Perhaps we can just call them an `UnsafeCommand` to distinguish them?

>What about making Command more of a state container? Something like AtomicReference, where there is no more than one instance in memory for a given Command/CommandStore

I think this would be an immediate improvement for sure. I think the only difference between our proposals is that I think it might be better to _hide_ this reference (let's call it an `UnsafeCommandRef`), as exposing it directly potentially masks some bugs. Since we expect there to be only one active operation touching the latest value at once, if the `SafeCommand` we pass around internally references an `UnsafeCommandRef` and _also_ the `UnsafeCommand` that was pointed to _when the `SafeCommand` was constructed_, then on each access we can validate that nobody has changed the underlying value, and fail an assertion if they have. Another option is to have CAS semantics, but this only catches erroneous updates and potentially makes our control flow a bit more complicated.

If we only ever construct and _immediately use_ a `SafeCommand` (and we don't have any concurrency) then the approaches are equivalent. But it helps protects against cases (including in the future) where we don't follow this for whatever reason, and reifies concerns of safety around these references to better flag them for consideration by future authors. 🤷 YMMV.

(Also, not committing 100% to my proposed nomenclature, feels like we can maybe do better.)

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: I'm definitely not certain either way, just raising it for some discussion so we can iron it out. I think it is good that we haven't crowded out the `Command` file, but also not great that there's an imperfect separation of concerns. The constraints of how the state machine is advanced are in part encoded in `Command`, and we also reference `Commands` and `CommandsForKey`. Static methods also tend to be a bit more cluttered, and have to cast between `Command` types that mightn't be necessary for instance methods.

For instance, it's possible that putting `Command` in its own package with all its versions in separate class files might make it tractable to put the functionality of `Commands` in the relevant type, and use the class hierarchy to in part encode the variations in behaviour.

Or, we might simply define `Command` as an almost-empty placeholder that _only_ defines the methods in `Commands` AND the existing abstract methods only, then have `AbstractCommand` in a separate class that defines the state machine types. We can still benefit from instance methods (that just map directly from their current static methods in `Commands`) and perhaps have slightly cleaner separation of concerns.

It might be that neither of these are cleaner, but exploring the possible structures here will have long term impact.

belliottsmith: I guess we also have another option if we introduce a reference type (which it sounds like we probably will in some shape or form), which is to move these transitions there. We don't get the benefit of instance methods, but we do associate it with the actual type we store the states in, so it feels a bit more natural for discoverability.

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -18,1178 +18,1435 @@
...
     {
-        return known().outcome.compareTo(outcome) >= 0;
+        TxnId txnId();
+        Status.Durability durability();
+        RoutingKey homeKey();
+        RoutingKey progressKey();
+        Route<?> route();
+        PartialTxn partialTxn();
+        PartialDeps partialDeps();
+        ImmutableSet<CommandListener> listeners();
```
belliottsmith: I think this breaks determinism?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -18,1178 +18,1435 @@
...
+        public String toString()
+        {
+            return "Command@" + System.identityHashCode(this) + '{' + txnId + ':' + status + '}';
+        }
 
-        setStatus(Accepted);
-        safeStore.progressLog().accepted(this, shard);
-        notifyListeners(safeStore);
+        @Override
+        public int hashCode()
```
belliottsmith: maybe throw `UnsupportedOperationException`, since we don't expect this method to be used?

accord-core/src/main/java/accord/local/AsyncCommandStores.java
bdeggleston: I think I’d meant to talk to you about this and never did. Some of the state tracking happening in C* was moved into Accord as explicit state changes in abstract base classes (ie: AbstractSafeCommandStore), and these changes were not compatible with the split sync/async class hierarchy. Since accord is meant to be async, I removed the sync class. I’m not sure this is necessary with the current state of the code. 

While we’re talking about it though, it seemed that these classes were meant to make sync and async operation a bit cleaner and remove some code duplication. Was there another goal that’s  less obvious?

bdeggleston:
> CommandsForKey has been woven throughout the API. This was excised for range transactions, and while I'm not sure if range transactions work right now, it is inconsistent with how their persistence is managed. I think this is quite easy to avoid, and permits us the flexibility to avoid using CommandsForKey when we have suitable caches.

Yeah the range transactions patch that removed CommandsForKey landed pretty close to when the final rebase began and adapting the CFK changes in the immutable state patch ended up being more of a rebase kludge than a proper refactor. This was on my list of things to polish a bit, now that I’m more familiar with the the cfk changes. Just to be clear though, you’re referring to the reintroduction of cfk getters etc to the safe command store, not the changes to CommandsForKey itself?

accord-core/src/main/java/accord/impl/CommandsForKey.java
bdeggleston: this is used in the C* integration

belliottsmith: 👍. We should probably annotate such methods in some way, and suppress linting. I thought I had previously seen VisibleForImplementation but it looks like I imagined it.

accord-core/src/main/java/accord/local/AsyncCommandStores.java
belliottsmith: Yeah, the sync stores avoid any allocations or additional work in the normal case. Having both sync and async maintained ensures our API decisions are compatible with both approaches. Removing one is likely to lead to a situation where unpicking some dependency is much harder in future. Since the sync approach is potentially more efficient long term, I put quite a bit of work into ensuring both paradigms could be supported equally efficiently, so that we can more easily migrate in future.

This is the second time this work has been erased, so it would be nice to try and avoid that in future if possible. Perhaps we can add some commentary to this effect.

> Since accord is meant to be async

Not sure what you mean by that? Most of the `CommandStore` processing is synchronous, we currently have a thread per `CommandStore` but its work serving these requests is synchronous. I am far from certain that this will continue to be true indefinitely, so we shouldn't assume it I don't think. For data that is in cache, using the thread that happens to be running will likely be more efficient. Also, the separation of loading/writing concerns and thread in general in Cassandra is very likely to evolve. Virtual threads is a mid-term horizon that may very fundamentally alter what is optimal too.

belliottsmith:
Right, and the `PreExecutionContext` and `PostExecutionContext`, `completeOperation` etc. You should be able to see in my suggestion patch, where I have moved all of these details to the impl package alongside `AbstractSafeCommandStore`.

accord-core/src/main/java/accord/impl/SimpleProgressLog.java
```diff
@@ -712,34 +713,39 @@ public void preaccepted(Command command, ProgressShard shard)
...
         {
+            Command command = safeStore.command(txnId);
             recordApply(command.txnId());
             // this is the home shard's state ONLY, so we don't know it is fully durable locally
             ensureSafeOrAtLeast(command, shard, ReadyToExecute, Expected);
         }
 
         @Override
-        public void invalidated(Command command, ProgressShard shard)
+        public void invalidated(SafeCommandStore safeStore, TxnId txnId, ProgressShard shard)
```
belliottsmith: Why don't we pass `Command` anymore here?

accord-core/src/main/java/accord/local/ImmutableState.java
belliottsmith: I think probably it will be easier to discuss tweaks like this once we have the code structured broadly this way, whichever way seems appropriate. 

Also, let me know when you've had a chance to look at my suggestion branch and if you want to a higher bandwidth discussion. I think it may simplify things a little but it certainly affects this stuff. 

We can potentially avoid buffering state in `AbstractSafeCommandStore` since we produce a new `Command` at the end of each transformation, and push updates either through `SafeCommand` or something similar and have them take effect globally immediately, avoiding multiple copies there.

bdeggleston:
> Right, and the PreExecutionContext and PostExecutionContext, completeOperation etc. You should be able to see in my suggestion patch, where I have moved all of these details to the impl package alongside AbstractSafeCommandStore.

That seems reasonable. I've looked through the suggestion patch, but haven't spent a lot of time reviewing it. I've mainly been focused on burn tests and working on the LiveCommand change we talked about

accord-core/src/main/java/accord/local/Command.java
bdeggleston: I'd added these because I was having problems with ambiguous create methods, ie calling Accepted.create and it calling Preaccepted.create instead. Renaming to createPreAccepted or preAccepted would be better though

accord-core/src/main/java/accord/impl/SimpleProgressLog.java
bdeggleston: to prevent stale objects being passed in. I'd expect that to accept LiveCommand or whatever once that's done

accord-core/src/main/java/accord/impl/CommandsForKey.java
bdeggleston: No we used to have it, but removed it in favor of `SerializerSupport` inner classes

accord-core/src/main/java/accord/local/Command.java
belliottsmith: +1 to preAccepted etc.

belliottsmith:
> I've looked through the suggestion patch, but haven't spent a lot of time reviewing it

It's not complete so probably not worth _reviewing_ it per se. But I went 90% of the way there and decided it was better to leave the rest until we could compare notes on how you want to proceed.

accord-core/src/main/java/accord/local/SafeCommandStore.java
bdeggleston: CommandsForKey are no longer visible/used outside of the impl package

accord-core/src/main/java/accord/local/ImmutableState.java
bdeggleston: Added LiveCommand, and it's definitely an improvement in readability and complexity. I agree that the naming isn’t my favorite, but I don’t have any better ideas at the moment. I think I may prefer SafeCommand to LiveCommand just for consistency with SafeCommandStore. Maybe CommandReference or CommandRef?

Although there isn’t an overarching builder for command updates, I have kept CommonAttributes separate which can be converted to mutable command attributes and updated via the boilerplate methods for setting things like route, partialTxn, etc, and is only converted to mutable if something needs changing. This seemed like a reasonable approach given the number of attributes set and some of the shared data, or in the case of durability, the frequency of updates. If anyone has strong feelings about that though, I can work on untangling them into some methods that return values which are passed into the command update.

accord-core/src/main/java/accord/local/Commands.java
bdeggleston: Having added the reference type, I’m now on the fence between keeping the logic in Commands and moving it to LiveCommand. I don’t think there’s a big difference either way. I put the commands for key logic in the LiveCommandsForKey class, and it seems fine there. Depending on how partial mutability goes, it may be slightly easier to perform operations as instance methods on LiveCommand if the mutable state lives on LiveCommand

accord-core/src/main/java/accord/local/ImmutableState.java
belliottsmith: Awesome, I'll try to take a proper look over the weekend, or else Monday.

> converted to mutable command attributes

I think this is a huge improvement - though I think we should shrink the boilerplate to {{.mutable()}}. 

I think we should also still consider removing the mutable step entirely - at least in the common path. This approach requires two copies of the sets we are updating, for instance, which is a cost we can avoid with a direct factory. It also probably makes the optimiser work harder, as each time we maybe make mutable it has to compile multiple branches that might involve copying a bunch of fields, making our methods larger and maybe impacting things like inlining.

I think, surprisingly, there aren't actually usually many fields updated and the ones we update are quite consistent. Basically it mostly happens in `set`, so we want to update `route`, `progressKey` and then whichever fields you actually need for the new state (some of deps, txn, executeAt usually). `updateHomeKey` is basically a no-op for most paths. 

I think it could make sense to keep the mutable approach for the less common updates (e.g. updating home key or progress key, which usually won't be set by themselves), and have dedicated methods for the more common ones.

If you take a look at my suggestions patch, `Commands.update` shows the main scope of the change. It can probably be cleaned up further based on your recent improvements. If you're amenable to it, I can try to rebase on what you've done here and clean it up.

belliottsmith: To be fair we might get most of the benefit by just having set accept a Mutable and having a dedicated add/remove listener method

belliottsmith: On third consideration maybe it’s better to leave further changes here until we have the commit log stuff in, as we may need to tweak things a little more then and no point overthinking this beforehand. I may see which of the more minor changes in my suggestion patch are worth retaining.

belliottsmith: On third consideration I think it might be better not to overthink this before some later work comes along. But, I think there remain some small clarity improvements that are possible. For instance, I don't think we necessarily need `pendingSeekableRegistrations` or `pendingSeekablesRegistrations` particularly if we move to invoking the transition method inside of `set`, so we know we have fully assembled everything, as `set` is responsible for invoking register.

Another option is to just remove `register` entirely and leave it to the implementation to manage this when we `update` the `LiveCommand`, when it already has access to a mutable set of listeners. This might be both cleaner and easier.

A third option would be to add it to `LiveCommand` itself. This might be easiest.

belliottsmith: Regarding `LiveCommand`, I agree `SafeCommand` is nicer for consistency yeah. Though I think it would be better for symmetry for `SafeCommand` to refer to the temporary reference we take for the duration of an operation. This would permit us to also define restrictions we think might help safety. For instance, most state transitions we can easily require that they update the command at most once. Maybe we can for all of them? 

I think the cases where we don't may have been the source of a number of the issues we encountered, and could potentially cause future issues as they have the potential to leave the state machine in a partially updated state. It would be nice to be able to assemble an entire new image at termination within our state machine update methods in fact, rather than leaving it to `pending` methods.

I think it's possible that all of the updates fulfil this criteria now except for `NotifyWaitingOn`? I could have a look at updating `NotifyWaitingOn` if we agree it would be clearer to enforce this rule.

At the very least it would be nice to declare with the type system where we _can't_ follow this rule so we know to be more careful there.

accord-core/src/main/java/accord/api/ProgressLog.java
```diff
@@ -26,6 +26,7 @@
 import accord.local.Command;
 import accord.local.CommandStore;
 import accord.local.Node.Id;
+import accord.local.SafeCommandStore;
```
dcapwell: unused import

accord-core/src/main/java/accord/api/VisibleForImplementation.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package accord.api;
+
+/**
+ * Indicates an element exists or is more visible than otherwise needed to support implementations
+ */
+
+public @interface VisibleForImplementation
```
dcapwell, line 32: does this imply binary/source compatibility?  If so, what's the difference between this and normal `public`?

If this is for documentation, then should add the following

```
@Documented
@Inherited
@Retention(RetentionPolicy.SOURCE)
```

This will scope the annotation to the source only, but also tell javadoc to annotate this as part of the public api

```
$ javap -v -cp accord-core/build/classes/java/main/ accord.local.Command.SerializerSupport | grep VisibleForImplementation
$
```

```
$ grep VisibleForImplementation accord-core/build/docs/javadoc/accord/local/Command.SerializerSupport.html
<pre><a href="../../accord/api/VisibleForImplementation.html" title="annotation in accord.api">@VisibleForImplementation</a>
```

Without that change, the following is what we get


```
$ javap -v -cp accord-core/build/classes/java/main/ accord.local.Command.SerializerSupport | grep VisibleForImplementation
Picked up _JAVA_OPTIONS: -Djava.net.preferIPv4Stack=true
  #64 = Utf8               Laccord/api/VisibleForImplementation;
$ grep VisibleForImplementation accord-core/build/docs/javadoc/accord/local/Command.SerializerSupport.html
$
```

accord-core/src/main/java/accord/local/ImmutableState.java
bdeggleston: I like the idea of a request scoped command container, though I don't think we can restrict them to a single update per request without some reworking of `Commands#maybeExecute` as well

dcapwell:
Burn test hangs for me on `496289992086939886` seed; below is the stack trace

```
"Test worker@1" prio=5 tid=0x1 nid=NA waiting
  java.lang.Thread.State: WAITING
	  at jdk.internal.misc.Unsafe.park(Unsafe.java:-1)
	  at java.util.concurrent.locks.LockSupport.park(LockSupport.java:194)
	  at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:885)
	  at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1039)
	  at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1345)
	  at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:232)
	  at accord.utils.async.AsyncResults.getBlocking(AsyncResults.java:346)
	  at accord.impl.InMemoryCommandStore$Synchronized.executeSync(InMemoryCommandStore.java:822)
	  at accord.local.SyncCommandStores$SyncMapReduceAdapter.apply(SyncCommandStores.java:75)
	  at accord.local.SyncCommandStores$SyncMapReduceAdapter.apply(SyncCommandStores.java:60)
	  at accord.local.CommandStores.mapReduce(CommandStores.java:395)
	  at accord.local.CommandStores.mapReduceConsume(CommandStores.java:371)
	  at accord.local.SyncCommandStores.mapReduceConsume(SyncCommandStores.java:102)
	  at accord.local.Node.mapReduceConsumeLocal(Node.java:280)
	  at accord.messages.Commit.process(Commit.java:148)
	  at accord.messages.Defer.onChange(Defer.java:90)
	  at accord.local.SafeCommandStore.notifyListeners(SafeCommandStore.java:111)
	  at accord.local.Commands.preacceptOrRecover(Commands.java:144)
	  at accord.local.Commands.recover(Commands.java:91)
	  at accord.messages.BeginRecovery.apply(BeginRecovery.java:87)
	  at accord.messages.BeginRecovery.apply(BeginRecovery.java:46)
	  at accord.impl.InMemoryCommandStore$State.executeInContext(InMemoryCommandStore.java:455)
	  at accord.impl.InMemoryCommandStore$State.executeInContext(InMemoryCommandStore.java:470)
	  at accord.impl.InMemoryCommandStore$Synchronized.lambda$executeSync$2(InMemoryCommandStore.java:812)
	  at accord.impl.InMemoryCommandStore$Synchronized$$Lambda$402.402808045.run(Unknown Source:-1)
	  at accord.impl.InMemoryCommandStore$Synchronized.maybeRun(InMemoryCommandStore.java:762)
	  - locked <0xe62> (a accord.impl.InMemoryCommandStore$Synchronized)
	  at accord.impl.InMemoryCommandStore$Synchronized.enqueueAndRun(InMemoryCommandStore.java:777)
	  at accord.impl.InMemoryCommandStore$Synchronized.executeSync(InMemoryCommandStore.java:809)
	  at accord.local.SyncCommandStores$SyncMapReduceAdapter.apply(SyncCommandStores.java:75)
	  at accord.local.SyncCommandStores$SyncMapReduceAdapter.apply(SyncCommandStores.java:60)
	  at accord.local.CommandStores.mapReduce(CommandStores.java:395)
	  at accord.local.CommandStores.mapReduceConsume(CommandStores.java:371)
	  at accord.local.SyncCommandStores.mapReduceConsume(SyncCommandStores.java:102)
	  at accord.local.Node.mapReduceConsumeLocal(Node.java:280)
	  at accord.messages.BeginRecovery.process(BeginRecovery.java:79)
	  at accord.messages.TxnRequest.process(TxnRequest.java:144)
	  at accord.local.Node.lambda$receive$16(Node.java:499)
	  at accord.local.Node$$Lambda$364.1806880779.run(Unknown Source:-1)
	  at accord.impl.basic.Cluster.now(Cluster.java:206)
	  at accord.local.Node.receive(Node.java:499)
	  at accord.impl.basic.Cluster.processNext(Cluster.java:172)
	  at accord.impl.basic.Cluster.processPending(Cluster.java:124)
	  at accord.impl.basic.Cluster.run(Cluster.java:242)
	  at accord.burn.BurnTest.burn(BurnTest.java:269)
	  at accord.burn.BurnTest.run(BurnTest.java:341)
	  at accord.burn.BurnTest.testOne(BurnTest.java:329)
```

dcapwell:
A lot of my feedback was not applied with regard to the AsyncChains patch.  I also see you commented that it was updated but I only see the feedback in this PR; so created a PR addressing the more critical issues (left nits)

Take a look at my feedback branch: https://github.com/dcapwell/cassandra-accord/tree/feedback/immutable-state

dcapwell:
Looking at the burn test this is what I see, `accord.impl.InMemoryCommandStore.Synchronized#maybeRun` has `active != null` causing the lambda not to trigger, this is caused by the fact `maybeRun` exists twice on the stack frame, so the `AsyncChains.getBlocking` deadlocks the burn test (as its actually single threaded)

dcapwell:
here are a few burn test seeds that have issues for me

`-6934708324826057613L` super slow... I gave up after 5m... I changed the test code so `Timeout` doesn't fill a stack trace, it looks like this seed hits a very bad edge case and takes forever to run... its doing work... just takes forever...

`-7237239287958330617L` and `-870708989819710792L`

```
15:00:24.490 [Test worker] ERROR accord.impl.InMemoryCommandStore - Uncaught exception
java.lang.IllegalStateException: null
	at accord.utils.Invariants.illegalState(Invariants.java:44)
	at accord.utils.Invariants.illegalState(Invariants.java:49)
	at accord.utils.Invariants.checkState(Invariants.java:91)
	at accord.local.Commands.preacceptOrRecover(Commands.java:117)
	at accord.local.Commands.preaccept(Commands.java:86)
	at accord.coordinate.CheckOn$OnDone.apply(CheckOn.java:210)
	at accord.coordinate.CheckOn$OnDone.apply(CheckOn.java:128)
```

belliottsmith:
Doing work but never terminating usually means something is bad about data propagation somewhere for completing a transaction so the progress log never terminates

dcapwell:
Added more logging for seed `4580351883406638196L` to see what messages are going on, and its all

```
...
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:5, to:3, id:25126, body:TaskRequest{Sync:[1,33,2,1]:1:50}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:9, to:2, id:17414, body:TaskRequest{Sync:[1,37,2,1]:1:2}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:8, to:1, id:36527, body:TaskRequest{Sync:[1,37,2,1]:1:14}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:2, to:7, id:24062, body:TaskRequest{Sync:[1,206,2,1]:1:28}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:5, to:7, id:25127, body:TaskRequest{Sync:[1,33,2,1]:1:9}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:8, to:4, id:36528, body:TaskRequest{Sync:[1,37,2,1]:1:67}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:7, to:2, id:17728, body:TaskRequest{Sync:[1,37,2,1]:1:2}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:8, to:1, id:36529, body:TaskRequest{Sync:[1,37,2,1]:1:14}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:8, to:2, id:36530, body:TaskRequest{Sync:[1,37,2,1]:1:2}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:5, to:7, id:25128, body:TaskRequest{Sync:[1,33,2,1]:1:9}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:3, to:7, id:23708, body:TaskRequest{Sync:[1,206,2,1]:1:28}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:3, to:5, id:23709, body:TaskRequest{Sync:[1,206,2,1]:1:50}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:6, to:3, id:16359, body:TaskRequest{Sync:[1,37,2,1]:1:50}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:8, to:2, id:36531, body:TaskRequest{Sync:[1,37,2,1]:1:2}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:6, to:1, id:16360, body:TaskRequest{Sync:[1,37,2,1]:1:58}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:7, to:2, id:17729, body:TaskRequest{Sync:[1,37,2,1]:1:2}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:4, to:7, id:34039, body:TaskRequest{Sync:[1,33,2,1]:1:9}}
INFO  2023-02-22-12:30:40 BurnTest.java:223 - Pending {from:4, to:2, id:34040, body:TaskRequest{Sync:[1,37,2,1]:1:67}}
...
```

> Doing work but never terminating usually means something is bad about data propagation somewhere for completing a transaction so the progress log never terminates

Yeah, that's what it looks like is happening

accord-core/src/main/java/accord/impl/SimpleProgressLog.java
```diff
@@ -231,15 +233,15 @@ void run(Command command)
                                         // should have found enough information to apply the result, but in case we did not reset progress
                                         if (progress() == Investigating)
                                             setProgress(Expected);
-                                    });
+                                    }).begin(commandStore.agent());
```
dcapwell, line 236: I really wish we can improve our build to detect a `AsyncChain` returned without calling `begin`; I manually had to inspect every call site...

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -629,6 +629,8 @@ public <T> T mapReduce(Seekables<?, ?> keysOrRanges, Ranges slice, TestKind test
             Map<Range, List<Command>> collect = new TreeMap<>(Range::compare);
             commandStore.rangeCommands.forEach(((txnId, rangeCommand) -> {
                 Command command = rangeCommand.command.value();
+                if (command == null)
```
belliottsmith: I don't think this should ever be null, so we might want to look more closely into why this happens?

accord-core/src/main/java/accord/local/SafeState.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package accord.local;
+
+/**
+ * State scoped to a single request that references global state
+ */
+public interface SafeState<T>
```
belliottsmith: This could be an implementation-only interface, I think? We only need `SafeCommand` outside of the `CommandForKey` implementation context.

accord-core/src/main/java/accord/local/SafeState.java
```diff
@@ -0,0 +1,40 @@
...
+    T current();
+    void invalidate();
+    boolean invalidated();
+
+    default boolean isEmpty()
+    {
+        return current() == null;
+    }
+
+    default void checkNotInvalidated()
```
belliottsmith: Does this need to be an interface method?

accord-core/src/main/java/accord/utils/ThreadPoolScheduler.java
```diff
@@ -96,7 +95,7 @@ public void stop()
         }
         catch (InterruptedException e)
         {
-            throw new UncheckedInterruptedException(e);
+            throw new RuntimeException(e);
```
belliottsmith, line 98: Still need to figure out a solution to this, as it will interplay with simulation in C*

accord-core/src/main/java/accord/utils/SortedArrays.java
belliottsmith: I think still unused?

accord-core/src/main/java/accord/utils/DeterministicIdentitySet.java
```diff
@@ -117,6 +121,12 @@ public boolean add(T item)
         entry.next.prev = entry;
         return true;
     }
+    // we add to the front, and iterate in reverse order, so that we can add and remove while iterating without modifying the set we iterate over
+    @Override
+    public boolean add(T item)
```
belliottsmith, line 126: defunct to split these?

accord-core/src/main/java/accord/primitives/Txn.java
```diff
@@ -150,7 +149,13 @@ && read().equals(txn.read())
         @Override
         public int hashCode()
         {
-            throw new UnsupportedOperationException();
```
belliottsmith: Where do we need this implementation?

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -130,18 +134,19 @@ public synchronized void onChange(SafeCommandStore safeStore, Command command)
...
-            read(safeStore, command);
+            read(safeStore, command.asCommitted());
     }
 
     @Override
     public synchronized ReadNack apply(SafeCommandStore safeStore)
     {
-        Command command = safeStore.command(txnId);
-        Status status = command.status();
+        SafeCommand liveCommand = safeStore.command(txnId);
```
belliottsmith: liveCommand -> safeCommand?

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -61,6 +61,7 @@ public static ReadData create(TxnId txnId, Seekables<?, ?> scope, long executeAt
     private transient boolean isObsolete; // TODO (low priority, semantics): respond with the Executed result we have stored?
     private transient BitSet waitingOn;
     private transient int waitingOnCount;
+    private transient BitSet submitted;
```
belliottsmith: is this for debug purposes?

accord-core/src/main/java/accord/coordinate/Coordinate.java
```diff
@@ -32,21 +32,21 @@
...
+import accord.utils.async.AsyncResults;
+
 /**
  * Perform initial rounds of PreAccept and Accept until we have reached agreement about when we should execute.
  * If we are preempted by a recovery coordinator, we abort and let them complete (and notify us about the execution result)
  *
  * TODO (desired, testing): dedicated burn test to validate outcomes
  */
-public class Coordinate extends AsyncFuture<Result> implements Callback<PreAcceptReply>, BiConsumer<Result, Throwable>
+public class Coordinate extends AsyncResults.Settable<Result> implements Callback<PreAcceptReply>, BiConsumer<Result, Throwable>
```
belliottsmith: nit: Maybe call it a `SettableResult` so we can import the inner class and make the use sites a bit cleaner?

accord-core/src/main/java/accord/messages/PreAccept.java
```diff
@@ -100,12 +100,12 @@ public PreAcceptReply apply(SafeCommandStore safeStore)
                     calculatePartialDeps(safeStore, txnId, partialTxn.keys(), txnId, safeStore.ranges().between(minUnsyncedEpoch, txnId.epoch())));
         }
 
-        Command command = safeStore.command(txnId);
-        switch (command.preaccept(safeStore, partialTxn, route != null ? route : scope, progressKey))
```
belliottsmith, line 104: Not a strongly held suggestion, just wondering: might it be cleaner in some cases to pass the `safeCommand` in here so we have it for use after? Or even perhaps to invoke the preaccept method on `safeCommand` itself?

accord-core/src/main/java/accord/messages/Apply.java
```diff
@@ -120,7 +123,7 @@ public Iterable<TxnId> txnIds()
     @Override
     public Seekables<?, ?> keys()
     {
-        return Keys.EMPTY;
+        return keys;
```
belliottsmith, line 126: Do we need `keys` here? I don't think we need them loaded until we perform the _apply_, and we should have access to the local txn `keys` when scheduling that?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -0,0 +1,1083 @@
...
+    }
+
+    protected static void postApply(SafeCommandStore safeStore, TxnId txnId)
+    {
+        logger.trace("{} applied, setting status to Applied and notifying listeners", txnId);
+        Command command = safeStore.command(txnId).applied();
+        safeStore.notifyListeners(command);
+    }
+
+    private static Function<SafeCommandStore, Void> callPostApply(TxnId txnId)
```
belliottsmith: Maybe merge this with `postApply`? My preference would be to call this method `postApply` and inline the old `postApply` here.

accord-core/src/main/java/accord/local/Command.java
belliottsmith: Thoughts? I think this is still valid, and I think it might be preferable to split NotWitnessed from AbstractCommand

accord-core/src/main/java/accord/local/PreLoadContext.java
```diff
@@ -52,6 +57,13 @@
      */
     Seekables<?, ?> keys();
 
+    default boolean isSubsetOf(PreLoadContext superset)
+    {
+        Set<TxnId> superIds = Sets.newHashSet(superset.txnIds());
```
belliottsmith: might be nice to try to cast first, or else to require that PreLoadContext fields are already sets?

accord-core/src/main/java/accord/local/ContextValue.java
```diff
@@ -0,0 +1,89 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package accord.local;
+
+public class ContextValue<Value>
```
belliottsmith: delete?

accord-core/src/main/java/accord/local/CommonAttributes.java
```diff
@@ -0,0 +1,208 @@
...
+    TxnId txnId();
+    Status.Durability durability();
+    RoutingKey homeKey();
+    RoutingKey progressKey();
+    Route<?> route();
+    PartialTxn partialTxn();
+    PartialDeps partialDeps();
+    Listeners.Immutable listeners();
+
+    default Mutable mutableAttrs()
```
belliottsmith: `mutable`?

accord-core/src/main/java/accord/local/CommonAttributes.java
```diff
@@ -0,0 +1,208 @@
...
+import accord.primitives.TxnId;
+import accord.utils.Invariants;
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
+
+import static accord.utils.Utils.ensureImmutable;
+import static accord.utils.Utils.ensureMutable;
+
+// TODO: remove or cleanup
```
belliottsmith: perhaps make this a longer term `TODO (desired)` or something? I think it's clean enough for now, only being used sparingly.

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -359,29 +353,23 @@ public <O> void mapReduceConsume(PreLoadContext context, RoutingKey key, long mi
...
+        AsyncChain<O> reduced = mapReduce(context, commandStoreIds, mapReduceConsume);
+        reduced.begin(mapReduceConsume);
     }
 
-    protected <T1, T2, O> T1 mapReduce(PreLoadContext context, Routables<?, ?> keys, long minEpoch, long maxEpoch, MapReduce<? super SafeCommandStore, O> mapReduce,
-                                       MapReduceAdapter<? super S, T1, T2, O> adapter)
+    public <O> AsyncChain<O> mapReduce(PreLoadContext context, Routables<?, ?> keys, long minEpoch, long maxEpoch, MapReduce<? super SafeCommandStore, O> mapReduce)
     {
-        T2 accumulator = adapter.allocate();
+        List<AsyncChain<O>> chains = new ArrayList<>();
```
belliottsmith: Perhaps this could be a simple `AsyncChain<O>` that is either null, a simple `AsyncChain` (the result of one submit), or a reducer that has already been constructed from the existing one and the new one - it looks like `AsyncChains.reduce` already adds to the existing reducer if it already exists.

accord-core/src/main/java/accord/local/CommandListener.java
```diff
@@ -22,7 +22,7 @@
 
 public interface CommandListener
 {
-    void onChange(SafeCommandStore safeStore, Command command);
+    void onChange(SafeCommandStore safeStore, TxnId txnId);
```
belliottsmith: pass either the `SafeCommand` or the current `Command`?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -18,85 +18,289 @@
...
+        }
     }
 
-    public boolean has(Outcome outcome)
+    public static class Listener implements CommandListener
     {
-        return known().outcome.compareTo(outcome) >= 0;
+        protected final TxnId listenerId;
+
+        private Listener(TxnId listenerId)
```
belliottsmith: Does it need to be private if we just have a simple identical static factory method?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -109,1087 +313,758 @@ public boolean is(Status status)
...
-            // dependency cannot be a predecessor if it executes later
-            logger.trace("{}: {} executes after us. Do not insert.", txnId(), dependency.txnId());
+            return isWaitingOnApply() ? waitingOnApply.firstEntry().getValue() : null;
         }
-        else if (dependency.hasBeen(Applied))
+
+        public boolean hasBeenWitnessed()
         {
-            logger.trace("{}: {} has been applied. Do not insert.", txnId(), dependency.txnId());
+            return partialTxn() != null;
```
belliottsmith: Perhaps use `Known`?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -109,1087 +313,758 @@ public boolean is(Status status)
...
+    {
+        return saveStatus().known;
+    }
 
-    public abstract Command addListener(CommandListener listener);
-    public abstract void removeListener(CommandListener listener);
-    protected abstract void notifyListeners(SafeCommandStore safeStore);
+    public boolean hasBeenWitnessed()
+    {
+        return partialTxn() != null;
```
belliottsmith: Do we need to override here?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -109,1087 +313,758 @@ public boolean is(Status status)
...
+            return executeAt.equals(that.executeAt)
+                    && Objects.equals(partialTxn, that.partialTxn)
+                    && Objects.equals(partialDeps, that.partialDeps);
+        }
 
-        safeStore.progressLog().committed(this, shard);
+        @Override
+        public int hashCode()
+        {
+            int hash = super.hashCode();
```
belliottsmith: Do we use this impl, or should we inherit the UOE?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -0,0 +1,1083 @@
...
+        // the command instance from memory between now and the write completing (and post apply being called)
+        CommandStore unsafeStore = safeStore.commandStore();
+        TxnId txnId = command.txnId();
+        PreLoadContext context = command.contextForSelf();
+        return command.writes().apply(safeStore).flatMap(unused -> unsafeStore.submit(context, callPostApply(txnId)));
+    }
+
+    private static void apply(SafeCommandStore safeStore, Command.Executed command)
+    {
+        applyChain(safeStore, command).begin(safeStore.agent());
```
belliottsmith: Similarly, maybe merge with `applyChain`?

belliottsmith (commented):
I think the patch is looking really good now, and the `AsyncChain` changes also look good woven through here. 

I have left a bunch of minor feedback that should be addressed, and I might have some more feedback in a couple of areas if I have time to take a more detailed looked, but I think overall it's pretty much ready to ship and it's just cleaning up (modulo any failures of course, and integration issues).

accord-core/src/main/java/accord/utils/ThreadPoolScheduler.java
bdeggleston: David and I were talking about checking the exception cause on the C* side and handling it like an UncheckedInterruptedException if it's an IterruptedException, though I still need to do it

accord-core/src/main/java/accord/local/CommonAttributes.java
bdeggleston: I'd  called it mutable at first, but felt it was a bit misleading for Command to have a `mutable` method that didn't return a mutable command. Writing that now, I realize that the return type probably clarifies things. WDYT?

belliottsmith: I don’t feel strongly about it, I just find mutableAttrs to slightly jar when reading. really the attrs are like a builder, which is not uncommon to see as a mutable version for constructing a new immutable object.

another option would be `deconstruct`, or perhaps `builder` and move Mutable in the hierarchy

accord-core/src/main/java/accord/utils/async/AsyncResults.java
```diff
@@ -0,0 +1,327 @@
...
+    public static <V> AsyncResult.Settable<V> settable()
+    {
+        return new Settable<>();
+    }
+
+    /**
+     * An AsyncResult that also implements Runnable
+     * @param <V>
+     */
+    public static class Unscheduled<V> extends AbstractResult<V> implements Runnable
```
belliottsmith: `RunnableResult`?

accord-core/src/main/java/accord/local/Command.java
bdeggleston: To clarify, other commands wouldn't need it because they should have a route which already has a home key?

belliottsmith: Right, and we don’t need to maintain it at all if we’re more than NotWitnessed. But I think it would in general potentially be good to have NotWitnessed be a separate thing

bdeggleston: What about an Unwitnessed route implementation that only has a home key for that case? It looks like we could move the progressKey in there as well, since it’s only updated alongside the homeKey. That would remove unneeded attributes from the other classes without complicating CommonAttributes.Mutable and/or the places that use it.

Since we no longer need an EMPTY command, we could collapse most of AbstractCommand into Command, with status and durability moved into PreAccepted, with NotWitnessed extending Command.

accord-core/src/main/java/accord/utils/DeterministicIdentitySet.java
bdeggleston: `add` throws an `UnsupportedOperationException` in `Listeners.Immutable`, so `addInternal` is used when making immutable copies. See the copy constructor in this class

accord-core/src/main/java/accord/primitives/Txn.java
bdeggleston: I don't think we do. Muscle memory implementation :)

accord-core/src/main/java/accord/messages/ReadData.java
bdeggleston: It's used in `ReadData#read`  to prevent submitting a read to the same command store multiple times, and having it prematurely mark the read complete. I don't remember the details, but there was definitely a burn test I fixed a while ago where that was happening.

belliottsmith: Hmm. That shouldn’t be happening, so we should maybe rollback and figure out what’s happening? I can take a look if you have a seed.

accord-core/src/main/java/accord/local/Command.java
belliottsmith: Oh, hmm. I'm not sure about complicating the idea of a Route, as it is used quite widely in the codebase. Having to worry about not-really-Routes could be complicating elsewhere.

I'm not sure why this would complicate Mutable?

bdeggleston: Should homeKey be removed from Route then? It seems odd that the route contains a homeKey some of the time, and other times it’s just an attribute on the command

belliottsmith: I don’t really follow? It’s possible to know the homeKey and not the Route, but it shouldn’t be possible to know the Route and not the homeKey

A Route is incomplete without a homeKey, as the whole concept of a Route is a homeKey+Unseekables, and the point of the Route is to declare where we expect both to be known. Again, they’re used throughout the codebase, so would be quite a disruptive change and potentially lead to mistakes with one being known and not the other.

bdeggleston: ~~is it possible for invalidated commands to have a homeKey, but no route?~~ Looks like it's possible

accord-core/src/main/java/accord/messages/ReadData.java
bdeggleston: It was before this and async-chain. I've removed it and put a check on removing command store ids from the waitingOn set

accord-core/src/main/java/accord/messages/Apply.java
bdeggleston: We do, yeah. `Write#apply` is called as part of this request. On the C* side we need the cfks to calculate the timestamps in `Write#apply` before submitting the mutations

accord-core/src/main/java/accord/local/CommonAttributes.java
bdeggleston: renamed to mutable

accord-core/src/main/java/accord/local/Command.java
bdeggleston: removed

accord-core/src/main/java/accord/local/Command.java
bdeggleston: So in that case, I’d say we should leave things as they are. AcceptedInvalidate and Invalidate use Accepted and Committed classes respectively, so they also need to support homeKey only and Route. It may make sense to introduce an additional Invalidated command class, though I’m not sure if it would cause problems in scenarios where an invalidation was accepted, then the command becomes committed after having forgotten it’s definition. Either way though, I think exploring that should be follow on work.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
bdeggleston: switched this to a nonNull check

accord-core/src/main/java/accord/local/Command.java
belliottsmith: Fair. I think probably we should consider removing the independent update of homeKey altogether, as I don’t think it was necessary. But agree it should be follow up work to clean this up further, it’s a minor thing.

accord-core/src/main/java/accord/messages/PreAccept.java
bdeggleston: Aesthetically, I think it _looks_ nicer this way. No real preference otherwise

dcapwell (commented):
forgot to submit...

accord-core/src/main/java/accord/utils/async/AsyncResults.java
```diff
@@ -0,0 +1,329 @@
...
+            Object current = state;
+            Invariants.checkState(current instanceof Result);
+            return (Result<V>) current;
+        }
+
+        public V result()
+        {
+            Result<V> result = getResult();
+            if (result.failure != null)
+                throw new IllegalStateException("Result failed " + result.failure);
```
dcapwell:
```suggestion
                throw new IllegalStateException("Result failed", result.failure);
```

so we don't loose the stack trace

accord-core/src/main/java/accord/messages/Commit.java
```diff
@@ -174,6 +178,11 @@ public ReadNack reduce(ReadNack r1, ReadNack r2)
     @Override
     public void accept(ReadNack reply, Throwable failure)
     {
+        if (failure != null)
+        {
+            logger.error("Unhandled exception during commit", failure);
```
dcapwell, line 183: should we not give this to the agent as it isn't handled?

```suggestion
            node.agent().onUncaughtException(failure); //TODO would be nice to provide "Unhandled exception during commit" for added context
```

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -207,26 +215,30 @@ else if (failure != null)
 
     private void ack()
     {
+        // wait for -1 because to ensure the setup phase has also completed. Setup calls ack in its callback
```
dcapwell:
```suggestion
        // wait for -1 to ensure the setup phase has also completed. Setup calls ack in its callback
```

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -207,26 +215,30 @@ else if (failure != null)
...
 
     private synchronized void readComplete(CommandStore commandStore, Data result)
     {
-        Preconditions.checkState(waitingOn.get(commandStore.id()));
+        Invariants.checkState(waitingOn.get(commandStore.id()));
         logger.trace("{}: read completed on {}", txnId, commandStore);
         if (result != null)
             data = data == null ? result : data.merge(result);
 
+        Invariants.checkState(waitingOn.get(commandStore.id()));
```
dcapwell: can we get a error msg?  if this fails in prod it can be hard to know was this pre or post `data.merge`, though I don't understand the case this happens.  We only remove in this method and we only mutate the BitSet around `synchronized` blocks, so its not clear how this condition could fail after the first was success

dcapwell (approved):
only nits remain, so +1

accord-core/src/main/java/accord/local/Command.java
belliottsmith: Could we add some todos to close this conversation out? 

`TODO (expected): register our listeners on-demand here, rather than en masse` to `NotifyWaitingOn`;
`TODO (expected): transient listeners should be maintained separately` to `Command.listeners`
`TODO (expected): listeners should be updated without an intervening mutable set` to `Command.listeners`

belliottsmith:
This has already been merged, I think? Closing for now.

PR #35, opened 2023-03-21
Add range assertions and docs

jacek-lewandowski:


accord-core/src/main/java/accord/primitives/AbstractRanges.java
```diff
@@ -41,6 +41,11 @@
 
     AbstractRanges(@Nonnull Range[] ranges)
     {
+        if (ranges.length > 0)
```
belliottsmith, line 44: We don't usually permit double nesting of no braces, the outer statement is usually braced (esp. when mixing if+for)

accord-core/src/main/java/accord/primitives/Range.java
```diff
@@ -366,4 +371,10 @@ public Range toUnseekable()
     {
         return this;
     }
+
+    void checkRangeType(Range that)
```
belliottsmith: final?

accord-core/src/main/java/accord/primitives/Range.java
```diff
@@ -289,9 +286,17 @@ public boolean intersects(AbstractKeys<?, ?> keys)
      */
     public Range intersection(Range that)
     {
+        checkRangeType(that);
+
         if (this.compareIntersecting(that) != 0)
             return null;
 
+        if (this.contains(that))
```
belliottsmith: Probably more efficient to check `start == this.start && end == this.end` below

accord-core/src/main/java/accord/primitives/Range.java
```diff
@@ -289,9 +286,17 @@ public boolean intersects(AbstractKeys<?, ?> keys)
...
+            return that;
+
+        if (that.contains(this))
```
belliottsmith: Probably more efficient to check start == that.start && end == that.end below

PR #36, opened 2023-03-21
CASSANDRA-18129 Accord <-> Paxos live migration

aweisberg:


accord-core/src/main/java/accord/coordinate/Barrier.java
```diff
@@ -0,0 +1,301 @@
...
+
+/**
+ * Local or global barriers that return a result once all transactions have their side effects visible.
+ *
+ * For local barriers the epoch is the only guarantee, but for global barriers a new transaction is created so it will
+ * be as of the timestamp of the created transaction.
+ *
+ * Note that reads might still order after, but side effect bearing transactions should not.
+ */
+public class Barrier extends AsyncResults.SettableResult<Timestamp>
```
bdeggleston: Could we make `AsyncResults.AbstractResult#trySetResult` protected and extend that class, so setResult/setFailure aren't visible to anyone interacting w/ Barrier?

accord-core/src/main/java/accord/coordinate/Barrier.java
```diff
@@ -0,0 +1,301 @@
...
+{
+    private static final Logger logger = LoggerFactory.getLogger(Barrier.class);
```
bdeggleston, line 63: unused

accord-core/src/main/java/accord/coordinate/Barrier.java
```diff
@@ -0,0 +1,301 @@
...
+     * For Applied we can return success immediately with the executeAt epoch. For PreApplied we can add
+     * a listener for when it transitions to Applied and then return success.
+     */
+    class ExistingTransactionCheck extends AsyncResults.SettableResult<BarrierTxn> implements MapReduceConsume<SafeCommandStore, BarrierTxn>
+    {
+        @Override
+        public BarrierTxn apply(SafeCommandStore safeStore)
+        {
+            // TODO Is this going to find the correct command store to listen for transaction progress?
+            BarrierTxn found = safeStore.mapReduceWithTerminate(
```
bdeggleston, line 255: there are a lot of TODOs still here indicating you're not confident in its correctness?

accord-core/src/main/java/accord/coordinate/Barrier.java
```diff
@@ -0,0 +1,301 @@
...
+                // TODO Should this lean harder into AsyncChain and not call begin?
+                safeStore.commandStore().execute(PreLoadContext.contextFor(found.txnId), safeStoreWithTxn -> safeStoreWithTxn.command(found.txnId).addListener(new BarrierCommandListener())).begin(node.agent());
+            }
+            return found;
+        }
+
+        @Override
+        public BarrierTxn reduce(BarrierTxn o1, BarrierTxn o2)
+        {
+            return o1 != null && o1.status.compareTo(o2.status) >= 0 ? o1 : o2;
```
bdeggleston: I it would be an error if multiple command stores owned the same key for the same epoch, especially if one was null. It would break it's barrier contract wouldn't it?

accord-core/src/main/java/accord/coordinate/Barrier.java
```diff
@@ -0,0 +1,301 @@
...
+            if (command.is(Status.Applied))
+            {
+                Timestamp executeAt = command.executeAt();
+                // In all the cases where we add a listener (listening to existing command, async completion of CoordinateSyncPoint)
+                // we want to notify the agent
+                doBarrierSuccess(executeAt);
+            }
+            else if (command.is(Status.Invalidated))
+                // TODO is this the right way to error out?
+                Barrier.this.tryFailure(new Timeout(command.txnId(), command.homeKey()));
```
bdeggleston, line 187: > TODO is this the right way to error out?

yes / good enough

accord-core/src/main/java/accord/coordinate/Barrier.java
```diff
@@ -0,0 +1,301 @@
...
+
+        }
+
+        @Override
+        public PreLoadContext listenerPreLoadContext(TxnId caller)
+        {
+            // TODO Surely the command passed to `onChange` is loaded?
+            return EMPTY_PRELOADCONTEXT;
```
bdeggleston, line 195: that's not guaranteed, it's possible that it was evicted in the time between the listener update being scheduled and being executed. Anyway, it needs to be in context or safeStore will complain

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Perform initial rounds of PreAccept and Accept until we have reached agreement about when we should execute.
+ * If we are preempted by a recovery coordinator, we abort and let them complete (and notify us about the execution result)
+ *
+ * TODO (desired, testing): dedicated burn test to validate outcomes
+ */
+abstract class CoordinatePreAccept<T> extends AsyncResults.SettableResult<T> implements Callback<PreAcceptReply>, BiConsumer<T, Throwable>
+{
+    private static final Logger logger = LoggerFactory.getLogger(CoordinatePreAccept.class);
```
bdeggleston, line 53: unused

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Perform initial rounds of PreAccept and Accept until we have reached agreement about when we should execute.
+ * If we are preempted by a recovery coordinator, we abort and let them complete (and notify us about the execution result)
+ *
+ * TODO (desired, testing): dedicated burn test to validate outcomes
+ */
+public class CoordinateSyncPoint extends CoordinatePreAccept<SyncPoint>
+{
+    private static final Logger logger = LoggerFactory.getLogger(CoordinateSyncPoint.class);
```
bdeggleston, line 53: unused

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
```diff
@@ -0,0 +1,135 @@
...
+            proposeAndCommitInvalidate(node, Ballot.ZERO, txnId, route.homeKey(), route, executeAt, this);
+        }
+        else
+        {
+            // SyncPoint transactions always propose their own txnId as their executeAt, as they are not really executed.
+            // They only create happens-after relationships wrt their dependencies, which represent all transactions
+            // that *may* execute before their txnId, so once these dependencies apply we can say that any action that
+            // awaits these dependencies applies after them. In the case of ExclusiveSyncPoint, we additionally guarantee
+            // that no lower TxnId can later apply.
+            new Propose<SyncPoint>(node, tracker.topologies(), Ballot.ZERO, txnId, txn, route, deps, txnId, this)
```
bdeggleston: can we add a static `Propose.propose` method instead of calling the Propose ctor directly?

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -210,56 +211,70 @@ private void recover()
...
+                    // TODO (required, consider): when writes/result are partially replicated, need to confirm we have quorum of these
+                    Persist.persistAndCommit(node, txnId, route, txn, executeAt, acceptedDeps, acceptOrCommit.writes, acceptOrCommit.result);
+                    accept(acceptOrCommit.result, null);
                     return;
+                }
 
                 case ReadyToExecute:
                 case PreCommitted:
                 case Committed:
+                {
```
bdeggleston, line 234: `node.withEpoch(executeAt.epoch(), () ->` has been removed from the various branches of the switch statement. I think this could lead to an exception when fetching topology for the executeAt epoch if we haven't received that epoch yet.

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -210,56 +211,70 @@ private void recover()
...
-                            else
-                            {
-                                // TODO (required, consider): when writes/result are partially replicated, need to confirm we have quorum of these
-                                Persist.persistAndCommit(node, txnId, route, txn, executeAt, deps, acceptOrCommit.writes, acceptOrCommit.result);
-                                accept(acceptOrCommit.result, null);
-                            }
-                        });
-                    });
+                {
+                    // must have gone through Accepted, so we must have witnessed >= Accepted for each shard
```
bdeggleston, line 222: I don't think this assumption holds. In the case of a fast path commit, it's possible to jump straight to commit/execute/apply, and if some shards have missed that message and others haven't, you'll have a high status that is accepted, but have missing keys in your accepted deps because some shards are still only preaccepted. Since sync points always run an accept phase, I think this would be safe for them, but you'd still need to run CollectDeps for normal txns

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -210,56 +211,70 @@ private void recover()
...
+                    accept(acceptOrCommit.result, null);
                     return;
+                }
 
                 case ReadyToExecute:
                 case PreCommitted:
                 case Committed:
+                {
+                    Deps acceptedDeps = tryMergeAcceptedDeps(); // must have gone through Accepted, so we must have witnessed >= Accepted for each shard
+                    Invariants.checkState(acceptedDeps != null);
```
bdeggleston, line 236: should add `mergeAcceptedDeps` for the places where we expect/require a full set to be assembled (Committed, Applied) and enforce the invariant there

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -359,6 +399,37 @@ public void completeOperation(SafeCommandStore store)
         }
     }
 
+    public void setRejectBefore(ReducingRangeMap<Timestamp> newRejectBefore)
+    {
+        this.rejectBefore = newRejectBefore;
+    }
+
+    @Override
+    public Timestamp preaccept(TxnId txnId, Seekables<?, ?> keys, SafeCommandStore safeStore)
```
bdeggleston, line 408: can some of this logic be moved into `CommandStore`? A lot of it seems easily generalizable and important for correctness.

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -49,6 +55,8 @@
  */
 public abstract class CommandStores<S extends CommandStore>
 {
+    private static final Logger logger = LoggerFactory.getLogger(CommandStores.class);
```
bdeggleston, line 59: unused

accord-core/src/main/java/accord/local/SafeCommandStore.java
```diff
@@ -76,13 +87,65 @@ default void addAndInvokeListener(TxnId txnId, TxnId listenerId)
         EXECUTES_AFTER
     }
     enum TestDep { WITH, WITHOUT, ANY_DEPS }
-    enum TestKind { Ws, RorWs }
+    enum TestKind implements Predicate<Kind>
+    {
+        Ws, RorWs, WsOrSyncPoint, SyncPoints, Any;
```
bdeggleston, line 92: With usage mostly contained in its own methods, let's expand Rs & Ws into Reads and Writes since it's slightly easier to read.

accord-core/src/main/java/accord/local/SafeCommandStore.java
```diff
@@ -76,13 +87,65 @@ default void addAndInvokeListener(TxnId txnId, TxnId listenerId)
...
+                default: throw new AssertionError();
+                case Any: return true;
+                case WsOrSyncPoint: return kind == Write || kind == Kind.SyncPoint || kind == ExclusiveSyncPoint;
+                case SyncPoints: return kind == Kind.SyncPoint || kind == ExclusiveSyncPoint;
+                case Ws: return kind == Write;
+                case RorWs: return kind == Read || kind == Write;
+            }
+        }
+
+        public static TestKind conflicts(Kind kind)
```
bdeggleston, line 107: correct usage of `conflicts` and `shouldHaveWitnessed` was confusing. Let's either improve method names or add comment (or both) indicating that `conflicts` determine which txns should be takes as deps for the given kind, and that `shouldHaveWitnessed` determines which types of txns should have taken the given kind as a dependency.

accord-core/src/main/java/accord/local/Status.java
```diff
@@ -422,4 +422,9 @@ public static Status max(Status a, Ballot acceptedA, Status b, Ballot acceptedB)
     {
         return max(a, a, acceptedA, b, b, acceptedB);
     }
+
+    public boolean isTerminal()
```
bdeggleston: unused

accord-core/src/main/java/accord/messages/BeginRecovery.java
```diff
@@ -128,10 +136,13 @@ public RecoverReply apply(SafeCommandStore safeStore)
             earlierAcceptedNoWitness = acceptedStartedBeforeWithoutWitnessing(safeStore, txnId, ranges, partialTxn.keys());
         }
 
+        Status status = command.status();
+        Ballot accepted = command.accepted();
+        Timestamp executeAt = command.executeAt();
+        PartialDeps acceptedDeps = status.phase.compareTo(Phase.Accept) >= 0 ? deps : PartialDeps.NONE;
```
bdeggleston, line 142: could we only include acceptedDeps for operations that propose deps?

accord-core/src/main/java/accord/messages/Commit.java
```diff
@@ -130,6 +143,24 @@ public static void commitMinimalAndRead(Node node, Topologies executeTopologies,
...
 
+    public static void commitMinimalAndBlockOnDeps(Node node, Topologies topologies, TxnId txnId, Txn txn, FullRoute<?> route, Deps deps, Callback<ExecuteReply> callback)
+    {
+        checkArgument(topologies.size() == 1);
+        Topology topology = topologies.get(0);
+        for (Node.Id to : topology.nodes())
+        {
+            // To simplify making sure the agent is notified once and is notified before the barrier coordination
+            // returns a result; we never notify the agent on the coordinator as part of WaitForDependenciesThenApply execution
+            boolean notifyAgent = to != node.id();
```
bdeggleston: should be `boolean notifyAgent = !to.equals(node.id());`

accord-core/src/main/java/accord/messages/PreAccept.java
```diff
@@ -35,12 +38,16 @@
...
 import static accord.local.SafeCommandStore.TestDep.ANY_DEPS;
+import static accord.local.SafeCommandStore.TestKind.Any;
 import static accord.local.SafeCommandStore.TestKind.RorWs;
 import static accord.local.SafeCommandStore.TestKind.Ws;
 import static accord.local.SafeCommandStore.TestTimestamp.STARTED_BEFORE;
+import static accord.primitives.Txn.Kind.ExclusiveSyncPoint;
 
 public class PreAccept extends WithUnsynced<PreAccept.PreAcceptReply>
 {
+    private static final Logger logger = LoggerFactory.getLogger(PreAccept.class);
```
bdeggleston, line 52: unused

accord-core/src/main/java/accord/messages/WaitForDependenciesThenApply.java
```diff
@@ -0,0 +1,114 @@
...
+        this.result = result;
+        this.notifyAgent = notifyAgent;
+    }
+
+    @Override
+    protected void readyToExecute(SafeCommandStore safeStore, Command.Committed command)
+    {
+        logger.trace("{}: executing WaitForDependenciesThenApply", command.txnId());
+        CommandStore unsafeStore = safeStore.commandStore();
+        if (notifyAgent)
```
bdeggleston, line 74: should this also check that `txnId.proposesDeps()`?

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -19,12 +19,22 @@
...
 public class Timestamp implements Comparable<Timestamp>
 {
+    private static final int REJECTED_FLAG = 0x8000;
+
+    /**
+     * The set of flags we want to retain as we merge timestamps (e.g. when taking mergeMax).
+     * Today this is only the REJECTED_FLAG, but we may include additional flags in future (such as Committed, Applied..)
+     * which we may also want to retain when merging in other contexts (such as in Deps).
+     */
+    private static final int MERGE_FLAGS = 0x8000;
```
bdeggleston, line 36: can you assign the value from `REJECTED_FLAG`? So in the future (with additional hypothetical flags) it would look like: `MERGED_FLAG = REJECTED_FLAG | ANOTHER_FLAG;`?

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -108,11 +118,38 @@ public int flags()
...
+    {
+        return withExtraFlags(REJECTED_FLAG);
+    }
+
     public Timestamp withEpochAtLeast(long minEpoch)
     {
         return minEpoch <= epoch() ? this : new Timestamp(minEpoch, hlc(), flags(), node);
     }
 
+    public Timestamp withExtraFlags(int flags)
```
bdeggleston: can be private

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -108,11 +118,38 @@ public int flags()
...
+    {
+        checkArgument(flags <= MAX_FLAGS);
+        long newLsb = lsb | flags;
+        if (lsb == newLsb)
+            return this;
+        return new Timestamp(msb, newLsb, node);
+    }
+
+    public Timestamp mergeFlags(Timestamp mergeFlags)
```
bdeggleston: can be private

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -108,11 +118,38 @@ public int flags()
         return flags(lsb);
     }
 
+    public boolean isRejected()
+    {
+        return (lsb & REJECTED_FLAG) != 0;
+    }
+
+    public Timestamp asRejected()
```
bdeggleston, line 126: can you add a comment noting this should only be used in `preaccept`

accord-core/src/main/java/accord/primitives/TxnId.java
```diff
@@ -120,7 +120,7 @@ private static Domain domain(int flags)
 
     private static int rwOrdinal(int flags)
     {
-        return (flags >> 1) & 1;
+        return (flags >> 1) & 3;
```
bdeggleston, line 123: can you replace `3` with a mask constant?

accord-core/src/main/java/accord/primitives/TxnId.java
```diff
@@ -120,7 +120,7 @@ private static Domain domain(int flags)
 
     private static int rwOrdinal(int flags)
```
bdeggleston, line 121: can you rename this (and `rw`) to `kindOrdinal` and `kind`, respectively?

accord-core/src/main/java/accord/topology/Topologies.java
```diff
@@ -18,20 +18,28 @@
...
-import com.google.common.collect.Sets;
 import accord.utils.Invariants;
 
-import java.util.*;
-
 // TODO (desired, efficiency/clarity): since Topologies are rarely needed, should optimise API for single topology case
 //  (e.g. at least implementing Topologies by Topology)
 public interface Topologies extends TopologySorter
 {
+    Logger logger = LoggerFactory.getLogger(Topologies.class);
```
bdeggleston, line 41: unused

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -52,6 +60,8 @@
  */
 public class TopologyManager implements ConfigurationService.Listener
 {
+    private static final Logger logger = LoggerFactory.getLogger(TopologyManager.class);
```
bdeggleston, line 64: unused

accord-core/src/main/java/accord/utils/Functions.java
```diff
@@ -57,4 +57,12 @@ public static <I, O> O mapReduceNonNull(Function<I, O> map, BiFunction<O, O, O>
         return result;
     }
 
+    public static <I, O> O foldl(List<I> list, BiFunction<I, O, O> foldl, O zero)
```
bdeggleston, line 60: could you rename `zero` to `identity` or `initial` or something, since this isn't strictly numerical?

bdeggleston (changes requested):

accord-core/src/main/java/accord/coordinate/Barrier.java
aweisberg: I left those in because it's definitely an area during review that needs to be called out and looked at closely. High ROI in terms of bringing in more experience.

I don't think it returns a wrong answer, but it might be pessimistic about what constitutes a completed transaction or a transaction that could be listened to for completion.

I think the choice of `TestKind.Any` needs to be thought through as well because `Any` could mean something different later, and `ExclusiveSync` might not actually be sufficient? @belliottsmith

accord-core/src/main/java/accord/coordinate/Barrier.java
aweisberg: That is true, but I kind of expected this to get expanded later to support multiple keys and ranges. The only reason I restricted it was that it was all that I needed and it created a smaller surface area for problems.

It simplified finding the correct command store to listen to. That is why it's restricted to `key`.

Happy to have this generate an error.

accord-core/src/main/java/accord/coordinate/Recover.java
aweisberg: This is more `exclsync` changes https://github.com/belliottsmith/cassandra-accord/compare/ffd3664ab0e40d5aadc218a6c3f26caa09161b59...exclsync#diff-d169314280a3ade4500e27f6595a3fcbef42b4a9f1cf3075b5059f47083305bd...e79c7341592244a7c13ef603700639731a030258
@belliottsmith  why did that end up getting removed in the exclsync branch?

I am also not clear on what ensures the `executeAt` epoch is known when calling things like `Persist.persistAndCommit` and `Execute.execute`

accord-core/src/main/java/accord/messages/WaitForDependenciesThenApply.java
aweisberg: I think it would have to be even more specific because `ExclusiveSyncPoint`  apply's asynchronously and doesn't wait on dependencies so it wouldn't qualify. Nor should it reach here.

I could just check for `Txn.Kind` is `SyncPoint`?

accord-core/src/main/java/accord/coordinate/Barrier.java
aweisberg: I did that and also put `trySuccess` and `tryFailure` in `AbstractResult`. `SettableResult` now forward to those since it is trying to make them public for the `Settable` interface.

accord-core/src/main/java/accord/coordinate/Barrier.java
aweisberg: Let's leave this unresolved for now. I removed some of them, but I still want to go over it with Benedict and I am pretty sure that `TestKind.Any` doesn't work because an exclusive sync doesn't wait for dependencies to apply.

accord-core/src/main/java/accord/coordinate/Barrier.java
aweisberg: It generates an error now.

accord-core/src/main/java/accord/coordinate/Barrier.java
aweisberg: I forgot about this TODO. I think I do want to have retries from the repair process to allow for these to occasionally be invalidated and not throw away the whole repair. I updated the TODO where the barrier invocation is.

accord-core/src/main/java/accord/messages/Commit.java
aweisberg: Fixed TY!

PR #37, opened 2023-03-23
add shareable APPLIED and INVALIDATED implementations of Result

maedhroz:
patch by Caleb Rackliffe; reviewed by ? for CASSANDRA-18355

### DIFF
--- a/accord-core/src/main/java/accord/api/Write.java
+++ b/accord-core/src/main/java/accord/api/Write.java
@@ -19,6 +19,7 @@
 package accord.api;
 
 import accord.local.SafeCommandStore;
+import accord.primitives.PartialTxn;
 import accord.primitives.Seekable;
 import accord.primitives.Timestamp;
 import accord.utils.async.AsyncChain;
@@ -30,5 +31,5 @@ import accord.utils.async.AsyncChain;
  */
 public interface Write
 {
-    AsyncChain<Void> apply(Seekable key, SafeCommandStore safeStore, Timestamp executeAt, DataStore store);
+    AsyncChain<Void> apply(Seekable key, SafeCommandStore safeStore, Timestamp executeAt, DataStore store, PartialTxn txn);
 }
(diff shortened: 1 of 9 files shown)


maedhroz:
https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=CASSANDRA-18355

accord-core/src/main/java/accord/api/Write.java
```diff
@@ -30,5 +31,5 @@
  */
 public interface Write
 {
-    AsyncChain<Void> apply(Seekable key, SafeCommandStore safeStore, Timestamp executeAt, DataStore store);
+    AsyncChain<Void> apply(Seekable key, SafeCommandStore safeStore, Timestamp executeAt, DataStore store, Command.Executed command);
```
belliottsmith: Do we need to provide the whole command, or could we provide just the `PartialTxn` or `Update`?

dcapwell: I kinda don't like exposing `Command` in the API and kinda prefer `TxnId`... you do the following in C*

```
      // Pull updates for this key that did not require completion form the transaction itself and apply them: 
            TxnUpdate txnUpdate = (TxnUpdate) command.partialTxn().update();
            assert txnUpdate != null : "PartialTxn should contain an update if we're applying a write!";
            List<Update> updates = txnUpdate.completeUpdatesForKey((RoutableKey) key);
            updates.forEach(update -> results.add(update.write(timestamp, nowInSeconds)));
```

so you will need it, but I just kinda don't like it...

maedhroz: @belliottsmith above suggests using just the `PartialTxn` or the `Update`. You're suggesting perhaps passing `TxnId` and getting the command from `SafeCommandStore` (along w/ the key, which I think we also have available)?

aweisberg (approved):

dcapwell (approved):

PR #38, opened 2023-03-24
CASSANDRA-18364: CEP-15: (C*) Accord message processing should avoid being passed on to a Stage and run directly in the messageing handler

dcapwell:


accord-core/src/main/java/accord/messages/CallbackHolder.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package accord.messages;
+
+import accord.coordinate.Timeout;
+import accord.local.CommandStore;
+import accord.local.Node;
+import accord.local.PreLoadContext;
+
+public class CallbackHolder<T extends Reply>
```
dcapwell: open to better names...

accord-core/src/main/java/accord/coordinate/CheckShards.java
```diff
@@ -61,7 +61,8 @@ private static Topologies topologyFor(Node node, TxnId txnId, Unseekables<?, ?>
     protected void contact(Id id)
     {
         Unseekables<?, ?> unseekables = contact.slice(topologies().computeRangesForNode(id));
-        node.send(id, new CheckStatus(txnId, unseekables, txnId.epoch(), untilRemoteEpoch, includeInfo), this);
+        node.send(id, new CheckStatus(txnId, unseekables, txnId.epoch(), untilRemoteEpoch, includeInfo),
+                  node.commandStores().from(unseekables), this);
```
belliottsmith: Shouldn't we already know the `CommandStore` - it should be the one we're executing on ourselves?

Perhaps we should introduce a static method for discovering the `CommandStore` we're executing on currently, and store it in a `ThreadLocal` (or `FastThreadLocal`)

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -172,7 +176,8 @@ public static Recover recover(Node node, Ballot ballot, TxnId txnId, Txn txn, Fu
 
     void start(Set<Id> nodes)
     {
-        node.send(nodes, to -> new BeginRecovery(to, tracker.topologies(), txnId, txn, route, ballot), this);
+        node.send(nodes, to -> new BeginRecovery(to, tracker.topologies(), txnId, txn, route, ballot),
```
belliottsmith: does this need to wrap? doesn't provide much greater clarity

accord-core/src/main/java/accord/messages/CallbackHolder.java
belliottsmith: `SafeCallback`, `CallbackInvoker`, `IndirectCallback`, `CallbackContext`?

It might also be nice to make `CommandStore` extend `Executor` so that we can support a cheap `execute` and in this case we can also call this e.g. `CallbackWithExecutor` (but this would be a nice change, even if we don't like this name)

belliottsmith (changes requested):
This looks broadly what I had in mind, only that we should not be selecting a `CommandStore` for each send, as this doesn't solve concurrency issues with the issuing thread that could itself interact with state in the callback. We should ordinarily already be executing on a `CommandStore`, and if there are cases where that isn't true today we should make it true, and then we should have a mechanism for asking for the `CommandStore` we're currently executing on.

accord-core/src/main/java/accord/coordinate/CheckShards.java
dcapwell: > Perhaps we should introduce a static method for discovering the CommandStore we're executing on currently, and store it in a ThreadLocal (or FastThreadLocal)

I was thinking the same.  We don't depend on netty so don't think I can use FastThreadLocal though

accord-core/src/main/java/accord/coordinate/Recover.java
dcapwell: I did this for line length and consistency reasons.  I don't see an override in https://cassandra.apache.org/_/development/code_style.html, so our style guide should default to 80 chars (see https://www.oracle.com/java/technologies/javase/codeconventions-indentation.html#313)

at char 80 is `, txn, route, ballot), commandStore, this`, so I made a personal style choice to keep the whole `BeginRecovery` on the same line

dcapwell:
> and if there are cases where that isn't true today we should make it true

it should be true... easy to find out I guess (burn test + cluster benchmark)

dcapwell:
Working on POC to show where we *need* to provide vs use the CURRENT; the below are the paths BurnTest found


* accord.coordinate.Coordinate#start
* accord.utils.MessageTask#run from accord.topology.TopologyRandomizer#maybeUpdateTopology
* accord.coordinate.CheckShards#contact from accord.coordinate.ReadCoordinator#start() from accord.impl.list.ListRequest.ResultCallback#accept
* accord.coordinate.Propose#propose(accord.local.Node, accord.topology.Topologies, accord.primitives.Ballot, accord.primitives.TxnId, accord.primitives.Txn, accord.primitives.FullRoute<?>, accord.primitives.Timestamp, accord.primitives.Deps, java.util.function.BiConsumer<accord.api.Result,java.lang.Throwable>)

Updated to handle just these cases and BurnTest is stable... going to flesh out and push soon

accord-core/src/main/java/accord/messages/CallbackHolder.java
dcapwell: I like `Context` a lot but `SafeCallback` is more consistent with all the other `Safe*` interfaces that imply the work is in-store...

accord-core/src/test/java/accord/utils/MessageTask.java
```diff
@@ -70,6 +71,7 @@ public String toString()
     private final List<Node.Id> recipients;
     private final String desc;
     private final Request request;
+    private final CommandStore commandStore;
```
dcapwell: need to debug this class as it looks like its crossing stores often... ATM I try to avoid that as I added a check to block that case (this is the only section that hit it)

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -398,6 +397,14 @@ protected <T> void executeInContext(InMemoryCommandStore commandStore, PreLoadCo
         }
     }
 
+    @Override
+    public String toString()
```
dcapwell, line 388: when you run in the wrong store we log the stores involved, so need this.  I felt that only `id` is useful so left everything else out.

accord-core/src/main/java/accord/coordinate/Recover.java
belliottsmith: "Try to keep lines under 120 characters, but use good judgment. It is better to exceed this limit, than to split a line that has no natural splitting points, particularly when the remainder of the line is boilerplate or easily inferred by the reader."

accord-core/src/test/java/accord/burn/TopologyUpdates.java
```diff
@@ -154,7 +153,8 @@ public static <T> BiConsumer<T, Throwable> dieOnException()
         return (result, throwable) -> {
             if (throwable != null)
             {
-                logger.error("", throwable);
+                logger.error("Unexpected exception", throwable);
+                logger.error("", new Throwable("Shutting down test"));
```
dcapwell, line 158: hit this once and was useful to know where shutdown was happening and not just what triggered it

accord-core/src/main/java/accord/coordinate/Recover.java
dcapwell: thanks, missed that

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -399,7 +399,10 @@ public boolean hasEpoch(long epoch)
 
     public Topology localForEpoch(long epoch)
     {
-        return epochs.get(epoch).local();
+        EpochState epochState = epochs.get(epoch);
+        if (epochState == null)
+            throw new IllegalStateException("Unknown epoch " + epoch);
+        return epochState.local();
```
dcapwell, line 412: here is the exception I am changing this to add more details

```
java.lang.NullPointerException 
	at accord.topology.TopologyManager.localForEpoch(TopologyManager.java:402) 
	at accord.local.Node.trySelectProgressKey(Node.java:467) 
	at accord.messages.TxnRequest.progressKey(TxnRequest.java:151) 
	at accord.messages.TxnRequest.process(TxnRequest.java:143) 
	at org.apache.cassandra.service.accord.AccordVerbHandler.doVerb(AccordVerbHandler.java:46) 
	at org.apache.cassandra.net.InboundSink.lambda$new$0(InboundSink.java:78)  
	at org.apache.cassandra.net.InboundSink.accept(InboundSink.java:97)  
	at org.apache.cassandra.net.InboundSink.accept(InboundSink.java:45)  
	at org.apache.cassandra.net.InboundMessageHandler$ProcessMessage.run(InboundMessageHandler.java:430) 
	at org.apache.cassandra.concurrent.ImmediateExecutor.execute(ImmediateExecutor.java:87)  
	at org.apache.cassandra.concurrent.Stage.execute(Stage.java:125) 
	at org.apache.cassandra.net.InboundMessageHandler.dispatch(InboundMessageHandler.java:399)
```

accord-core/src/main/java/accord/api/MessageSink.java
```diff
@@ -27,6 +28,6 @@
 public interface MessageSink
 {
     void send(Id to, Request request);
-    void send(Id to, Request request, Callback callback);
+    void send(Id to, Request request, CommandStore commandStore, Callback callback);
```
bdeggleston: Should `CommandStore` be added to the callback interface?

dcapwell: I felt that this way was cleaner for the following reason: callbacks do not normally pick a CommandStore, so adding there would require us to plumb this through

accord-core/src/main/java/accord/messages/Commit.java
```diff
@@ -162,7 +162,8 @@ public synchronized ReadNack apply(SafeCommandStore safeStore)
             case Insufficient:
                 SafeCommand safeCommand = safeStore.command(txnId);
                 Invariants.checkState(!safeCommand.current().known().isDefinitionKnown());
-                if (defer == null)
+                // When isDone=true this.process() is called and all CommandStores run apply again; so safe to create a new one
+                if (defer == null || defer.isDone)
```
aweisberg: I need to look closer, but why would we need to defer multiple times, shouldn't defer wait until the read is possible?

Is this a case where defer is created, and completed before all command stores were added to the defer?

accord-core/src/main/java/accord/coordinate/Persist.java
```diff
@@ -94,13 +94,13 @@ public void onSuccess(Id from, ApplyReply reply)
                     {
                         // TODO (low priority, consider, efficiency): send to non-home replicas also, so they may clear their log more easily?
                         Shard homeShard = node.topology().forEpochIfKnown(route.homeKey(), txnId.epoch());
-                        node.send(homeShard, new InformHomeDurable(txnId, route.homeKey(), executeAt, Durable, persistedOn));
+                        node.send(homeShard, new InformHomeDurable(txnId, route.homeKey(), executeAt, Durable, new HashSet<>(persistedOn)));
```
aweisberg, line 97: I get this fixes a potential concurrent modification, but on an unrelated note I am a little surprised to notice that we send an `InformHomeDurable` from every single `Persist`. Seems like in most cases it's already the original coordinator running this and it's just sending a message to itself. I guess that is basically free and it works for recovery coordinators as well?

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -416,6 +422,35 @@ public synchronized void shutdown()
...
+        for (ShardHolder holder : shards)
+        {
+            if (fn.test(holder.ranges().currentRanges()))
+                return holder.store;
+        }
+        return any();
+    }
+
+    @VisibleForTesting
+    public CommandStore any()
```
aweisberg, line 440: I low key wish that this method didn't have to exist because. Even having the lookup by key/range/routable is somewhat undesirable because in so many cases you already know what the correct command store and should use that.

What are the cases where we should be picking a random one vs just continuing to operate in the Netty thread?

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -203,7 +203,7 @@ public void withEpoch(long epoch, Runnable runnable)
         else
         {
             configService.fetchTopologyForEpoch(epoch);
-            topology.awaitEpoch(epoch).addCallback(runnable);
+            topology.awaitEpoch(epoch).addCallback(runnable).begin(agent());
```
aweisberg, line 206: Yikes, so without this fix `runnable` never runs?

dcapwell, line 206: I changed the type returned from `awaitEpoch` to be `AsyncChain`, so need to "begin" it.

The reason is that `AsyncResult` doesn't know what `CommandStore` to run in, so if you are in a `CommandStore` you need to start a new chain, so something has to `begin` it

accord-core/src/main/java/accord/api/MessageSink.java
aweisberg: Maybe make the storage for `Callback` intrusive so it doesn't need to be a separate allocation?

It would take a little jiggering for it to be a base class since it is currently an interface.

I don't feel strongly about it.

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -259,9 +260,14 @@ public synchronized void onTopologyUpdate(Topology topology)
             toComplete.trySuccess(null);
     }
 
-    public synchronized AsyncResult<Void> awaitEpoch(long epoch)
+    public synchronized AsyncChain<Void> awaitEpoch(long epoch)
     {
-        return epochs.awaitEpoch(epoch);
+        AsyncResult<Void> result = epochs.awaitEpoch(epoch);
```
aweisberg: This doesn't preserve the immediate success behavior, of `awaitEpoch`, but I don't think it matters because all the callers are first checking if the `Epoch` is already present.

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -331,6 +331,62 @@ public static <V> AsyncChain<V> failure(Throwable failure)
         return new Immediate<>(failure);
     }
 
+    public static <V, T> AsyncChain<T> map(AsyncChain<V> chain, Function<? super V, ? extends T> mapper, Executor executor)
+    {
+        return chain.flatMap(v -> new Head<T>()
```
aweisberg, line 336: This calls `flatMap` not map?

aweisberg, line 336: This calls `flatMap` not `map`?

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -331,6 +331,62 @@ public static <V> AsyncChain<V> failure(Throwable failure)
...
+                        }
+                        catch (Throwable t)
+                        {
+                            callback.accept(null, t);
+                        }
+                    });
+                }
+                catch (Throwable t)
+                {
+                    callback.accept(null, t);
```
aweisberg, line 389: This doesn't run in the expected executor, maybe it should go to the agent? And same above.

Could be surprising cleanup code doesn't run in the executor requested.

accord-core/src/test/java/accord/burn/BurnTestConfigurationService.java
```diff
@@ -269,10 +272,15 @@ public synchronized void acknowledgeEpoch(long epoch)
...
         epochs.acknowledge(epoch);
         Topology topology = getTopologyForEpoch(epoch);
-        Node originator = lookup.apply(node);
+        Node originator = originator();
         topologyUpdates.syncEpoch(originator, epoch - 1, topology.nodes());
     }
 
+    private Node originator()
+    {
+        return lookup.apply(node);
```
aweisberg, line 278: `MessageSink` is unused now

accord-core/src/test/java/accord/burn/TopologyUpdates.java
```diff
@@ -142,7 +140,7 @@ else if (outcome == Nothing)
                         dieExceptionally(invalidate.addCallback(((unused, failure) -> onDone.accept(failure == null))).beginAsResult());
                 }
                 return null;
-            }).beginAsResult();
+            }, node.commandStores().any()).beginAsResult();
```
aweisberg, line 144: Why is a random command store acceptable/necessary here? Because the contract for send eventually requires there to be a command store?

aweisberg (changes requested):
I looked a bit to see how many things we run from `Request.process` that looks risky in terms of running from a Netty thread. The danger is that we block on some async request that actually needs the Netty thread we are blocking to run.

Most things look totally reasonable in that they immediately go to map reduce over local nodes with a callback.

accord-core/src/main/java/accord/api/MessageSink.java
dcapwell: after running `BurnTest` for a long time I found I actually have to use `Executor` as some things are run in `CommandStore` only because I forced them to with this change... stuff like our test epoch population logic shouldn't be running in a CommandStore

> Maybe make the storage for Callback intrusive so it doesn't need to be a separate allocation?
> It would take a little jiggering for it to be a base class since it is currently an interface.

I don't follow this comment, what do you mean?

accord-core/src/main/java/accord/messages/Commit.java
dcapwell: this was reverted.  I fixed this bug on trunk by not relying on how many counts were seen without taking into account the `accept` callback in the MapReduce phase.  This patch would fix things but would have a double call to `TxnRequest.process`, the fix on trunk solves that.

accord-core/src/main/java/accord/local/CommandStores.java
dcapwell: I think I can get rid of this.  This used to play a bigger role but after a few refactors its now only used as the fallback case with 2 test cases using... I will look to see if I can move the test usage off `CommandStore` and into the new global executor for `BurnTest`...  that would allow this method to become private

> Even having the lookup by key/range/routable is somewhat undesirable because in so many cases you already 
know what the correct command store and should use that.

that's for the case where you are not running in a CommandStore, such as `accord.coordinate.Coordinate#start` which runs in C* space (executed by `TransactionStatement`)

accord-core/src/main/java/accord/local/Node.java
dcapwell: this wasn't a bug before, its due to a change in this patch.

`AsyncResult` is just like a `Future`, so `begin` noops...  `AsyncChain` needs a `begin` to start the chain (we delay execution of the chain until the whole chain is known; `begin` marks the completion of the chain).

When you `awaitEpoch` you need to know if the caller is in a `CommandStore` or not, and if in a `CommandStore` you need to start a new chain where all logic will run in the `CommandStore` and not w/e thread populated the `epoch`... This means there are 2 different returns: `AsyncResult` (future that will eventually have the epoch) and `AsyncChain` (`result.map(a -> a, CommandStore.current())`)

accord-core/src/main/java/accord/topology/TopologyManager.java
dcapwell: can always optimize and check that... LHF =)

accord-core/src/main/java/accord/utils/async/AsyncChains.java
dcapwell: yes, because I need to create a `AsyncChain` where the result is the result of executing the function in the `Executor`, so `flatMap` is needed.

dcapwell: yes, because I need to create a `AsyncChain` where the result is the result of executing the function in the `Executor`, so `flatMap` is needed.

accord-core/src/main/java/accord/utils/async/AsyncChains.java
dcapwell: `AsyncChains` doesn't know about `Agent`... so this is a hard edge case... if the `Executor` rejects then we either YOLO log or we give the `callback`, which to your point may not be thread safe in this case

Ill think about this

accord-core/src/test/java/accord/burn/BurnTestConfigurationService.java
dcapwell: I don't understand this comment.. this function is used, and not sure the relation with `MessageSink`?

accord-core/src/test/java/accord/burn/TopologyUpdates.java
dcapwell: the API demanded it, but I had to change that due to epoch sync, so ill look to remove this

dcapwell: sadly I wasn't able to remove, `accord.coordinate.FetchData#fetch` has to run in a `CommandStore` and there isn't a clear one to pick atm; `CommandStore.current()` fails as this is done outside a store, so I need to select a random store...

accord-core/src/main/java/accord/local/CommandStores.java
dcapwell: I tried getting rid of the 2 other test usage and both need this... `MessageTask` looks fine to get rid of (testing now), but `accord.burn.TopologyUpdates.CommandSync#process` needs to run inside a `CommandStore` and there isn't anything really to select which one... so need a random one...

accord-core/src/main/java/accord/api/MessageSink.java
aweisberg: `SafeCallback` just stores the `CommandStore` that `Callback` will need. So why not put the storage in `Callback` and make it an abstract base class. Then you save an allocation for `SafeCallback`.

accord-core/src/test/java/accord/burn/BurnTestConfigurationService.java
aweisberg: `MessageSink` is unused because replaced it with `originator()` and calling `Node.send`[ on line 236](https://github.com/apache/cassandra-accord/pull/38/files/bff2a2b6b1aebbc871be4d0d34ba04acf19454a6#diff-8e19f7549abc451cea111cebf36df3a0bdc0c8b42c4f61193cdf05c7c80fdcdcR236)

accord-core/src/main/java/accord/api/MessageSink.java
dcapwell: > So why not put the storage in Callback and make it an abstract base class.

This has an annoying impact to all code, as all code must now figure out what store to run in rather than the small set that need to do this today.  ATM this is inferred in the majority of cases, so only a few cases need to think about this.

aweisberg: I am only suggesting putting the storage in the base class. Setting the value of the `CommandStore` would still be done the same way.

dcapwell: @aweisberg and I spoke in slack, here is the jist of the convo...

We have `AccordCallback` in C* that translates C* callbacks to Accord callback
We have 3 different networking implementations in accord testing (maelstrom, MockCluster, BurnTest)

This means we have 4 different places that need to implement callbacks correctly, so need to centralize this logic to avoid copy/paste duplication; this is what `SafeCallback` does.


Now, we "could" move the `Executor` pointer to `Callback` and make this `abstract` (which @aweisberg was proposing) but this doesn't lower memory costs, and now adds mutable state into the system.

We could also have a `executor()` method, but would need to update all implementations... at the money this is figured out when you try to send a message, so this inference is centralized; I feel there is more value to centralize the inference than having each callback try

aweisberg: I didn't realize that you always have to allocate a wrapper for `MessagingService' to adapt to the different callback interface. So there is nothing to be gained by moving stuff into `Callback`.

accord-core/src/main/java/accord/utils/async/AsyncChains.java
aweisberg: I noticed this pattern in a couple of places so we don't have to resolve it right now, but we should settle on a desired outcome. Either YOLO run the callback or it's better to just not run it and log. Liveness vs correctness tradeoff, but also it errored out in a very unexpected way so how much value is there in running the callback?

accord-core/src/main/java/accord/utils/async/AsyncChain.java
```diff
@@ -63,6 +63,13 @@ default AsyncChain<Void> accept(Consumer<? super V> action, Executor executor)
         }, executor);
     }
 
+    default AsyncChain<V> withExecutor(Executor e)
```
aweisberg, line 66: Random musings.

The abstraction async chain has where maybe you specify an executor and maybe you don't seems somewhat of an anti-pattern. I do think Guava got one thing right with `ListenableFuture` which is that specifying where your callback will execute is never optional.

But it's also pretty much implicit that it defaults to a direct execution. Still it's something of a hint to have to manually specify "direct execution", but also people will do it automatically a lot not knowing which thread it might execute on.

accord-core/src/test/java/accord/burn/BurnTestConfigurationService.java
```diff
@@ -130,11 +129,10 @@ EpochHistory syncComplete(long epoch)
         }
     }
 
-    public BurnTestConfigurationService(Node.Id node, Executor executor, MessageSink messageSink, Supplier<RandomSource> randomSupplier, Topology topology, Function<Node.Id, Node> lookup, TopologyUpdates topologyUpdates)
+    public BurnTestConfigurationService(Node.Id node, Executor executor, Supplier<RandomSource> randomSupplier, Topology topology, Function<Node.Id, Node> lookup, TopologyUpdates topologyUpdates)
```
aweisberg: Now there is an unused import style error

accord-core/src/test/java/accord/utils/MessageTask.java
```diff
@@ -119,7 +118,7 @@ public void onSuccess(Node.Id from, Reply reply)
             Invariants.checkArgument(reply == SUCCESS || reply == FAILURE);
             if (reply == FAILURE)
             {
-                originator.send(from, request, commandStore, this);
+                originator.send(from, request, MoreExecutors.directExecutor(), this);
```
aweisberg: I don't 100% follow why some of the time we are OK with direct executor, but other times we want to pick a random command store.

If we have things that don't neatly fit in a command store would it make sense to have a small thread pool just for those?

One of the things I was thinking about was observability and how that could tie into the statistics we expose for thread pools and stages. We currently don't expose Accord as a stage.

Maybe we will head in a different direction, but it seems nice to be able to pull both Accord and other stages stats from the same place.

dcapwell: This is a testing class that is used for burn test to change epoch over the run.  I could change this to run in the new global executor that the burn test creates, but we are already running there, so went with direct just to simplify the diff.


I pushed a patch that uses the burn test executor, and in the tracker test to use the same delay simulation executor

aweisberg (approved):
+1 TY!

accord-core/src/main/java/accord/utils/async/AsyncChain.java
dcapwell: the common case is that your "chain" is just a sequence of callbacks on the same thread, so we only ever need to define the executor at the start of the chain rather than maintain it for each step in the chain like you do with `ListenableFuture`, but you are right, not thinking about executors can be a problem and was one of the biggest issues getting this patch to work... but if people just put `direct` you kinda shot yourself in the foot just to compile =)

accord-core/src/main/java/accord/coordinate/Persist.java
belliottsmith: It sends this to every other replica in the home shard also.

accord-core/src/main/java/accord/utils/async/AsyncChains.java
belliottsmith: This is a good point. I think we can probably fix this by using our own Executor interface on which you can submit mutually-exclusive executions in case the executor were providing mutual exclusion. i.e., even if the executor is shutdown, it may permit waiting until it is terminated before taking a mutex and then invoking some continuation.

But let's tag that as a low priority follow-up ticket.

accord-core/src/main/java/accord/coordinate/Coordinate.java
```diff
@@ -72,7 +72,8 @@ private void start()
     {
         // TODO (desired, efficiency): consider sending only to electorate of most recent topology (as only these PreAccept votes matter)
         // note that we must send to all replicas of old topology, as electorate may not be reachable
-        node.send(tracker.nodes(), to -> new PreAccept(to, tracker.topologies(), txnId, txn, route), this);
+        node.send(tracker.nodes(), to -> new PreAccept(to, tracker.topologies(), txnId, txn, route),
+                  node.commandStores().from(route), this);
```
belliottsmith: memoize? no need to recompute.

also, should probably use home key by default to pick?

dcapwell: > memoize? no need to recompute.

`node.commandStores().from(route)` is called outside of the loop, so it is computed once.  Or do you mean do this in the constructor in case `start` is called multiple times?

belliottsmith: oh, my mistake. it looked like it was called within the continuation

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -43,7 +48,88 @@ CommandStore create(int id,
                             RangesForEpochHolder rangesForEpoch);
     }
 
+    @VisibleForTesting
+    class Unsafe
```
belliottsmith: Should this be in the InMemoryCommandStore, maybe, since it's an implementation detail?

accord-core/src/main/java/accord/coordinate/Coordinate.java
dcapwell: > also, should probably use home key by default to pick?

Pushed the change and removed one of the `from` methods as it was only used by `Routes`

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -416,6 +422,35 @@ public synchronized void shutdown()
             shard.store.shutdown();
     }
 
+    public CommandStore from(Routables<?, ?> routables)
```
belliottsmith: select?

accord-core/src/main/java/accord/messages/SafeCallback.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public void timeout(Node.Id to)
+    {
+        failure(to, new Timeout(null, null));
+    }
+
+    private void safeCall(Node.Id src, Runnable fn)
+    {
+        executor.execute(() -> {
```
belliottsmith, line 66: leave a TODO to handle `RejectedExecutionException`?

accord-core/src/main/java/accord/messages/SafeCallback.java
```diff
@@ -0,0 +1,87 @@
...
+            }
+            catch (Throwable t)
+            {
+                try
+                {
+                    callback.onCallbackFailure(src, t);
+                }
+                finally
+                {
+                    agent.onUncaughtException(t);
```
belliottsmith: There's a good chance the callback implementation invokes this, so maybe only invoke in case of another exception?

accord-core/src/main/java/accord/messages/SafeCallback.java
```diff
@@ -0,0 +1,87 @@
...
+    public SafeCallback(Executor executor, Agent agent, Callback<T> callback)
+    {
+        this.executor = Objects.requireNonNull(executor, "executor");
+        this.agent = Objects.requireNonNull(agent, "agent");
+        this.callback = Objects.requireNonNull(callback, "callback");
+    }
+
+    public void success(Node.Id src, T reply)
+    {
+        safeCall(src, () -> callback.onSuccess(src, reply));
```
belliottsmith: I don't love the double wrapping of capturing continuations here. Perhaps something like:

```
    public void success(Node.Id src, T reply)
    {
        safeCall(src, reply, Callback::onSuccess);
    }

    public void slowResponse(Node.Id src)
    {
        safeCall(src, null, (id, ignore) -> callback.onSlowResponse(id));
    }

    public void failure(Node.Id to, Throwable t)
    {
        safeCall(to, t, Callback::onFailure);
    }

    interface SafeCall<T, P>
    {
        void accept(Callback<T> callback, Node.Id id, P param) throws Throwable;
    }

    private <P> void safeCall(Node.Id src, P param, SafeCall<T, P> call)
    {
        executor.execute(() -> {
            try
            {
                call.accept(callback, src, param);
            }
            catch (Throwable t)
            {
                try
                {
                    callback.onCallbackFailure(src, t);
                }
                catch (Throwable t2)
                {
                    t.addSuppressed(t2);
                    agent.onUncaughtException(t);
                }
            }
        });

    }
```

accord-core/src/main/java/accord/messages/SafeCallback.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.concurrent.Executor;
+
+import accord.api.Agent;
+import accord.coordinate.Timeout;
+import accord.local.Node;
+
+public class SafeCallback<T extends Reply>
+{
+    private final Executor executor;
+    private final Agent agent;
```
belliottsmith: Maybe we can extract a common parent interface of `CommandStore` with an `agent` method?

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -65,19 +67,23 @@ CommandStores<?> create(NodeTimeService time,
...
+            this.random = random;
             this.progressLogFactory = progressLogFactory;
             this.shardFactory = shardFactory;
         }
 
         CommandStore create(int id, RangesForEpochHolder rangesForEpoch)
         {
-            return shardFactory.create(id, time, agent, store, progressLogFactory, rangesForEpoch);
+            CommandStore store = shardFactory.create(id, time, agent, this.store, progressLogFactory, rangesForEpoch);
+            store.execute(() -> CommandStore.register(store));
```
belliottsmith: shouldn't this be an implementation detail of the store?

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -331,6 +331,62 @@ public static <V> AsyncChain<V> failure(Throwable failure)
...
+        {
+            @Override
+            protected void start(BiConsumer<? super T, Throwable> callback)
+            {
+                try
+                {
+                    executor.execute(() -> {
+                        try
+                        {
+                            callback.accept(mapper.apply(v), null);
```
belliottsmith: I believe this is a potential bug, as we generally impose _precisely once_ semantics on our callbacks. If the `mapper.apply` fails we should invoke `callback.accept(null, t)` however if `callback.accept(v, null)` fails we should not, as the callback is now responsible for the outcome, and is expected to be exception-safe. Anything it propagates should be ignored, as its state is unknown to us, and no longer our responsibility.

accord-core/src/main/java/accord/utils/async/AsyncChains.java
dcapwell: added this TODO ` // TODO (low priority, correctness): If the executor is shutdown then the callback may run in an unexpected thread, which may not be thread safe`

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
```diff
@@ -18,26 +18,97 @@
...
+            if (previous == null || previous.isDone())
+            {
+                executor.execute(task);
+            }
+            else
+            {
+                // SimulatedDelayedExecutorService can interleave tasks and is global; this violates a requirement for
+                // CommandStore; single threaded with ordered execution!  To simulate this behavior, add the callback
+                // to enqueue once the "previous" task completes.
+                long nowMilis = executor.nowMillis();
```
belliottsmith: typo

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
```diff
@@ -18,26 +18,97 @@
...
+                previous.addCallback((i1, i2) -> executor.executeWithObservedDelay(task, executor.nowMillis() - nowMilis, TimeUnit.MILLISECONDS));
```
belliottsmith: Is this better than a `Queue` and passing ourselves as the callback, that submits the next task from the `Queue`? Generally a bit easier to debug to have a collection of pending tasks IMO.

accord-core/src/main/java/accord/utils/async/AsyncChains.java
dcapwell: fixed

accord-core/src/main/java/accord/local/CommandStores.java
dcapwell: it could be, but then causes duplication of logic.  As executors are expected to be ordered, this makes sure that the first action is to register and we avoid having to handle in each store uniquely

The `Synchronized` has to redo all this logic as the register makes no sense in that case, rest are all thread based so avoids duplication.

Now, if we have `AccordCommandStore` extend `SingleThread` this would lower the amount of duplication as there are only 2 real implementations in this case... I don't remember why those classes were split.

accord-core/src/main/java/accord/messages/SafeCallback.java
dcapwell: Do you mean the following type setup?

```
Executor - java's
AsyncExecutor extends Executor - adds AsyncChain support
AgentExecutor extends AsyncExecutor - adds "Agent agent();"
CommandStore extends AsyncExecutor
```

Looking at the usage, since the sink API uses `Executor` then the sink needs access to the `Agent`, so is this idea to move away from `Executor` in favor of `AgentExecutor` which has a `agent`?

belliottsmith: Something like this, though I don't know that we need to implement the `Executor` interface at all, we might prefer a different method to `execute`.

accord-core/src/main/java/accord/local/CommandStores.java
belliottsmith: My concern isn't code reuse, but that this is _explicitly_ an implementation detail of the `CommandStore` that has now leaked into `CommandStores`, in a manner that is very far from obvious, and certainly brittle. If you want to make `CommandStore` an abstract class to implement this behaviour, or make a threaded `CommandStore` that does this for others to extend, that would be great of course.

accord-core/src/main/java/accord/messages/SafeCallback.java
dcapwell: pushed to use `SafeCall`

accord-core/src/main/java/accord/messages/SafeCallback.java
dcapwell: took your patch below which only does this if the `onCallbackFailure` failed

accord-core/src/main/java/accord/messages/SafeCallback.java
dcapwell: pushed

accord-core/src/main/java/accord/local/CommandStores.java
dcapwell: pushed this change

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
dcapwell: a `FIFOQueue` would work in this case, let me test this out and get back to you

dcapwell: pushed this change.  I think the code is more complex, but if you are in a debugger its easier to see when the queue is growing

accord-core/src/main/java/accord/local/CommandStores.java
dcapwell: ill look to move this to each impl

dcapwell: Ok I hope I have a solution that works here...

I added a new `register` method to `CommandStore` which `CommandStores` calls on init.  `CommandStore.register` defaults to this same logic and sync store overrides as a noop...

Why did I go this route?  I need to run in the `Executor` once and have access to `this`, but if I do that in the constructor I could access a non-finalized `CommandStore`, so felt its best to do this logic outside of the constructor... which means I need something to trigger...

This makes it so `CommandStores` doesn't know "how" we register, it only knows that we "need" to register...  since `CommandStores` no longer knows the "how" then it doesn't know this implementation detail....

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
belliottsmith: I think the code could be simplified by moving the callback logic to a member method that is _always_ registered as a callback to the currently executing task, e.g.

```
        private <T> void runNextTask()
        {
            Pending next = queue.peek();
            if (next == null)
                return;

            next.task.addCallback(agent()); // used to track unexpected exceptions and notify simulations
            next.task.addCallback(this::afterExecution);
            executor.executeWithObservedDelay(next.task, executor.nowMillis() - next.nowMillis, TimeUnit.MILLISECONDS);
        }

        private void afterExecution()
        {
            queue.poll();
            runNextTask();
        }

        @Override
        public <T> AsyncChain<T> submit(Callable<T> fn)
        {
            Task<T> task = new Task<>(() -> Unsafe.runWith(this, fn));
            boolean wasEmpty = queue.isEmpty();
            queue.add(new Pending(task, executor.nowMillis()));
            if (wasEmpty)
                runNextTask();
            return task;
        }
```

I will be honest, though, I don't quite follow the nowMillis delay logic. Could you explain your reasoning?

dcapwell: > I will be honest, though, I don't quite follow the nowMillis delay logic. Could you explain your reasoning?

I just deleted my long post explaining my reasoning as after finishing it highlights that we *shouldn't* account for the queue delay...

The main reason was to avoid compounding the jitters (if you schedule task1, then task2, then task3, task3's jitter is the task1.jitter + task2.jitter + task3.jitter), but due to the execution forced to be in-order this has to happen regardless... and since the `jitter` can return `0` then we can execute without a context switch in the current model (low probability, but possible)

TL;DR - I believe this logic should be purged...

accord-core/src/main/java/accord/local/CommandStores.java
belliottsmith: I'm a bit confused at this, taking a closer look. The `SingleThread` implementation seems to not override register, even though it seemingly has no reason to use the `Unsafe` implementation, since it is able to just consult its `Thread` register, as this is of course much simpler and cheaper.

However, the `register` method only makes logical sense for the `SingleThread` implementations? The others need to surely maintain it on entry/exit?

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
dcapwell: push change adding your refactor and removing the delay accounting logic

accord-core/src/main/java/accord/local/CommandStores.java
dcapwell: The default is `execute(() -> CommandStore.register(this));` which adds the current to the shared `ThreadLocal`; this default is used by `AccordCommandStore`, `SingleThread`, and `Debug` (as it extends `SingleThread`).

There are 2 non-threaded based `CommandStore`s that mark this method as a no-op: that is `Synchronized` and `DelayedCommandStore`

> The SingleThread implementation seems to not override register, even though it seemingly has no reason to use the Unsafe implementation, since it is able to just consult its Thread register, as this is of course much simpler and cheaper.

What do you mean by `Unsafe`?  The current logic calls `CommandStore.register(this)` and that method *hides* `Unsafe` type...

both `SingleThread` and `AccordCommandStore` schedule a `Runnable` in their `Executor` which will call `CommandStore.register(this)`

belliottsmith: I mean that the `SingleThread` (and extensions) do not need to register, or use the `Unsafe` implementation at all, since their implementation of `inStore()` is:

```
        @Override
        public boolean inStore()
        {
            return thread == Thread.currentThread();
        }
```

However, the other `Synchronised` and `DelayedCommandStore` do need some help, and they rely on `Unsafe` to implement `inStore`... except that their `register` call on construction won't work in tandem with these calls.

dcapwell: > I mean that the SingleThread (and extensions) do not need to register,

If we do not register, how can we fetch the current store? `accord.local.CommandStore#current` is used in `Node` to know which `CommandStore` is running so it can schedule the callback there.

> except that their register call on construction won't work in tandem with these calls.

Correct, which is why their `register` method is a `noop`.  Both need the `Unsafe` methods as they "weave" between stores using the same thread (main).

accord-core/src/main/java/accord/messages/SafeCallback.java
dcapwell: pushed change... I left the dependency of `Executor` alone as that is used directly in `SafeCallback`... we don't "need" the `async` methods, but `AgentExecutor` depending on `AsyncExecutor` I felt was still best...

accord-core/src/main/java/accord/local/CommandStore.java
dcapwell: > Should this be in the InMemoryCommandStore

Not really... If we are working with `CommandStore` we need the API to be present there.

`Unsafe` has 2 things going on:

1) holds the reference to the `ThreadLocal` pointing to the `current` `CommandStore`. 
2) operations for tests to mutate the `ThreadLocal` that maybe "unsafe"

I tried to limit the scope of the `ThreadLocal` reference to make access limited to only the "approved" methods, that was an attempt to limit bugs.  

Now, for the Unsafe operations (num 2), those are used by more than `InMemoryCommandStore`; current usages are

1) `accord.topology.TopologyManager#awaitEpoch` -  Need to know if running in a `CommandStore` or not, so we don't return a `AsyncChain` using the wrong store
2) `accord.local.Node#checkStore` - makes sure that the selected `CommandStore` is the current one; this has been critical to make sure that we don't do the wrong thing.
3) `Synchronized` and `DelayedCommandStores` - since no threading is done, this is to support the interleaving of tasks, which may be running in different `CommandStore`s.

For 1/2 we just need access to `current` that *does not* fail, I left this in `Unsafe` to at least denote that you must be careful, and to try to avoid the wrong code paths calling it... this "could" be moved outside of `Unsafe`, but then I would likely rename to `CommandStore.unsafeMaybeCurrent` or something (just trying to denote not to call it)

for 3, I "could" try to limit this to `InMemoryCommandStore`, but I would need to expose `register` and `remove`; that or have to make the `ThreadLocal` public, which I was trying to avoid

accord-core/src/main/java/accord/local/CommandStores.java
belliottsmith: Ah, I missed that this implements `current`. 

Why don't we make `CommandStore` an abstract class? It was before, and it will be again (Bootstrap does this to bring shared functionality into it). No need to have an `Unsafe`, just control it from there.

Also, we can safely register from within the constructor - we already "unsafe" publish the `CommandStore` into the thread to setup the `thread` field, this is no different, and it would be cleaner.

dcapwell: > we already "unsafe" publish the CommandStore into the thread to setup the thread field

Are you talking about `SingleThread`'s 

```
executor.execute(() -> thread = Thread.currentThread());
```

This does not expose `this` for other callers to see, and uses a param rather than a field, so we don't "unsafe publish"

> this is no different, and it would be cleaner.

I disagree, this is different as we are exposing the field for others to consume.  My understanding of JMM is that this publish isn't guarantied to be visible in the new thread even after the constructor completes, which could lead to a task seeing a partial `CommandStore`.

How about this compromise... we take a `CommandStore.Factory` and not actually directly call `new`... so the `Thread` based `CommandStores` could do `new` + `register` in that factory so this doesn't leak into `CommandStores`?

dcapwell: Pushed a change (see 12c558ea6cabdbb4beb42883545e1d7177006ca8) so `SingleThread` and `AccordCommandStore` both use a new `create` method that hides `new` and will register with the `ThreadLocal`.  This let me hide this detail from `CommandStores`

belliottsmith: This is definitely safe from a JMM perspective, as we cannot possibly _consume_ this pointer until we are executing a command submitted after the CommandStore is constructed. We only use the `current` pointer in a later `execute` method, and only from this specific `Thread` (by definition). We do not actually use the pointer until it is safely published via a pair of volatile writes and reads.

dcapwell: moved `CommandStore` to an `abstract class` and removed `Unsafe`; the `runWith` methods are now `protected` methods 

see 69cc0b0aed695a50329e6fa90346bb9c74b5262c

accord-core/src/main/java/accord/local/CommandStore.java
dcapwell: Pushed 69cc0b0aed695a50329e6fa90346bb9c74b5262c which made `CommandStore` and `abstract class` which let me get rid of `Unsafe`

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -837,44 +839,43 @@ public void shutdown() {}
...
-        private SingleThread(int id, NodeTimeService time, Agent agent, DataStore store, ProgressLog.Factory progressLogFactory, RangesForEpochHolder rangesForEpochHolder, ExecutorService executor)
-        {
-            super(id, time, agent, store, progressLogFactory, rangesForEpochHolder);
-            this.executor = executor;
+            });
+            executor.execute(() -> thread = Thread.currentThread());
         }
 
-        public static CommandStore.Factory factory(ExecutorService executor)
+        public static SingleThread create(int id, NodeTimeService time, Agent agent, DataStore store, ProgressLog.Factory progressLogFactory, RangesForEpochHolder rangesForEpochHolder)
```
belliottsmith: Minor refactor suggestion:

```
        interface Constructor<ST extends SingleThread>
        {
            ST construct(int id, NodeTimeService time, Agent agent, DataStore store, ProgressLog.Factory progressLogFactory, RangesForEpochHolder rangesForEpochHolder);
        }

        public static SingleThread create(int id, NodeTimeService time, Agent agent, DataStore store, ProgressLog.Factory progressLogFactory, RangesForEpochHolder rangesForEpochHolder)
        {
            return create(id, time, agent, store, progressLogFactory, rangesForEpochHolder, SingleThread::new);
        }

        protected static <ST extends SingleThread> ST create(int id, NodeTimeService time, Agent agent, DataStore store, ProgressLog.Factory progressLogFactory, RangesForEpochHolder rangesForEpochHolder, Constructor<ST> constructor)
        {
            ST st = constructor.construct(id, time, agent, store, progressLogFactory, rangesForEpochHolder);
            st.execute(() -> st.thread = Thread.currentThread());
            st.execute(() -> CommandStore.register(st));
            return st;
        }
```

and then use the same for `Debug` (this also requires making `thread` package-private, but it is probably overall cleaner this way)

(Though I would personally opt for just doing this in the constructor, since it is safe, I'm fine with this approach if we make it clear how extending classes should re-use the factory method)

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -43,16 +45,113 @@ CommandStore create(int id,
                             RangesForEpochHolder rangesForEpoch);
     }
 
-    int id();
-    Agent agent();
-    AsyncChain<Void> execute(PreLoadContext context, Consumer<? super SafeCommandStore> consumer);
-    <T> AsyncChain<T> submit(PreLoadContext context, Function<? super SafeCommandStore, T> apply);
+    @Nullable
+    public static CommandStore maybeCurrent()
```
belliottsmith: [Preference]: I would prefer not to clutter the preamble with these methods, and would stick them at the bottom of the class.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
dcapwell: reading more on JMM it does seem that grouping `final` fields *before* this "leak" is safe from the final field point of view (scheduling on the thread before the `final` fields may not see them set it seems) and may not be safe for non-final fields... in our case those are all local thread processing, so would be fine...

Given this, I can revert back to constructor and document this assumption.

belliottsmith (approved):

belliottsmith:
I am fairly sure this has merged also, including on the C* side. So closing for now.

PR #39, opened 2023-04-02
CASSANDRA-18221: Add AccordConfig to track accord specific configs

kamalesh0406:
patch by Kamalesh Palanisamy;

### DIFF
--- a/accord-core/src/main/java/accord/impl/SimpleProgressLog.java
+++ b/accord-core/src/main/java/accord/impl/SimpleProgressLog.java
@@ -29,6 +29,7 @@ import java.util.function.BiConsumer;
 
 import javax.annotation.Nullable;
 
+import accord.utils.AccordConfig;
 import accord.utils.IntrusiveLinkedList;
 import accord.utils.IntrusiveLinkedListNode;
 import accord.coordinate.*;
@@ -88,11 +89,13 @@ public class SimpleProgressLog implements ProgressLog.Factory
     enum DisseminateStatus { NotExecuted, Durable, Done }
 
     final Node node;
+    final AccordConfig config;
     final List<Instance> instances = new CopyOnWriteArrayList<>();
 
-    public SimpleProgressLog(Node node)
+    public SimpleProgressLog(Node node, AccordConfig config)
     {
         this.node = node;
+        this.config = config;
     }
 
     class Instance extends IntrusiveLinkedList<Instance.State.Monitoring> implements ProgressLog, Runnable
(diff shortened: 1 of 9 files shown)


accord-core/src/main/java/accord/impl/SimpleProgressLog.java
```diff
@@ -88,11 +89,13 @@ boolean isAtLeastCommitted()
     enum DisseminateStatus { NotExecuted, Durable, Done }
 
     final Node node;
+    final AccordConfig config;
     final List<Instance> instances = new CopyOnWriteArrayList<>();
 
-    public SimpleProgressLog(Node node)
+    public SimpleProgressLog(Node node, AccordConfig config)
```
dcapwell, line 95: *should* `Node` expose config?  That way you don't need to pass all over?

accord-core/src/main/java/accord/impl/SimpleProgressLog.java
```diff
@@ -29,6 +29,7 @@
 
 import javax.annotation.Nullable;
 
+import accord.utils.AccordConfig;
```
dcapwell, line 32: I feel we should have `accord.config` package, this is likely to expand over time

accord-core/src/main/java/accord/impl/SimpleProgressLog.java
```diff
@@ -822,7 +825,7 @@ void ensureScheduled()
                 return;
 
             isScheduled = true;
-            node.scheduler().once(() -> commandStore.execute(PreLoadContext.empty(), ignore -> run()).begin(commandStore.agent()), 200L, TimeUnit.MILLISECONDS);
+            node.scheduler().once(() -> commandStore.execute(PreLoadContext.empty(), ignore -> run()).begin(commandStore.agent()), config.progress_log_scheduler_delay_in_ms, TimeUnit.MILLISECONDS);
```
dcapwell, line 828: we are trying to move away from `_in_ms` postfix... this may call out that we should likely move some of the config classes to accord?

accord-core/src/main/java/accord/utils/AccordConfig.java
```diff
@@ -0,0 +1,11 @@
+package accord.utils;
+
+public class AccordConfig {
+    public long progress_log_scheduler_delay_in_ms = 200L;
+
+    public AccordConfig(long progress_log_scheduler_delay_in_ms) {
+        this.progress_log_scheduler_delay_in_ms = progress_log_scheduler_delay_in_ms;
+    }
```
dcapwell, line 8: this constructor isn't something we can maintain, as we add more and more configs this will become an issue

dcapwell (changes requested):

PR #40, opened 2023-04-04
CASSANDRA-18422: when status changes after read result, reject

dcapwell:


accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -219,12 +248,22 @@ private void ack()
...
+            {
+                node.reply(replyTo, replyContext, new ReadOk(data));
+            }
+        }
     }
 
     private synchronized void readComplete(CommandStore commandStore, Data result)
     {
-        Invariants.checkState(waitingOn.get(commandStore.id()), "Waiting on does not contain store %d; waitingOn=%s", commandStore.id(), waitingOn);
+        Invariants.checkState(waitingOn.get(commandStore.id()), "Txn %s's waiting on does not contain store %d; waitingOn=%s", txnId, commandStore.id(), waitingOn);
```
dcapwell, line 286: fixed by https://issues.apache.org/jira/browse/CASSANDRA-18364

accord-core/src/main/java/accord/utils/async/AsyncResult.java
```diff
@@ -79,7 +79,11 @@ default void setSuccess(V value)
         default void setFailure(Throwable throwable)
         {
             if (!tryFailure(throwable))
-                throw new IllegalStateException("Result has already been set on " + this);
+            {
+                IllegalStateException e = new IllegalStateException("Result has already been set on " + this);
```
dcapwell, line 83: `ReadData.readComplete` some times throws causing this to happen.  Rather than dropping the throwable that we are rejecting, add it to the throwable we are throwing; that way logs can show what caused this to happen.

This doesn't happen once https://issues.apache.org/jira/browse/CASSANDRA-18364 is applied

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -191,10 +198,32 @@ static void burn(RandomSource random, TopologyFactory topologyFactory, List<Id>
...
+                return globalExecutor.submit(fn::get, delay, unit);
+            }
+
+            @Override
+            public AsyncChain<Void> execute(Runnable fn)
+            {
+                return globalExecutor.submit(fn).map(ignore -> null);
+            }
+        };
+        Function<CommandStore, ListExecutor> executor = ignore -> listExecutor;
```
dcapwell: I did this for backwards compatibility, wondering if I should simplify this by just passing in a `SimulatedDelayedExecutorService` or create a new `AsyncExecutor` interface?

accord-core/src/test/java/accord/impl/basic/SimulatedDelayedExecutorService.java
```diff
@@ -18,20 +18,15 @@
...
-import java.util.Collections;
-import java.util.List;
-import java.util.concurrent.AbstractExecutorService;
-import java.util.concurrent.Callable;
-import java.util.concurrent.Executors;
-import java.util.concurrent.FutureTask;
-import java.util.concurrent.TimeUnit;
-
-public class SimulatedDelayedExecutorService extends AbstractExecutorService
+public class SimulatedDelayedExecutorService extends TaskExecutorService
```
dcapwell, line 29: copy/pasted change from https://issues.apache.org/jira/browse/CASSANDRA-18364

accord-core/src/test/java/accord/impl/basic/TaskExecutorService.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.RunnableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import accord.utils.async.AsyncResults;
+
+public abstract class TaskExecutorService extends AbstractExecutorService
```
dcapwell: copy pasted from https://issues.apache.org/jira/browse/CASSANDRA-18364

accord-core/src/test/java/accord/impl/list/ListExecutor.java
```diff
@@ -0,0 +1,58 @@
...
+package accord.impl.list;
+
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import accord.local.CommandStore;
+import accord.local.PreLoadContext;
+import accord.utils.async.AsyncChain;
+
+public interface ListExecutor
```
dcapwell: not a fan of this, wondering if we should make a `AsyncExecutor` and have `CommandStore` implement as well?

accord-core/src/test/java/accord/messages/ReadDataTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+        Mockito.verifyNoInteractions(sink);
+
+        check(store.execute(PreLoadContext.contextFor(txnId, keys), safe -> {
+            CheckedCommands.apply(safe, txnId, safe.latestEpoch(), route, executeAt, deps, writes, result);
+        }));
+
+        readResult.setSuccess(data);
+
+        Mockito.verify(sink, Mockito.atMost(2)).reply(Mockito.eq(node.id()), Mockito.eq(replyContext), Mockito.eq(ReadData.ReadNack.Redundant));
```
dcapwell: `accord.messages.ReadData#apply` sets `isObsolete=true` without sending the nack, so there could be the following setup

store1: status=ReadyToExecute, send read
store2: status=PreApplied, set isObsolete

in this case when the read from `store1` completes it needs to know to ignore and that the nack wasn't set... I could switch this to an enum `{ALL_GOOD, SENT_NACK, NEED_TO_SEND_NACK}` which would avoid the second nack...  figured this was simplest

dcapwell: actually, in that case `apply` sets the field but returns `Redundant` so `accord.messages.ReadData#accept` should send that...

and `accord.messages.ReadData#obsolete` sends directly... so under all cases we send?  ill play around with this

dcapwell: fixed this and pushed tests to hit all 3 cases

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -53,7 +54,35 @@ public static ReadData create(TxnId txnId, Seekables<?, ?> scope, long executeAt
             return new ReadData(txnId, scope, executeAtEpoch, waitForEpoch);
         }
     }
+    private class ObsoleteTracker implements CommandListener
```
aweisberg, line 57: Instead of allocating a separate obsolete tracker could it be done in the ReadData `onChange` method?

It seems like the obsolete field is concurrently read/written by multiple command stores and might end up sending multiple replies back?

Would it be better/simpler to mark it as obsolete then ack the local map reduce so that the existing code that waits to hear from all the command stores runs as normal?

I don't understand the threading of this in general since command listeners run on arbitrary threads at arbitrary times?

It seems like `apply` takes care to be synchronized as does `onChange`

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -53,7 +54,35 @@ public static ReadData create(TxnId txnId, Seekables<?, ?> scope, long executeAt
...
+    {
+        @Override
+        public void onChange(SafeCommandStore safeStore, SafeCommand safeCommand)
+        {
+            switch (safeCommand.current().status())
```
aweisberg, line 62: This doesn't check whether the read already successfully completed before the transition to `PreApplied` or `Applied` for this command store. It may be that it wasn't obsoleted.

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -324,6 +334,7 @@ public static void main(String[] args)
     @Test
     public void testOne()
     {
+        for (; true; )
```
aweisberg: Is this supposed to be here still?

aweisberg (changes requested):
I think I found some concurrency issues in `ReadData`. Everything else looks good I am just drilling down into the unit test.

accord-core/src/main/java/accord/messages/ReadData.java
dcapwell: > Instead of allocating a separate obsolete tracker could it be done in the ReadData onChange method?

we would then require special handle to avoid dup calls to `read`.  If we see the status `ReadyToExecute` twice (due to something else changing) then we would trigger `read` another time.  By removing the listener and adding a new one, we avoid this

> It seems like the obsolete field is concurrently read/written by multiple command stores

read by different stores is fine.  write is guarded by `accord.messages.ReadData#obsolete` but that isn't sync, so we could have duplicates if 2 stores try to call at the same time; I can add a test for this to make sure its a single reply, though don't think I can actually force this timing event to happen... so not 100% sure I can test this... but I can at least have both stores fail in a test... good to do anyways

ATM duplicate replies is fine as the second one is ignored, but better to optimize it away.

> Would it be better/simpler to mark it as obsolete then ack the local map reduce

the MapReduce has a state where it doesn't know the result as it relies on listeners... so only the case where `apply` marks as `obsolete` can be handled in MapReduce; which it currently does.

> I don't understand the threading of this in general since command listeners run on arbitrary threads at arbitrary times?

listeners are per-store and are single threaded.  You attach a listener to a pair of `(command, store)`, and always run in that same `store`.  

Now, there is a bug with listeners that is improved in https://issues.apache.org/jira/browse/CASSANDRA-18364 and solved in a followup from Benedict... so this logic is currently safe based off expected semantics, but current patch could produce multiple `read` events.

The bug that is fixed in CASSANDRA-18364 is a recursion bug: 
* an event causes a command to trigger listeners
** a listener changes the command, causing listeners to be called
*** listener removes self, so should not be called again
* removed listener is called again as its referenced from an `immutable` list of listeners... 

The last step is fixed in CASSANDRA-18364 by double checking that the listener still exists and Benedict said he was reworking all this logic.

accord-core/src/main/java/accord/messages/ReadData.java
dcapwell: good call... 

if we reply back, the listener still exists (as we didn't remove) so the next step would trigger an obsolete reply; which would get ignored...

We do the following on the `apply` failure case

```
node.commandStores().mapReduceConsume(this, waitingOn.stream(), forEach(in -> removeListener(in, txnId), node.agent()));
```

so could do the same in `ack`.

Now, the better solution is to do the same, but switch from `boolean` to `enum`.  if we reply back *before* `PreAccept`, then that's fine... so we should ignore in that case

ill look to improve this

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: nope... I add it to run for a long time... forgot to remove

accord-core/src/main/java/accord/messages/ReadData.java
dcapwell: there was a bug I added to `accord.messages.ReadData#obsolete` as that was always guarded by `this` but with my new listener it wasn't, so I fixed that.

accord-core/src/main/java/accord/messages/Defer.java
```diff
@@ -80,7 +80,7 @@ public void onChange(SafeCommandStore safeStore, SafeCommand safeCommand)
 
         int id = safeStore.commandStore().id();
         if (!waitingOn.contains(id))
-            throw new IllegalStateException();
+            throw new IllegalStateException("Not waiting on CommandStore " + id);
```
dcapwell: this is from https://issues.apache.org/jira/browse/CASSANDRA-18364, I hit in burn test so patched here to get stable

accord-core/src/main/java/accord/messages/Commit.java
```diff
@@ -162,7 +162,8 @@ public synchronized ReadNack apply(SafeCommandStore safeStore)
             case Insufficient:
                 SafeCommand safeCommand = safeStore.command(txnId);
                 Invariants.checkState(!safeCommand.current().known().isDefinitionKnown());
-                if (defer == null)
+                // When isDone=true this.process() is called and all CommandStores run apply again; so safe to create a new one
```
dcapwell: this is from https://issues.apache.org/jira/browse/CASSANDRA-18364, I hit in burn test so patched here to get stable

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -53,12 +54,44 @@ public static ReadData create(TxnId txnId, Seekables<?, ?> scope, long executeAt
             return new ReadData(txnId, scope, executeAtEpoch, waitForEpoch);
         }
     }
+    private class ObsoleteTracker implements CommandListener
+    {
+        @Override
+        public void onChange(SafeCommandStore safeStore, SafeCommand safeCommand)
+        {
+            synchronized (ReadData.this)
```
aweisberg: This could be narrowed to just synchronizing while calling `obsolete`?

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -251,9 +314,9 @@ private void read(SafeCommandStore safeStore, Command.Committed command)
 
     void obsolete()
     {
-        if (!isObsolete)
+        if (state == State.PENDING)
```
aweisberg, line 314: Does this also have an illegal state we could check for if we try to mark it as obsolete after already having sent the ack?

dcapwell: if its `RETURNED` then that's fine, it just means that the txn was invalidated or applied after the fact.

If its OBSOLETE that just means another store set it

so both cases should be fine to ignore

accord-core/src/main/java/accord/messages/ReadData.java
aweisberg: I don't feel strongly about trying to only have one listener. It's might be a little clearer to have separate listener and not try and figure out in a single listener what he current state is.

aweisberg (approved):
I think you got it. If the burn tests are happy 🚢 it

belliottsmith:
Again, this (or some version) has been merged already, 99.999% sure

PR #41, opened 2023-04-07
CASSANDRA-18378: Fixed Defer to not complete until all stages have been processed

dcapwell:


bdeggleston (approved):
+1

belliottsmith:
@iamaleksey has removed `Defer` entirely, so this is now redundant.

PR #44, opened 2023-04-25
CASSANDRA-18471: CEP-15 Accord: NotWitnessed commands can receive an invalidate promise but would return Zero instead

dcapwell:


belliottsmith (approved):

belliottsmith:
Some version of this has also merged.

PR #45, opened 2023-04-27
Bootstrap

belliottsmith:
Introduce mechanisms for joining a cluster without receiving the full transaction log


.gitignore
```diff
@@ -3,3 +3,4 @@
 build/
 accord-core/build/
 accord-maelstrom/build/
+.rat-excludes.txt
```
dcapwell, line 6: I don't see this patch touching this?

accord-core/src/main/java/accord/api/ConfigurationService.java
```diff
@@ -53,19 +55,67 @@
...
+         * The node has successfully replicated the underlying DataStore information for the new epoch, but may need
+         * to perform some additional coordination before it can execute the read portion of a transaction.
+         */
+        public final AsyncResult<Void> data;
+
+        /**
+         * The node has retrieved enough remote information to safely process reads, including replicating all
+         * necessary DataStore information, and any additional transactions necessary for consistency.
+         */
+        public final AsyncResult<Void> reads;
```
dcapwell, line 87: this api maybe problematic... if you are in a command store and the future is not completed, then your callback will run in the epoch ready publisher's thread... the existing topology logic I had to use `AsyncChain` and detect this for callers...

.gitignore
belliottsmith: nope, it's so I can build locally

accord-core/src/main/java/accord/api/ConfigurationService.java
belliottsmith: I think that's up to the consumer to guarantee?

accord-core/src/main/java/accord/coordinate/CheckOn.java
```diff
@@ -206,8 +208,9 @@ public Void apply(SafeCommandStore safeStore)
                         break;
 
                 case PreAccepted:
-                    if (!safeStore.ranges().at(txnId.epoch()).isEmpty())
-                        Commands.preaccept(safeStore, txnId, partialTxn, maxRoute, progressKey);
+                    // only preaccept if we coordinate the transaction
+                    if (untilLocalEpoch <= txnId.epoch() || safeStore.ranges().coordinates(txnId).intersects(maxRoute))
```
bdeggleston, line 212: I'm not understanding the purpose of this change. I think this would only be true (and the other side of the || not) would be if we learned of a txnId < our current epoch

accord-core/src/main/java/accord/coordinate/CoordinateNoOp.java
```diff
@@ -0,0 +1,100 @@
...
+                void onAccepted()
+                {
+                    Topologies applyTo = node.topology().forEpoch(route, executeAt.epoch());
+                    Topologies persistTo = txnId.epoch() == executeAt.epoch() ? applyTo : node.topology().preciseEpochs(route, txnId.epoch(), executeAt.epoch());
+                    Writes writes = txn.execute(txnId, txnId, null);
+                    Result result = txn.result(txnId, executeAt, null);
+                    // TODO (now): permit Apply to Commit in these cases
+                    Set<Node.Id> nodes = persistTo.nodes();
+                    node.send(nodes, id -> new Commit(Maximal, id, persistTo.forEpoch(txnId.epoch()), persistTo, txnId, txn, route, Ranges.EMPTY, executeAt, deps, false));
+                    node.send(nodes, id -> new Apply(id, persistTo, applyTo, executeAt.epoch(), txnId, route, txn, executeAt, deps, writes, result));
```
bdeggleston, line 94: wouldn't the Commit message be redundant for any node also receiving an Apply message?

accord-core/src/main/java/accord/coordinate/Persist.java
```diff
@@ -53,18 +53,22 @@ public class Persist implements Callback<ApplyReply>
...
     {
-        Topologies sendTo = node.topology().preciseEpochs(route, txnId.epoch(), executeAt.epoch());
+        Topologies coordinate = node.topology().forEpoch(route, txnId.epoch());
         Topologies applyTo = node.topology().forEpoch(route, executeAt.epoch());
-        Persist persist = new Persist(node, sendTo, txnId, route, txn, executeAt, deps);
-        node.send(sendTo.nodes(), to -> new Apply(to, sendTo, applyTo, executeAt.epoch(), txnId, route, txn, executeAt, deps, writes, result), persist);
+        Topologies persistTo = txnId.epoch() == executeAt.epoch() ? applyTo : node.topology().preciseEpochs(route, txnId.epoch(), executeAt.epoch());
+        Persist persist = new Persist(node, persistTo, txnId, route, txn, executeAt, deps);
+        node.send(persistTo.nodes(), to -> new Commit(Maximal, to, coordinate.current(), persistTo, txnId, txn, route, null, executeAt, deps, false), persist);
+        node.send(applyTo.nodes(), to -> new Apply(to, persistTo, applyTo, executeAt.epoch(), txnId, route, txn, executeAt, deps, writes, result), persist);
```
bdeggleston, line 71: same question here wrt redundant messages

accord-core/src/main/java/accord/coordinate/tracking/FastPathTracker.java
```diff
@@ -30,17 +31,50 @@
...
+        }
+
+        public boolean hasRejectedFastPath()
+        {
+            return false;
+        }
+
+        public boolean hasMetFastPathCriteria()
+        {
+            return true;
```
bdeggleston, line 70: don't we want to disable fast path commits if we need to talk with multiple epochs, or has that been relaxed? I vaguely remember hearing that it was, but am not sure

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -291,18 +333,49 @@ private void forEach(Routable keyOrRange, Ranges slice, Consumer<RoutableKey> fo
...
+        deps.keyDeps.keys().forEach(allRanges, key -> {
+            SafeCommandsForKey cfk = commandsForKey(key).createSafeReference();
+            deps.keyDeps.forEach(key, txnId -> {
+                // TODO (desired, efficiency): this can be made more efficient by batching by epoch
+                if (rangesForEpochHolder.get().coordinates(txnId).contains(key))
+                    return; // already coordinates, no need to replicate
+                if (!rangesForEpochHolder.get().allBefore(txnId.epoch()).contains(key))
+                    return;
+
+                cfk.registerNotWitnessed(txnId);
```
bdeggleston, line 349: it's not clear why we're registering txnIds for ranges we used to replicate.

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -83,6 +123,65 @@ public Agent agent()
 
     public abstract void shutdown();
 
+    // implementations are expected to override this for persistence
+    protected void setRejectBefore(ReducingRangeMap<Timestamp> newRejectBefore)
+    {
```
bdeggleston, line 128: could you make this abstract, since the implementor is expected to override?

bdeggleston: same for `setBootstrapBeganAt` and `setSafeToRead`

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -113,23 +120,71 @@ RangesForEpoch ranges()
...
+
+        public String toString()
+        {
+            return store.id() + " " + ranges.current;
+        }
     }
 
     public static class RangesForEpoch
     {
+        public enum Subset
```
bdeggleston: this is unused

accord-core/src/main/java/accord/messages/MessageType.java
```diff
@@ -47,5 +47,7 @@
     INFORM_DURABLE_REQ,
     INFORM_HOME_DURABLE_REQ,
     CHECK_STATUS_REQ,
-    CHECK_STATUS_RSP
+    CHECK_STATUS_RSP,
+    GET_SYNC_POINT_REQ,
+    GET_SYNC_POINT_RSP,
```
bdeggleston: these are unused

accord-core/src/main/java/accord/messages/PreAccept.java
```diff
@@ -239,6 +251,76 @@ private static <T extends Deps> T calculateDeps(SafeCommandStore commandStore, T
...
+                Shard ov = originalShards.get(o);
+                {
+                    int c = nv.range.compareIntersecting(ov.range);
+                    if (c < 0) { ++n; continue; }
+                    else if (c > 0) { ++o; continue; }
+                }
+                int nvi = 0, ovi = 0;
+                while (nvi < nv.nodes.size() && ovi < ov.nodes.size())
+                {
+                    int c = nv.nodes.get(nvi).compareTo(ov.nodes.get(ovi));
```
bdeggleston: this should be using `Shard.sortedNodes` for it's comparison

accord-core/src/main/java/accord/messages/PreAcceptSyncPoint.java
```diff
@@ -0,0 +1,85 @@
...
+import accord.primitives.PartialRoute;
+import accord.primitives.PartialTxn;
+import accord.primitives.Ranges;
+import accord.primitives.Txn;
+import accord.primitives.TxnId;
+import accord.topology.Topologies;
+
+import static accord.primitives.Txn.Kind.ExclusiveSyncPoint;
+
+public class PreAcceptSyncPoint extends PreAccept
```
bdeggleston: this is unused (and is missing a message type)

accord-core/src/main/java/accord/primitives/TxnId.java
```diff
@@ -120,7 +128,7 @@ private static Domain domain(int flags)
 
     private static int rwOrdinal(int flags)
     {
-        return (flags >> 1) & 1;
+        return (flags >> 1) & 7;
```
bdeggleston, line 131: not specific to this patch, but could we rename `rw` -> `kind`?

accord-core/src/main/java/accord/topology/Shard.java
```diff
@@ -151,4 +165,14 @@ public int hashCode()
     {
         return range.hashCode();
     }
+
+    private static boolean isSortedUnique(List<Id> test)
```
bdeggleston: this is unused

bdeggleston (changes requested):
I'm not finished reviewing yet, but figured I'd put up the feedback I have so far, since I'll be out at the end of the week. Overall this looks ok, I just need to spend some more time thinking about how this all fits together. I think I've also missed (ie was not the reviewer on) some behavioral changes, so some of my comments are just asking for some clarification on things that aren't making sense.

.gitignore
bdeggleston: can you add a basic excludes file? Otherwise it doesn't want to build

accord-core/src/main/java/accord/api/ConfigurationService.java
dcapwell: that's possible, I moved away from that for topology manager as it was touched all over..  this was a bug that took a while to get correct in my patch, which is why I am calling out

accord-core/src/main/java/accord/coordinate/CheckOn.java
belliottsmith: we can Accept commands that aren't owned by us for either PreAccept OR Execute, and in these cases we may know of a transaction. Also, we may learn of a transaction that we adopt a dependency on (and need to see the execution of) but also never actually intersects with us for coordination or execution. I think this case might be able to happen simply for transactions we don't coordinate but do execute, though (maybe not). Plus no doubt other weird things I'm not thinking of.

.gitignore
belliottsmith: hmm, does it work to check-in a file and also add to .gitignore? I vaguely recall it might

accord-core/src/main/java/accord/coordinate/CoordinateNoOp.java
belliottsmith: I believe we only send _new_ dependencies as part of `Apply` and also of course the topologies may be different. I believe there's TODOs to improve this, though I could tackle it now.

accord-core/src/main/java/accord/coordinate/tracking/FastPathTracker.java
belliottsmith: Only when talking to _new_ epochs; if the latest epoch is the same as the txnId then it's fine (and should always have been fine)

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
belliottsmith: we might not have witnessed it?

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: I can introduce an additional method, perhaps, though we probably need to think this through when we integrate anyway so it probably doesn't help much yet. But, to explain my thinking here: all of the implementation details are kept within Accord so that the implementation cannot mess with it (though of course it must invoke `super`). I don't really want the implementation controlling the data structure itself, just to have the ability to persist.

accord-core/src/main/java/accord/primitives/TxnId.java
belliottsmith: We have a bunch of `Kind`, but I agree `rw` is now bad terminology. I'll ponder what alternatives there might be, and same for the name within Txn

.gitignore
belliottsmith: for now I've updated the build to only use it if present

dcapwell: https://app.circleci.com/pipelines/github/belliottsmith/cassandra/452/workflows/866c7678-a67b-45cd-97b1-986fca944af3/jobs/7917

```
[exec]   - In plugin 'org.nosphere.apache.rat-base' type 'org.nosphere.apache.rat.RatTask' property 'excludeFile' specifies file '/home/cassandra/cassandra/modules/accord/.rat-excludes.txt' which doesn't exist.
     [exec]  
```

dcapwell: > hmm, does it work to check-in a file and also add to .gitignore? I vaguely recall it might

yes, you are telling git to not add/track changes... so if you to something like `git commit -a` it won't add the changes to that file to your commit

bdeggleston: you can manually add the change though. This is even better though:

> for now I've updated the build to only use it if present

accord-core/src/main/java/accord/coordinate/CoordinateNoOp.java
bdeggleston: ah yes, the TODO is right above this code :)

That's fine then, obviously it would be better to be able to just send apply, but no need to block bootstrap on it.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
bdeggleston: Sure, but we're not registering txnIds for ranges we didn't used to replicate (asking in the context of bootstrap) so why the distinction there?

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -83,6 +123,69 @@ public Agent agent()
...
+    protected void setBootstrapBeganAt(NavigableMap<TxnId, Ranges> newBootstrapBeganAt)
+    {
+        this.bootstrapBeganAt = newBootstrapBeganAt;
+    }
+
+    protected abstract void registerHistoricalTransactions(Deps deps);
+<<<<<<< Updated upstream
+=======
+    protected abstract void evictLog(TxnId txnIdBefore);
+>>>>>>> Stashed changes
```
bdeggleston: unresolved conflict

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -154,4 +257,219 @@ public static void checkNotInStore()
...
+        FullRoute<?> route = node.computeRoute(id, ranges);
+        CollectDeps.withDeps(node, id, route, ranges, before, (deps, fail) -> {
+            if (fail != null)
+            {
+                fetchMajorityDeps(coordination, node, epoch, ranges);
+            }
+            else
+            {
+                execute(contextFor(null, deps.txnIds()), safeStore -> {
+                    registerHistoricalTransactions(deps);
```
bdeggleston, line 308: don't we also want to determine the fast path outcomes of each of these txns before committing on the fast path in the new epoch/topology?

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
belliottsmith: This is for preaccept dependencies. If we are *currently* a coordination replica then if we don’t know about a transaction it doesn’t affect any quorum, as either it had reached a quorum without us and that’s fine, or it hadn’t and that’s also fine. This is just the transaction log history we need to replicate to ourselves (for ranges we own) so that transactions no longer need to contact older epochs for their dependencies.

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: Oh, whoops that seems to have peeked into the future

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: They don’t really vote, they just provide dependencies. Same is essentially true for future epochs and the Accept round, only we don’t currently distinguish there. Might be clearer to have dedicated messages for these purposes, but it might also complicate things a little.

belliottsmith: Oh wait, maybe I misunderstood your query as the context is the registering if "historical transactions" (may be we can name it better - we're registering transactions for use in `calculateDependencies`)

But their fast or slow path outcomes should not be relevant here, we're just propagating deps before fully deprecating the old epoch.

bdeggleston: I'm referring to the part in the paper where we say new fast path electorates can't commit on the fast path until they've been made aware of fast path commits from the previous epoch. However I realized this morning that this should be sufficient, since this will affect the cfk max timestamp and include txns from the previous epoch in preaccept deps, which would allow the new fast path electorate to safely commit on the fast path

bdeggleston: ah nevermind, I'm conflating `CommandStore#sync` calling this with the caller of `ConfigurationService.Listener#onEpochSyncComplete`

bdeggleston (approved):
+1

PR #46, opened 2023-05-16
CASSANDRA-18519: (CEP-15: (C*) Add notion of CommandsForRanges and make this durable in C*)

dcapwell:


accord-core/src/main/java/accord/impl/CommandTimeseries.java
```diff
@@ -0,0 +1,224 @@
...
+import accord.primitives.Seekable;
+import accord.primitives.Timestamp;
+import accord.primitives.TxnId;
+
+import static accord.local.SafeCommandStore.TestDep.ANY_DEPS;
+import static accord.local.SafeCommandStore.TestDep.WITH;
+import static accord.utils.Utils.ensureSortedImmutable;
+import static accord.utils.Utils.ensureSortedMutable;
+
+public class CommandTimeseries<D>
```
dcapwell, line 45: just moved this out of `CommandsForKeys` so that the range logic didn't have to duplicate anything...  I just build this time series from the data on-demand

accord-core/src/main/java/accord/impl/CommandTimeseries.java
```diff
@@ -0,0 +1,224 @@
...
+{
+    public enum TestTimestamp
+    {BEFORE, AFTER}
+
+    private final Seekable keyOrRange;
```
dcapwell, line 50: this type change is the only real change to this class

accord-core/src/main/java/accord/impl/CommandTimeseriesHolder.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package accord.impl;
+
+public interface CommandTimeseriesHolder
```
dcapwell, line 23: not a fan of the name, so any recommendations welcome =)

accord-core/src/main/java/accord/topology/Topology.java
```diff
@@ -236,7 +236,9 @@ private Topology forSubset(int[] newSubset, Collection<Id> nodes)
         return new Topology(epoch, shards, ranges, nodeLookup, rangeSubset, newSubset);
     }
 
-    private <P1> int[] subsetFor(Unseekables<?, ?> select, IndexedPredicate<P1> predicate, P1 param)
+    private enum OnUnknown { REJECT, IGNORE }
```
dcapwell: this was actually a hack to make progress...

What I saw was that the local epoch=1, but we saw epoch=2, we try to see if there is overlap and fail as epoch=1 doesn't have the new table range that is defined in epoch=2

I changed C* code to defer exec of messages and `PreLoadContext` until the epoch they depend on is ready... which may remove the need for this hack...

This likely hasn't been found in burn due to working with a single range, so only would impact new ranges being added and used right away (which happens in C* tests)

accord-core/src/main/java/accord/coordinate/RangeUnavailable.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package accord.coordinate;
+
+import accord.primitives.Ranges;
+import accord.primitives.TxnId;
+
+public class RangeUnavailable extends Exhausted
```
dcapwell, line 24: `Exhausted` could be different things, so this makes it easier to see what happened; what table ranges could not be read?

accord-core/src/main/java/accord/impl/CommandTimeseries.java
```diff
@@ -0,0 +1,233 @@
...
+    protected final CommandLoader<D> loader;
+    public final ImmutableSortedMap<Timestamp, D> commands;
+    private final boolean ignoreTestKind;
```
dcapwell: added this as the test kind is filtering out sync txn... its kinda a hack to have reusable code

accord-core/src/main/java/accord/impl/CommandTimeseries.java
```diff
@@ -0,0 +1,233 @@
...
+     */
+    public <T> T mapReduce(SafeCommandStore.TestKind testKind, TestTimestamp testTimestamp, Timestamp timestamp,
+                           SafeCommandStore.TestDep testDep, @Nullable TxnId depId,
+                           @Nullable Status minStatus, @Nullable Status maxStatus,
+                           SafeCommandStore.CommandFunction<T, T> map, T initialValue, T terminalValue)
+    {
+
+        for (D data : (testTimestamp == TestTimestamp.BEFORE ? commands.headMap(timestamp, false) : commands.tailMap(timestamp, false)).values())
+        {
+            TxnId txnId = loader.txnId(data);
```
dcapwell, line 123: I changed the ordering of this method due to cost to deserialize for `CommandsForKeys`, idea was to only deserialize what is needed and avoid touching deps until needed

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -371,9 +376,12 @@ private synchronized TopologyUpdate updateTopology(Node node, Snapshot prev, Top
                 RangesForEpochHolder rangesHolder = new RangesForEpochHolder();
                 ShardHolder shardHolder = new ShardHolder(supplier.create(nextId++, rangesHolder), rangesHolder);
                 rangesHolder.current = new RangesForEpoch(epoch, add, shardHolder.store);
-                // the first epoch we assume is either empty, or correctly initialised by whatever system is migrating
-                if (epoch == 1) bootstrapUpdates.add(() -> shardHolder.store.initialise(epoch, add));
-                else bootstrapUpdates.add(shardHolder.store.bootstrapper(node, add, newLocalTopology.epoch()));
+
+                Ranges.Partition partition = add.partition(range -> shouldBootstrap(node, prev.local, newLocalTopology, range));
```
aweisberg: I don't understand this. The partition will put all the ranges on the left or right based on `newLocalTopology.epoch` so why partition at all?

accord-core/src/main/java/accord/primitives/Ranges.java
```diff
@@ -227,4 +230,27 @@ else if (count == result.length)
         return construct(cachedRanges.completeAndDiscard(result, count));
     }
 
+    public Partition partition(Predicate<Range> test)
+    {
+        List<Range> left = new ArrayList<>();
+        List<Range> right = new ArrayList<>();
+        for (Range range : this)
+            (test.test(range) ? left : right).add(range);
```
aweisberg: Left and right is odd terminology when this is really filtering into true/false buckets for the predicate.

accord-core/src/main/java/accord/primitives/TxnId.java
```diff
@@ -35,6 +35,13 @@ public static TxnId fromBits(long msb, long lsb, Id node)
         return new TxnId(msb, lsb, node);
     }
 
+    public static TxnId fromTimestamp(Timestamp t)
```
aweisberg: This bypasses the `TxnId` vs `Timestamp` distinction for test code and makes it seem like they are the same thing. Maybe I am being to pedantic.

Since the only usage is `fromString` maybe add `TxnId.fromString` instead and have it create the `TxnId` from the `Timestamp`?

accord-core/src/main/java/accord/topology/Topology.java
```diff
@@ -258,16 +260,36 @@ private <P1> int[] subsetFor(Unseekables<?, ?> select, IndexedPredicate<P1> pred
...
                     if (ailim < ai)
-                        throw new IllegalArgumentException("Range not found for " + as.get(ailim));
+                    {
+                        switch (onUnknown)
+                        {
+                            default:
+                                throw new IllegalArgumentException("Unknown option: " + onUnknown);
+                            case REJECT: throw new IllegalArgumentException("Range not found for " + as.get(ailim));
+                            case IGNORE:
+                                bi = (int)(abi >>> 32);
```
aweisberg: Unused

accord-core/src/main/java/accord/topology/Topology.java
aweisberg: It kind of makes sense as a work around for the case you are describing, where is this code that tries to compare different epochs that fails?

It seems like this change is opting everything into the `IGNORE` behavior?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -230,16 +230,20 @@ public synchronized void onTopologyUpdate(Topology topology)
...
         List<Set<Id>> pendingSync = new ArrayList<>(current.pendingSyncComplete);
         Set<Id> alreadySyncd = Collections.emptySet();
         if (!pendingSync.isEmpty())
         {
-            EpochState currentEpoch = current.epochs[0];
-            if (current.epochs[0].syncComplete())
-                currentEpoch.markPrevSynced();
-            alreadySyncd = pendingSync.remove(0);
+            // if empty, then notified about an epoch from a peer before first epoch seen
+            if (current.epochs.length != 0)
```
aweisberg, line 240: Burn test fix?

aweisberg (changes requested):
Nothing major, some questions. The `Topology` change is the main thing I need to understand better since it is always opting into `IGNORE` at the moment.

accord-core/src/main/java/accord/topology/Topology.java
dcapwell: Went with ignore for all cases as I couldn’t find any caller that allowed the failure. At the dr office so can’t show,  it changing default and running AccordCQLTest you hit this right away

accord-core/src/main/java/accord/local/CommandStores.java
dcapwell: Ranges can have many Range, so this groups into 2 groups: those where the predicate are true, and those where they are false; the best example why is in C*

In C* we have two cases:
* key space is known, so getting new ranges; we need to bootstrap 
* key space is not known (CREATE KEYSPACE), so bootstrap is not needed or desired

accord-core/src/main/java/accord/primitives/Ranges.java
dcapwell: Have a better name?

dcapwell: Trues and falses?

accord-core/src/main/java/accord/primitives/TxnId.java
dcapwell: I think I can delete. I did it to test out DB crawling logic…

accord-core/src/main/java/accord/topology/TopologyManager.java
dcapwell: Burn didn’t hit this but C* did. Epoch 1 is known on  N1 and a request is sent to N2 before it learns of epoch 1… so this is an empty array

accord-core/src/main/java/accord/primitives/Ranges.java
dcapwell: was trying to find other APIs to see if there are better names... and best I see in JDK is

```
Map<Boolean, List<Integer>> groups = intList.stream().collect(Collectors.partitioningBy(s -> s > 6));
```

So rather than an object with true/false, this returns a `Map<Boolean, T>`

I don't mind mimicking this API if you think its best

accord-core/src/main/java/accord/topology/Topology.java
dcapwell: here is the exception

```
java.lang.IllegalArgumentException: Range not found for TokenKey{keyspace=demo_ks, key=-9079300900759362424}

	at accord.topology.Topology.subsetFor(Topology.java:266)
	at accord.topology.Topology.visitNodeForKeysOnceOrMore(Topology.java:337)
	at accord.topology.TopologyManager.withUnsyncedEpochs(TopologyManager.java:341)
	at accord.topology.TopologyManager.withUnsyncedEpochs(TopologyManager.java:310)
	at accord.coordinate.CoordinatePreAccept.<init>(CoordinatePreAccept.java:129)
	at accord.coordinate.CoordinateTransaction.<init>(CoordinateTransaction.java:49)
	at accord.coordinate.CoordinateTransaction.coordinate(CoordinateTransaction.java:54)
	at accord.local.Node.initiateCoordination(Node.java:447)
	at accord.local.Node.lambda$coordinate$9(Node.java:439)
	at accord.local.Node.withEpoch(Node.java:220)
	at accord.local.Node.coordinate(Node.java:439)
```

So, the use case is the following

1) create a new Keyspace w/ table
2) run a txn over that new table

The `org.apache.cassandra.distributed.test.accord.NewSchemaTest` test stresses this edge case to make sure the different epochs are supported

accord-core/src/main/java/accord/primitives/Ranges.java
```diff
@@ -227,4 +234,17 @@ else if (count == result.length)
         return construct(cachedRanges.completeAndDiscard(result, count));
     }
 
+    public Map<Boolean, Ranges> partitioningBy(Predicate<? super Range> test)
```
dcapwell, line 237: moved away from the `Partition` class as this actually is better the more I looked at it... `Partition` class didn't properly handle `this.isEmpty()`, but with a map we have a few states

1) map is empty
2) map(true)
3) map(false)
4) map(true, false)

With this change, we at least have the same signature as `Collectors.partitioningBy`

accord-maelstrom/src/test/java/accord/maelstrom/Runner.java
```diff
@@ -185,7 +185,14 @@ void run(String ... commands) throws IOException
         {
             logger.info("Seed {}; rerun with -D{}={}", seed, key, seed);
             RandomSource randomSource = new DefaultRandom(seed);
-            Runner.run(nodeCount, new StandardQueue.Factory(randomSource), randomSource::fork, factory, commands);
+            try
+            {
+                Runner.run(nodeCount, new StandardQueue.Factory(randomSource), randomSource::fork, factory, commands);
```
dcapwell, line 190: these tests are not deterministic, so same seed doesn't actually get the same results =(

accord-core/src/main/java/accord/topology/Topology.java
aweisberg: Huh, interesting consequence of having the table id in the MSB of the key.

accord-core/src/main/java/accord/primitives/Ranges.java
aweisberg: That WFM.

accord-core/src/main/java/accord/local/CommandStores.java
aweisberg: Ah, I missed the override that adds the contains check.

accord-core/src/main/java/accord/topology/Topology.java
aweisberg: Still unused FYI

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -333,38 +344,6 @@ private void forEach(Routable keyOrRange, Ranges slice, Consumer<RoutableKey> fo
         }
     }
 
-    @Override
-    protected void registerHistoricalTransactions(Deps deps)
```
dcapwell, line 337: moved this to safe command store... we can't handle this in `CommandStore` for C* unless we lazy load the dependencies (which requires https://github.com/apache/cassandra-accord/pull/27)

accord-core/src/main/java/accord/topology/Topology.java
dcapwell: in this case, `TokenKey` is scoped to the `keyspace`, so a new table in an existing keyspace is fine, but adding a new keyspace and touching it causes this.

accord-core/src/main/java/accord/topology/Topology.java
dcapwell: I have been holding off touching this class until we know what to actually do... let me debug this more and maybe ask benedict

accord-core/src/main/java/accord/topology/Topology.java
aweisberg: Ah right, I was concerned about this because it causes false conflicts across tables in a key space. Discussed it with Benedict and we might want to scope it differently at some point.

aweisberg (approved):
+1 TY

accord-core/src/main/java/accord/local/SafeCommandStore.java
```diff
@@ -155,6 +156,7 @@ <T> T mapReduce(Seekables<?, ?> keys, Ranges slice,
     NodeTimeService time();
     CommandStores.RangesForEpoch ranges();
     Timestamp maxConflict(Seekables<?, ?> keys, Ranges slice);
+    void registerHistoricalTransactions(Deps deps);
```
belliottsmith, line 159: I had intended to keep this to `CommandStore` to isolate the APIs - `SafeCommandStore` is for working with `Command` and `SafeCommand` whereas `CommandStore` is for internal book-keeping operations.

accord-core/src/main/java/accord/local/SaveStatus.java
```diff
@@ -74,6 +74,18 @@ public boolean hasBeen(Status status)
...
 
+    public boolean isComplete()
+    {
+        switch (this)
+        {
+            case Applied:
+            case Invalidated:
+                return true;
+            default:
+                return false;
```
belliottsmith, line 85: I would prefer not to use `default` properties, as it makes it harder to safely modify the state machine in future.

belliottsmith: Is this actually used?

accord-core/src/main/java/accord/local/SafeCommandStore.java
dcapwell: sorry, just saw this after merge...

In order to make the outcome of this method durable, it has to be in the safe store for C*, so felt it was best to move it there as that's were we mutate state normally

accord-core/src/main/java/accord/local/SaveStatus.java
dcapwell: it was, but then deleted in favor of something else; but forgot to purge this

PR #47, opened 2023-05-23
TCM/Bootstrap integration

bdeggleston:


accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+            for (long addEpoch = maxEpoch + 1; addEpoch <= epoch; addEpoch++)
+                epochs.add(createEpochState(addEpoch));
+
+            return epochs.get(idx);
+        }
+
+        public void receive(Topology topology)
+        {
+            long epoch = topology.epoch();
+            Invariants.checkState(lastReceived == epoch - 1 || epoch == 0 || lastReceived == 0);
```
dcapwell, line 143: can you add a useful error msg?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+
+        @VisibleForTesting
+        int size()
+        {
+            return epochs.size();
+        }
+
+        EpochState getOrCreate(long epoch)
+        {
+            Invariants.checkArgument(epoch > 0);
```
dcapwell, line 110: useful error msg plz

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    @Override
+    public synchronized Topology currentTopology()
+    {
+        return epochs.topologyFor(epochs.lastReceived);
+    }
+
+    @Override
+    public synchronized Topology getTopologyForEpoch(long epoch)
```
dcapwell, line 215: this is nullable here... so can be confusing and likely lead to NPEs; feel we should return `AsyncResult` instead

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+            return epochs.get(idx);
+        }
+
+        public void receive(Topology topology)
+        {
+            long epoch = topology.epoch();
+            Invariants.checkState(lastReceived == epoch - 1 || epoch == 0 || lastReceived == 0);
+            lastReceived = epoch;
+            EpochState state = getOrCreate(epoch);
+            if (state != null)
```
dcapwell, line 146: you are calling `getOrCreate`, so this can't be null right?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+        }
+
+        public void acknowledge(EpochReady ready)
+        {
+            long epoch = ready.epoch;
+            Invariants.checkState(lastAcknowledged == epoch - 1 || epoch == 0 || lastAcknowledged == 0);
+            lastAcknowledged = epoch;
+            EpochState state = getOrCreate(epoch);
+            state.reads = ready.reads;
+            state.acknowledged.setSuccess(null);
```
dcapwell, line 170: think this fails on double `set*` calls right?  For TCM do we have exactly once semantics?  Can it retry to make sure the notification is idempotent?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+            return getOrCreate(epoch).topology;
+        }
+
+        public void acknowledge(EpochReady ready)
+        {
+            long epoch = ready.epoch;
+            Invariants.checkState(lastAcknowledged == epoch - 1 || epoch == 0 || lastAcknowledged == 0);
+            lastAcknowledged = epoch;
+            EpochState state = getOrCreate(epoch);
+            state.reads = ready.reads;
```
dcapwell, line 169: there are no locks here (yet, still looking at this one class)... what is the expected threading model?  If this is single producer multi consumer, then the `setSuccess` should publish this pointer update, but access would need a volatile read first to fetch the latest state; but we do `synchronized` instead

I only ask due to this being `public`

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+        }
+
+        AsyncResult<Void> acknowledgeFuture(long epoch)
+        {
+            return getOrCreate(epoch).acknowledged;
+        }
+
+        void truncateUntil(long epoch)
+        {
+            Invariants.checkArgument(epoch <= maxEpoch());
```
dcapwell, line 180: can we add a useful error msg?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+        fetchTopologyInternal(epoch);
+    }
+
+    protected abstract void localSyncComplete(Topology topology);
+
+    @Override
+    public synchronized void acknowledgeEpoch(EpochReady ready)
+    {
+        ready.metadata.addCallback(() -> epochs.acknowledge(ready));
+        ready.coordination.addCallback(() ->  localSyncComplete(epochs.getOrCreate(ready.epoch).topology));
```
dcapwell, line 237: this only triggers if the result was success, but should we have a state for failed bootstrap?  If we failed, what happens if a remote peer uses this epoch?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+
+        fetchTopologyInternal(epoch);
+    }
+
+    protected abstract void localSyncComplete(Topology topology);
+
+    @Override
+    public synchronized void acknowledgeEpoch(EpochReady ready)
+    {
+        ready.metadata.addCallback(() -> epochs.acknowledge(ready));
```
dcapwell, line 236: the code doesn't mark this as `@Nullable` but `accord.local.CommandStore#bootstrapper` sets this to `null`; which could lead to a NPE?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+    public synchronized AsyncResult<Void> reportTopology(Topology topology, boolean startSync)
+    {
+        long lastReceived = epochs.lastReceived;
+        if (topology.epoch() <= lastReceived)
+            return AsyncResults.success(null);
+
+        if (lastReceived > 0 && topology.epoch() > lastReceived + 1)
+        {
+            fetchTopologyForEpoch(lastReceived + 1);
+            epochs.receiveFuture(lastReceived + 1).addCallback(() -> reportTopology(topology, startSync));
```
dcapwell, line 252: if `latestRecieved = 3` and `epoch() = 5`, then this would "fetch" `epoch = 4` but not `epoch = 5`?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+            return AsyncResults.success(null);
```
dcapwell, line 253: this API is said void, but every instance is not async and always success... why is this async?  Shouldn't it be based off the epoch being acked?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+    public synchronized void acknowledgeEpoch(EpochReady ready)
+    {
+        ready.metadata.addCallback(() -> epochs.acknowledge(ready));
+        ready.coordination.addCallback(() ->  localSyncComplete(epochs.getOrCreate(ready.epoch).topology));
+    }
+
+    protected void topologyUpdatePreListenerNotify(Topology topology) {}
+    protected void topologyUpdatePostListenerNotify(Topology topology) {}
+
+    public synchronized AsyncResult<Void> reportTopology(Topology topology, boolean startSync)
```
dcapwell, line 243: why have `boolean startSync`?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+        epochs.truncateUntil(epoch);
+    }
+
+    public synchronized AsyncResult<Void> epochReady(long epoch)
+    {
+        EpochState state = epochs.getOrCreate(epoch);
+        if (state.reads != null)
+            return state.reads;
+
+        return state.acknowledged.flatMap(r -> state.reads).beginAsResult();
```
dcapwell, line 304: maybe return `AsyncChain` to avoid `beginAsResult`?

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,278 @@
...
+        {
+            this.id = id;
+            this.ranges = ranges;
+        }
+
+        @Override
+        public int hashCode()
+                         {
+                            return id.hashCode() + ranges.hashCode();
+                                                                     }
```
dcapwell, line 84: formatting

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,278 @@
...
+import accord.primitives.Timestamp;
+import accord.primitives.TxnId;
+import accord.utils.Invariants;
+import accord.utils.async.AsyncChains;
+import accord.utils.async.AsyncResult;
+import accord.utils.async.AsyncResults;
+
+import static accord.primitives.Routables.Slice.Minimal;
+
+public abstract class AbstractFetchCoordinator extends FetchCoordinator
```
dcapwell, line 51: this class looks like a copy/paste of `accord.impl.list.ListStore.SyncCoordinator`?

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,278 @@
...
+                {
+                    fail(to, new RuntimeException(reply.toString()));
+                    inflight.remove(key).cancel();
+                    switch ((ReadData.ReadNack) reply)
+                    {
+                        default: throw new AssertionError("Unhandled enum");
+                        case Invalid:
+                        case Redundant:
+                        case NotCommitted:
+                            throw new AssertionError();
```
dcapwell, line 136: nit, can you show the enum value in the error msg?

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,278 @@
...
+            {
+                if (!reply.isOk())
+                {
+                    fail(to, new RuntimeException(reply.toString()));
+                    inflight.remove(key).cancel();
+                    switch ((ReadData.ReadNack) reply)
+                    {
+                        default: throw new AssertionError("Unhandled enum");
+                        case Invalid:
+                        case Redundant:
```
dcapwell, line 134: trying to wrap my head around this... doesn't this just mean concurrent txn (original coordinator + recovery coordinator)?  So this is just saying that "we are reusing message reply, and these states are not valid for us" right?

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,278 @@
...
+    protected synchronized void success(Node.Id to, Ranges ranges)
+    {
+        super.success(to, ranges);
+    }
+
+    @Override
+    protected synchronized void fail(Node.Id to, Ranges ranges, Throwable failure)
+    {
+        super.fail(to, ranges, failure);
+    }
```
dcapwell, line 193: why override to add `synchronized`?

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,278 @@
...
+                            if (f == null) result.setSuccess(ranges);
+                            else result.setFailure(f);
+                        });
+    }
+
+    @Override
+    public void start()
+                      {
+                         super.start();
+                                       }
```
dcapwell, line 216: formatting

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,278 @@
...
+    {
+        return result;
+    }
+
+    @Override
+    protected void onDone(Ranges success, Throwable failure)
+    {
+        if (success.isEmpty()) result.setFailure(failure);
+        else if (persisting.isEmpty()) result.setSuccess(null);
+        else AsyncChains.reduce(persisting, (a, b)-> null)
```
dcapwell, line 205:
```suggestion
        else AsyncChains.reduce(persisting, (a, b) -> null)
```

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,278 @@
...
+        public FetchRequest(long sourceEpoch, TxnId syncId, Ranges ranges, PartialDeps partialDeps, PartialTxn partialTxn)
+        {
+            super(ranges, sourceEpoch, Status.Applied, partialDeps, Timestamp.MAX, syncId, partialTxn);
+            this.partialDeps = partialDeps;
+        }
+
+        @Override
+        protected void readComplete(CommandStore commandStore, Data result, Ranges unavailable)
+        {
+            commandStore.execute(PreLoadContext.empty(), safeStore -> {
```
dcapwell, line 237: know this is copy/paste, but we have `accord.local.AgentExecutor#execute` now, so could avoid the `PreLoadContext.empty()` and `.begin(node.agent());` boiler plate

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -358,7 +358,7 @@ private synchronized TopologyUpdate updateTopology(Node node, Snapshot prev, Top
                 bootstrapUpdates.add(shard.store.interruptBootstraps(epoch, newRanges.currentRanges()));
             }
             // TODO (desired): only sync affected shards
-            if (epoch > 1)
+            if (epoch > 1 && startSync)
```
dcapwell, line 367: rewritten in https://issues.apache.org/jira/browse/CASSANDRA-18519, the logic is now

```
Map<Boolean, Ranges> partitioned = add.partitioningBy(range -> shouldBootstrap(node, prev.local, newLocalTopology, range));
if (partitioned.containsKey(true))
    bootstrapUpdates.add(shardHolder.store.bootstrapper(node, partitioned.get(true), newLocalTopology.epoch()));
if (partitioned.containsKey(false))
    bootstrapUpdates.add(() -> shardHolder.store.initialise(epoch, partitioned.get(false)));
```

dcapwell: I feel the `startSync` is to work around the bug where we don't want to bootstrap when a keyspace is added?  If so I added a new `accord.local.CommandStores#shouldBootstrap` for that... 

One issue to deal with is that we could have a new keyspace at the same time we see a range movement; so we need to bootstrap the range movement but not the new keyspace

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -150,9 +149,10 @@ public Node(Id id, MessageSink messageSink, ConfigurationService configService,
         configService.registerListener(this);
     }
 
+    // FIXME: remove, only used byy Maelstrom
```
dcapwell, line 152: can we delete Maelstrom!??!?!1

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -174,6 +176,16 @@ public long nextEpoch()
             return current().epoch + 1;
         }
 
+        public long minEpoch()
+        {
+            return currentEpoch - epochs.length + 1;
```
dcapwell, line 181: if empty isn't this `0 - 0 + 1 = 1`, and should be `0`?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -223,14 +237,14 @@ public TopologyManager(TopologySorter.Supplier sorter, Id node)
...
-        this.epochs = new Epochs(new EpochState[0]);
+        this.epochs = Epochs.EMPTY;
     }
 
     public synchronized void onTopologyUpdate(Topology topology)
     {
         Epochs current = epochs;
 
-        checkArgument(topology.epoch == current.nextEpoch());
+        checkArgument(topology.epoch == current.nextEpoch() || epochs == Epochs.EMPTY);
```
dcapwell: useful error msg would be nice =D

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -270,6 +284,22 @@ public void onEpochSyncComplete(Id node, long epoch)
         epochs.syncComplete(node, epoch);
     }
 
+    public synchronized void truncateTopologyUntil(long epoch)
+    {
+        Epochs current = epochs;
+        checkArgument(current.epoch() >= epoch);
```
dcapwell, line 295:
```suggestion
        checkArgument(current.epoch() >= epoch, "Unable to truncate; epoch %d is > current epoch %d", epoch , current.epoch());
```

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -270,6 +284,22 @@ public void onEpochSyncComplete(Id node, long epoch)
...
+
+        if (current.minEpoch() >= epoch)
+            return;
+
+        int newLen = current.epochs.length - (int) (epoch - current.minEpoch());
+        Invariants.checkState(current.epochs[newLen - 1].syncComplete());
```
dcapwell, line 301:
```suggestion
        Invariants.checkState(current.epochs[newLen - 1].syncComplete(), "Epoch %d's sync is not complete", current.epochs[newLen - 1].epoch);
```

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -270,6 +284,22 @@ public void onEpochSyncComplete(Id node, long epoch)
...
+
+        EpochState[] nextEpochs = new EpochState[newLen];
+        System.arraycopy(current.epochs, 0, nextEpochs, 0, newLen);
+        epochs = new Epochs(nextEpochs, current.pendingSyncComplete, current.futureEpochFutures);
```
dcapwell, line 305: nit: `java.util.Arrays#copyOfRange(U[], int, int, java.lang.Class<? extends T[]>)`?

accord-core/src/test/java/accord/messages/PreAcceptTest.java
```diff
@@ -196,7 +196,8 @@ void multiKeyTimestampUpdate() throws ExecutionException
             messageSink.assertHistorySizes(0, 1);
             Assertions.assertEquals(ID3, messageSink.responses.get(0).to);
             PartialDeps expectedDeps = new PartialDeps(Ranges.of(range(0, 12)), KeyDeps.NONE, RangeDeps.NONE);
-            Assertions.assertEquals(new PreAccept.PreAcceptOk(txnId2, Timestamp.fromValues(1, 110, ID1), expectedDeps),
+            Timestamp expectedTs = Timestamp.fromValues(1, 110, ID1).withExtraFlags(txnId2.flags());
```
dcapwell, line 199: what change caused this?  I get the following locally (only changed this test, nothing else)

```
Expected :PreAcceptOk{txnId:[1,50,2,3], witnessedAt:[1,110,2,1], deps:[Range(0, 12]]:{}, {}}
Actual   :PreAcceptOk{txnId:[1,50,2,3], witnessedAt:[1,110,0,1], deps:[Range(0, 12]]:{}, {}}
```

dcapwell (commented):
did first pass, will look closer at the tests next

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
dcapwell:
```suggestion
            Invariants.checkArgument(epoch > 0, "Epoch must be positive but given %d", epoch);
```

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
dcapwell:
```suggestion
                        default: throw new AssertionError("Unhandled enum");
                        case Invalid:
                        case Redundant:
                        case NotCommitted:
                            throw new AssertionError(String.format("Unexpected reply: %s", reply));
```

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,278 @@
...
+            @Override
+            public void onSuccess(Node.Id from, ReadData.ReadReply reply)
+            {
+                if (!reply.isOk())
+                {
+                    fail(to, new RuntimeException(reply.toString()));
+                    inflight.remove(key).cancel();
+                    switch ((ReadData.ReadNack) reply)
+                    {
+                        default: throw new AssertionError("Unhandled enum");
```
dcapwell, line 132:
```suggestion
                        default: throw new AssertionError("Unhandled enum: " + reply);
```

accord-core/src/main/java/accord/local/Node.java
dcapwell:
```suggestion
    // TODO (cleanup, testing): remove, only used by Maelstrom
```

accord-core/src/main/java/accord/topology/TopologyManager.java
dcapwell:
```suggestion
        checkArgument(topology.epoch == current.nextEpoch() || epochs == Epochs.EMPTY, "Unexpected epoch %d; expected %d", topology.epoch, current.nextEpoch();
```

dcapwell (commented):
Overall LGTM, no big comments.  Think there are a few bugs, and it feels like `startSync` is trying to address when to bootstrap, and if so I changed that logic in https://issues.apache.org/jira/browse/CASSANDRA-18519 to allow C* to override

bdeggleston:
merged manually

PR #49, opened 2023-06-07
Bootstrap tcm integration

bdeggleston:


accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+
+        @VisibleForTesting
+        int size()
+        {
+            return epochs.size();
+        }
+
+        EpochState getOrCreate(long epoch)
+        {
+            Invariants.checkArgument(epoch > 0);
```
dcapwell:
```suggestion
            Invariants.checkArgument(epoch > 0, "Epoch must be positive but given %d", epoch);
```

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+            for (long addEpoch = maxEpoch + 1; addEpoch <= epoch; addEpoch++)
+                epochs.add(createEpochState(addEpoch));
+
+            return epochs.get(idx);
+        }
+
+        public void receive(Topology topology)
+        {
+            long epoch = topology.epoch();
+            Invariants.checkState(lastReceived == epoch - 1 || epoch == 0 || lastReceived == 0);
```
dcapwell:
```suggestion
            Invariants.checkState(lastReceived == epoch - 1 || epoch == 0 || lastReceived == 0, "Epoch %d != %d + 1", epoch, lastReceived);
```

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+            lastReceived = epoch;
+            EpochState state = getOrCreate(epoch);
```
dcapwell, line 150: you are calling `getOrCreate`, so this can't be null right?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+
+        Topology topologyFor(long epoch)
+        {
+            return getOrCreate(epoch).topology;
+        }
+
+        public void acknowledge(EpochReady ready)
+        {
+            long epoch = ready.epoch;
+            Invariants.checkState(lastAcknowledged == epoch - 1 || epoch == 0 || lastAcknowledged == 0);
```
dcapwell:
```suggestion
            Invariants.checkState(lastAcknowledged == epoch - 1 || epoch == 0 || lastAcknowledged == 0, "Epoch %d != %d + 1", epoch, lastAcknowledged);
```

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+            lastAcknowledged = epoch;
+            EpochState state = getOrCreate(epoch);
+            state.reads = ready.reads;
```
dcapwell, line 173: there are no locks here (yet, still looking at this one class)... what is the expected threading model?  If this is single producer multi consumer, then the `setSuccess` should publish this pointer update, but access would need a volatile read first to fetch the latest state; but we do `synchronized` instead

I only ask due to this being `public`

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+            state.acknowledged.setSuccess(null);
```
dcapwell, line 174: think this fails on double `set*` calls right?  For TCM do we have exactly once semantics?  Can it retry to make sure the notification is idempotent?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+        }
+
+        AsyncResult<Void> acknowledgeFuture(long epoch)
+        {
+            return getOrCreate(epoch).acknowledged;
+        }
+
+        void truncateUntil(long epoch)
+        {
+            Invariants.checkArgument(epoch <= maxEpoch());
```
dcapwell:
```suggestion
            Invariants.checkArgument(epoch <= maxEpoch(), "epoch %d > %d", epoch, maxEpoch());
```

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    @Override
+    public synchronized Topology currentTopology()
+    {
+        return epochs.topologyFor(epochs.lastReceived);
+    }
+
+    @Override
+    public synchronized Topology getTopologyForEpoch(long epoch)
```
dcapwell, line 219: this is nullable here... so can be confusing and likely lead to NPEs; feel we should return `AsyncResult` instead

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+        fetchTopologyInternal(epoch);
+    }
+
+    protected abstract void localSyncComplete(Topology topology);
+
+    @Override
+    public synchronized void acknowledgeEpoch(EpochReady ready)
+    {
+        ready.metadata.addCallback(() -> epochs.acknowledge(ready));
+        ready.coordination.addCallback(() ->  localSyncComplete(epochs.getOrCreate(ready.epoch).topology));
```
dcapwell: this only triggers if the result was success, but should we have a state for failed bootstrap?  If we failed, what happens if a remote peer uses this epoch?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    protected void topologyUpdatePreListenerNotify(Topology topology) {}
+    protected void topologyUpdatePostListenerNotify(Topology topology) {}
+
+    public synchronized AsyncResult<Void> reportTopology(Topology topology, boolean startSync)
```
dcapwell: why have `boolean startSync`?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+    {
+        long lastReceived = epochs.lastReceived;
+        if (topology.epoch() <= lastReceived)
+            return AsyncResults.success(null);
+
+        if (lastReceived > 0 && topology.epoch() > lastReceived + 1)
+        {
+            fetchTopologyForEpoch(lastReceived + 1);
+            epochs.receiveFuture(lastReceived + 1).addCallback(() -> reportTopology(topology, startSync));
```
dcapwell, line 266: if `latestRecieved = 3` and `epoch() = 5`, then this would "fetch" `epoch = 4` but not `epoch = 5`?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+            return AsyncResults.success(null);
```
dcapwell: this API is said void, but every instance is not async and always success... why is this async?  Shouldn't it be based off the epoch being acked?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+        epochs.truncateUntil(epoch);
+    }
+
+    public synchronized AsyncResult<Void> epochReady(long epoch)
+    {
+        EpochState state = epochs.getOrCreate(epoch);
+        if (state.reads != null)
+            return state.reads;
+
+        return state.acknowledged.flatMap(r -> state.reads).beginAsResult();
```
dcapwell: maybe return `AsyncChain` to avoid `beginAsResult`?

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+import accord.primitives.Timestamp;
+import accord.primitives.TxnId;
+import accord.utils.Invariants;
+import accord.utils.async.AsyncChains;
+import accord.utils.async.AsyncResult;
+import accord.utils.async.AsyncResults;
+
+import static accord.primitives.Routables.Slice.Minimal;
+
+public abstract class AbstractFetchCoordinator extends FetchCoordinator
```
dcapwell, line 53: this class looks like a copy/paste of `accord.impl.list.ListStore.SyncCoordinator`?

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+        {
+            this.id = id;
+            this.ranges = ranges;
+        }
+
+        @Override
+        public int hashCode()
+                         {
+                            return id.hashCode() + ranges.hashCode();
+                                                                     }
```
dcapwell: formatting

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+            @Override
+            public void onSuccess(Node.Id from, ReadData.ReadReply reply)
+            {
+                if (!reply.isOk())
+                {
+                    fail(to, new RuntimeException(reply.toString()));
+                    inflight.remove(key).cancel();
+                    switch ((ReadData.ReadNack) reply)
+                    {
+                        default: throw new AssertionError("Unhandled enum");
```
dcapwell:
```suggestion
                        default: throw new AssertionError("Unhandled enum: " + reply);
```

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+                        case Invalid:
+                        case Redundant:
```
dcapwell, line 143: trying to wrap my head around this... doesn't this just mean concurrent txn (original coordinator + recovery coordinator)?  So this is just saying that "we are reusing message reply, and these states are not valid for us" right?

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+                        case NotCommitted:
+                            throw new AssertionError();
```
dcapwell:
```suggestion
                            throw new AssertionError(String.format("Unexpected reply: %s", reply));
```

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+    protected synchronized void success(Node.Id to, Ranges ranges)
+    {
+        super.success(to, ranges);
+    }
+
+    @Override
+    protected synchronized void fail(Node.Id to, Ranges ranges, Throwable failure)
+    {
+        super.fail(to, ranges, failure);
+    }
```
dcapwell: why override to add `synchronized`?

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+    {
+        return result;
+    }
+
+    @Override
+    protected void onDone(Ranges success, Throwable failure)
+    {
+        if (success.isEmpty()) result.setFailure(failure);
+        else if (persisting.isEmpty()) result.setSuccess(Ranges.EMPTY);
+        else AsyncChains.reduce(persisting, (a, b)-> null)
```
dcapwell:
```suggestion
        else AsyncChains.reduce(persisting, (a, b) -> null)
```

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+                            if (f == null) result.setSuccess(ranges);
+                            else result.setFailure(f);
+                        });
+    }
+
+    @Override
+    public void start()
+                      {
+                         super.start();
+                                       }
```
dcapwell: formatting

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -364,7 +364,7 @@ private synchronized TopologyUpdate updateTopology(Node node, Snapshot prev, Top
                 bootstrapUpdates.add(shard.store.interruptBootstraps(epoch, newRanges.currentRanges()));
             }
             // TODO (desired): only sync affected shards
-            if (epoch > 1)
+            if (epoch > 1 && startSync)
```
dcapwell: rewritten in https://issues.apache.org/jira/browse/CASSANDRA-18519, the logic is now

```
Map<Boolean, Ranges> partitioned = add.partitioningBy(range -> shouldBootstrap(node, prev.local, newLocalTopology, range));
if (partitioned.containsKey(true))
    bootstrapUpdates.add(shardHolder.store.bootstrapper(node, partitioned.get(true), newLocalTopology.epoch()));
if (partitioned.containsKey(false))
    bootstrapUpdates.add(() -> shardHolder.store.initialise(epoch, partitioned.get(false)));
```

dcapwell: I feel the `startSync` is to work around the bug where we don't want to bootstrap when a keyspace is added?  If so I added a new `accord.local.CommandStores#shouldBootstrap` for that... 

One issue to deal with is that we could have a new keyspace at the same time we see a range movement; so we need to bootstrap the range movement but not the new keyspace

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -150,9 +149,10 @@ public Node(Id id, MessageSink messageSink, ConfigurationService configService,
         configService.registerListener(this);
     }
 
+    // FIXME: remove, only used byy Maelstrom
```
dcapwell:
```suggestion
    // TODO (cleanup, testing): remove, only used byy Maelstrom
```

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -174,6 +186,16 @@ public long nextEpoch()
             return current().epoch + 1;
         }
 
+        public long minEpoch()
+        {
+            return currentEpoch - epochs.length + 1;
```
dcapwell, line 193: if empty isn't this `0 - 0 + 1 = 1`, and should be `0`?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -274,6 +299,22 @@ public void onEpochSyncComplete(Id node, long epoch)
         epochs.syncComplete(node, epoch);
     }
 
+    public synchronized void truncateTopologyUntil(long epoch)
+    {
+        Epochs current = epochs;
+        checkArgument(current.epoch() >= epoch);
```
dcapwell:
```suggestion
        checkArgument(current.epoch() >= epoch, "Unable to truncate; epoch %d is > current epoch %d", epoch , current.epoch());
```

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -274,6 +299,22 @@ public void onEpochSyncComplete(Id node, long epoch)
...
+
+        if (current.minEpoch() >= epoch)
+            return;
+
+        int newLen = current.epochs.length - (int) (epoch - current.minEpoch());
+        Invariants.checkState(current.epochs[newLen - 1].syncComplete());
```
dcapwell:
```suggestion
        Invariants.checkState(current.epochs[newLen - 1].syncComplete(), "Epoch %d's sync is not complete", current.epochs[newLen - 1].epoch);
```
```suggestion
        Invariants.checkState(current.epochs[newLen - 1].syncComplete());
```

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -274,6 +299,22 @@ public void onEpochSyncComplete(Id node, long epoch)
...
+
+        EpochState[] nextEpochs = new EpochState[newLen];
+        System.arraycopy(current.epochs, 0, nextEpochs, 0, newLen);
+        epochs = new Epochs(nextEpochs, current.pendingSyncComplete, current.futureEpochFutures);
```
dcapwell: nit: `java.util.Arrays#copyOfRange(U[], int, int, java.lang.Class<? extends T[]>)`?

accord-core/src/test/java/accord/messages/PreAcceptTest.java
```diff
@@ -196,7 +196,8 @@ void multiKeyTimestampUpdate() throws ExecutionException
             messageSink.assertHistorySizes(0, 1);
             Assertions.assertEquals(ID3, messageSink.responses.get(0).to);
             PartialDeps expectedDeps = new PartialDeps(Ranges.of(range(0, 12)), KeyDeps.NONE, RangeDeps.NONE);
-            Assertions.assertEquals(new PreAccept.PreAcceptOk(txnId2, Timestamp.fromValues(1, 110, ID1), expectedDeps),
+            Timestamp expectedTs = Timestamp.fromValues(1, 110, ID1).withExtraFlags(txnId2.flags());
```
dcapwell, line 199: what change caused this?  I get the following locally (only changed this test, nothing else)

```
Expected :PreAcceptOk{txnId:[1,50,2,3], witnessedAt:[1,110,2,1], deps:[Range(0, 12]]:{}, {}}
Actual   :PreAcceptOk{txnId:[1,50,2,3], witnessedAt:[1,110,0,1], deps:[Range(0, 12]]:{}, {}}
```

dcapwell (changes requested):
Ported comments from https://github.com/apache/cassandra-accord/pull/47/files

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -461,6 +461,19 @@ void update(Ranges add)
         }
     }
 
+    public AsyncChain<Timestamp> maxAppliedFor(Seekables<?, ?> keysOrRanges, Ranges slice)
```
dcapwell:
```suggestion
    @Override
    public AsyncChain<Timestamp> maxAppliedFor(Seekables<?, ?> keysOrRanges, Ranges slice)
```

accord-core/src/main/java/accord/utils/ReducingIntervalMap.java
```diff
@@ -62,6 +62,7 @@ public ReducingIntervalMap(boolean inclusiveEnds, V value)
     @VisibleForTesting
     ReducingIntervalMap(boolean inclusiveEnds, K[] ends, V[] values)
     {
+        Invariants.checkArgument(ends.length == values.length - 1);
```
dcapwell:
```suggestion
        Invariants.checkArgument(ends.length == values.length - 1, "Length %d != %d - 1; %s -> %s", ends.length, values.length, Arrays.toString(ends), Arrays.toString(values));
```

accord-core/src/main/java/accord/utils/ReducingIntervalMap.java
```diff
@@ -109,6 +115,13 @@ public V get(K key)
         return values[idx];
     }
 
+    public K key(int idx)
+    {
+        if (idx < 0 || idx > size() - 1)
+            throw new IndexOutOfBoundsException();
```
dcapwell:
```suggestion
            throw new IndexOutOfBoundsException(String.format("%d < 0 or > %d - 1", idx, size()));
```

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
bdeggleston: The parent class synchronizes access, though EpochState is extensible by subclasses of AbstractConfigurationService

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
bdeggleston: it can, but those updates are de-duped further upstream. We also shouldn't be replacing previously set futures (`state.reads` - I'll add a check)

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
bdeggleston: it's only used for testing, so I'd like to avoid reworking it. I can add a Nullable annotation and some docs though?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
bdeggleston: as discussed on slack, this is for C* startup when replaying topology state. We don't want to initiate sync on an epoch where it's already been completed

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
bdeggleston: this is for epoch sync, not bootstrap, but both epoch sync and bootstrap will continue to retry. However I'm not sure if this survives a restart on the C* side. I _think_ they do, but I'll double check.

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
bdeggleston: That's right, it would fill in any gaps, and then re-submit topology at epoch 5.

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
bdeggleston: you're right, this should just be `void`, I think the returned result is vestigial

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
bdeggleston: epoch sync is handled correctly, but it looks like bootstrap will have the opposite problem: there's nothing to keep it from restarting each time C* restarts. I'll fix that, and may rework completion status handling to remove the need for a `startSync` flag that's passed in here.

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
bdeggleston: mostly yes. The non-list store specifc stuff became AbstractFetchCoordinator, from which non-abstract implementations were created for ListStore & C*

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
bdeggleston: basically, yeah. ReadReply does everything needed for fetch, but has some error states that don't apply

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
bdeggleston: I don't think I did that, but I imagine it would have been done to avoid imposing a concurrency strategy on the base class. We can move it up to FetchCoordinator if you'd like

dcapwell: > We can move it up to FetchCoordinator if you'd like

I would just drop as its not needed... the base class defines its concurrency, and the top level has to work with it... adding `synchronized` doesn't make things safer

accord-core/src/main/java/accord/local/CommandStores.java
dcapwell: spoke In slack, this is for restarts to avoid work not needed

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+            if (epoch < minEpoch)
+            {
+                int prepend = Ints.checkedCast(minEpoch - epoch);
+                List<EpochState> next = new ArrayList<>(epochs.size() + prepend);
+                for (long addEpoch=epoch; addEpoch<minEpoch; addEpoch++)
+                    next.add(createEpochState(addEpoch));
+                next.addAll(epochs);
+                epochs = next;
+                minEpoch = minEpoch();
+                Invariants.checkState(minEpoch == epoch);
```
dcapwell:
```suggestion
                Invariants.checkState(minEpoch == epoch, "Epoch %d != %d", epoch, minEpoch);
```

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
dcapwell: I don't think we actually do synchronize

```
public synchronized void acknowledgeEpoch(EpochReady ready)
{
    ready.metadata.addCallback(() -> epochs.acknowledge(ready));
```

`acknowledgeEpoch` is `synchronized`, but there is no reason to as it just adds callbacks to a thread safe type... the callbacks can run wherever so these methods are not thread safe atm

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+        {
+            return getOrCreate(epoch).acknowledged;
+        }
+
+        void truncateUntil(long epoch)
+        {
+            Invariants.checkArgument(epoch <= maxEpoch(), "epoch %d > %d", epoch, maxEpoch());
+            long minEpoch = minEpoch();
+            int toTrim = Ints.checkedCast(epoch - minEpoch);
+            if (toTrim <=0)
```
dcapwell:
```suggestion
            if (toTrim <= 0)
```

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -0,0 +1,337 @@
...
+
+import accord.api.ConfigurationService;
+import accord.local.Node;
+import accord.topology.Topology;
+import accord.utils.Invariants;
+import accord.utils.async.AsyncChain;
+import accord.utils.async.AsyncResult;
+import accord.utils.async.AsyncResults;
+
+public abstract class AbstractConfigurationService<EpochState extends AbstractConfigurationService.AbstractEpochState,
```
dcapwell, line 37: reading this class I really wonder if we should have a `CommandStore` for metadata (owns no ranges)... this logic would be fine with `CommandStore` but feel that it isn't consistent when run outside, so gets harder to be correct

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+        Key(Node.Id id, Ranges ranges)
+        {
+            this.id = id;
+            this.ranges = ranges;
+        }
+
+        @Override
+        public int hashCode()
+        {
+            return id.hashCode() + ranges.hashCode();
```
dcapwell: should use `java.util.Objects#hash`. that will become

```
(31  + id.hashCode()) * 31 + ranges.hashCode()
```

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+
+    protected abstract void onReadOk(Node.Id from, CommandStore commandStore, Data data, Ranges ranges);
+
+    @Override
+    public void contact(Node.Id to, Ranges ranges)
+    {
+        Key key = new Key(to, ranges);
+        inflight.put(key, starting(to, ranges));
+        Ranges ownedRanges = ownedRangesForNode(to);
+        Invariants.checkArgument(ownedRanges.containsAll(ranges));
```
dcapwell:
```suggestion
        Invariants.checkArgument(ownedRanges.containsAll(ranges), "Got a reply from %s for ranges %s, but owned ranges %s does not contain all the ranges", to, ranges, ownedRanges);
```

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+        PartialDeps partialDeps = syncPoint.waitFor.slice(ownedRanges, ranges);
+        node.send(to, new FetchRequest(syncPoint.sourceEpoch(), syncPoint.syncId, ranges, partialDeps, rangeReadTxn(ranges)), new Callback<ReadData.ReadReply>()
```
dcapwell, line 130: nit: can lower memory cost if `AbstractFetchCoordinator` impl `Callback`...

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+                inflight.remove(key).started(ok.maxApplied);
+                onReadOk(to, commandStore, ok.data, received);
+                // received must be invoked after submitting the persistence future, as it triggers onDone
+                // which creates a ReducingFuture over {@code persisting}
+            }
+
+            @Override
+            public void onFailure(Node.Id from, Throwable failure)
+            {
+                inflight.remove(key).cancel();
```
dcapwell, line 179: it would be nice if `cancel` could take the failure... so we could have a global handling of the throwable rather than the caller of this coordinator being the only one that could

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
dcapwell: Also, this class runs in `CommandStore`, so is single threaded... so `synchronized` doesn't help

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+
+    public FetchResult result()
+    {
+        return result;
+    }
+
+    @Override
+    protected void onDone(Ranges success, Throwable failure)
+    {
+        if (success.isEmpty()) result.setFailure(failure);
```
dcapwell:
```suggestion
        if (failure != null) result.setFailure(failure);
```

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -0,0 +1,288 @@
...
+                        });
+    }
+
+    @Override
+    public void start()
+    {
+        super.start();
+    }
+
+    void abort(Ranges abort)
```
dcapwell, line 215: maybe make `abstract` for now?  `ListFetchCoordinator` is the only usage atm and could no-op it for now

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -461,6 +461,19 @@ void update(Ranges add)
...
+    {
+        Seekables<?, ?> sliced = keysOrRanges.slice(slice, Minimal);
+        Timestamp timestamp = Timestamp.NONE;
+        for (GlobalCommand globalCommand : commands.values())
```
dcapwell: I believe this doesn't include range commands; they are in `accord.impl.InMemoryCommandStore#rangeCommands`...  I feel it's best to do `mapReduce` here as that would also work in the C* case and makes sure range and key are both supported.... this is what we do in C*

```
@Override
    public Timestamp maxConflict(Seekables<?, ?> keysOrRanges, Ranges slice)
    {
        return mapReduce(keysOrRanges, slice, (ts, accum) -> Timestamp.max(ts.max(), accum), Timestamp.NONE, null);
    }
```


I think you would need to do `accord.local.SafeCommandStore#mapReduce` for that though

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -128,6 +128,8 @@ public RangesForEpochHolder rangesForEpochHolder()
 
     public abstract void shutdown();
 
+    public abstract AsyncChain<Timestamp> maxAppliedFor(Seekables<?, ?> keysOrRanges, Ranges slice);
```
dcapwell: I feel this is best in `SafeCommandStore`, at least in the C* case you would need to do (logically)

```
return submit(PreloadContext.contextFor(keysOrRanges), safe -> safe.maxAppliedFor(keysOrRanges, Ranges slice);
```

accord-core/src/test/java/accord/impl/list/ListFetchCoordinator.java
```diff
@@ -0,0 +1,68 @@
...
+    public ListFetchCoordinator(Node node, Ranges ranges, SyncPoint syncPoint, DataStore.FetchRanges fetchRanges, CommandStore commandStore, ListStore listStore)
+    {
+        super(node, ranges, syncPoint, fetchRanges, commandStore);
+        this.listStore = listStore;
+    }
+
+    @Override
+    protected PartialTxn rangeReadTxn(Ranges ranges)
+    {
+        return new PartialTxn.InMemory(ranges, Txn.Kind.Read, ranges, new ListRead(unsafeStore -> unsafeStore, ranges, ranges), new ListQuery(Node.Id.NONE, Long.MIN_VALUE), null);
```
dcapwell:
```suggestion
        return new PartialTxn.InMemory(ranges, Txn.Kind.Read, ranges, new ListRead(Function.identity(), ranges, ranges), new ListQuery(Node.Id.NONE, Long.MIN_VALUE), null);
```

dcapwell (commented):
ok did a full review, going to look a bit closer at tests now

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
dcapwell: I do think we should remove and push this logic into `org.apache.cassandra.service.accord.AccordFetchCoordinator.IncomingStream#maybeListen`, that class defines its own concurrency rules, so it should be there rather than push into this class... in this case its not hard to do

```
            future.addCallback((state, fail) -> {
                synchronized (this)
                {
                    if (fail == null) success(from, Ranges.of(range));
                    else fail(from, Ranges.of(range), fail);
                }
            });
```

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
bdeggleston: oh that's right, good catch

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
bdeggleston: synchronizing the callback won't help since the base FetchCoordinator class isn't threadsafe, I'll submit it to the command store executor

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
bdeggleston: what would be the benefit there?

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
bdeggleston: I'd rather leave it since it has a todo and is meant to be implemented later

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
bdeggleston: I believe that's intentional, since we're only interested in commands that have been applied to disk and range txns are (currently) only used as markers/sync points/etc

accord-core/src/main/java/accord/local/CommandStore.java
bdeggleston: I don't agree. The safe store you'd be calling couldn't be used for maxApplied since it wouldn't have the keys in scope

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
dcapwell: > what would be the benefit there?

really its just knowing why we failed rather than going silent...  This is a network failure most likely so its really more likely to be `UNKNOWN`, which isn't 100% helpful... but yeah (unknown means "go check the FROM", but if its a network related issue, it might be safe to just bounce and try again?), I just tend to get squeamish when I see failures not checked

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
dcapwell: > synchronizing the callback won't help since the base FetchCoordinator class isn't threadsafe

I don't mind the command store as that makes this single threaded still, but I don't follow this point...  currently the success/fail methods are `synchronized` but thats in the base class... by pushing this logic into the C* implementation we can keep the same behavior by doing `synchronized (this)` in the callback... so that doesn't change the threading model.

single threaded > locking though, so cool with the proposed change

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
dcapwell: > range txns are (currently) only used as markers/sync points/etc

The only caller atm is `org.apache.cassandra.service.accord.AccordFetchCoordinator` (outside of tests), which is focused on streaming user data... so unless we have range write support (think we only plan for range read?), then the only data present should be from key commands... so excluding range commands should be fine in this context...

So, assuming that we do only care about what data was applied to user tables in this context, don't we also want to filter out non-write txn?  `org.apache.cassandra.service.accord.AccordCommandStore#maxApplied` does, but not `accord.impl.InMemoryCommandStore#maxAppliedFor`?

dcapwell: double checked with Ariel and at least for v1 we don't plan to have user level range reads/writes and his work is a metadata txn, so excluding range txn should still be fine after his work

dcapwell: > don't we also want to filter out non-write txn? 

guess it doesn't matter too much...

accord-core/src/main/java/accord/local/CommandStore.java
dcapwell: why would it not have the keys?  Isn't the job of `submit(PreLoadContext)` to make sure the keys are present (for key and ranges)?

I do think @belliottsmith didn't like me moving `maxConflict` to the safe store... but keeping it in `CommandStore` made C* more complex, as we don't have everything in-memory... so we are forced to push this logic into the executor (aka `SafeCommandStore`)

bdeggleston: because FetchRequest is a read, which can be executed in 2 stages. The first determines if the command is eligible to be executed. The second stage executes the read. If the command can be executed, the second stage is started. If it's not, the read waits for it to become eligible, then starts the second stage. In this case where it's run in 2 parts, only the second part needs the ranges. If the ranges were loaded as part of the first part we would be loading them but unable to use them and/or have out of date data when performing the second stage

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
belliottsmith: I think this is actually potentially a bug. Range transactions take the place of key transactions on aggregate, including for the purpose of bootstrap. Let's say you bootstrap from node A to node B, then from node B to node C, and let's say you don't receive any new transactions for some range in between. You'll have a maxApplied of NONE when shipping to C, rather than the range transaction that brought data from A to B.

I'm not sure if this would in practice lead to any isolation violations, but we probably do want to include range transactions as they (pretty much only) provide ordering guarantees with respect to other transactions right now.

belliottsmith: Hmm, thinking about this we probably want to propagate the maxApplied we received to any future owners, rather than the range transactions.

bdeggleston: I got mixed up here. `InMemoryCommandStore.commands` contains _all_ commands, key and range alike. CommandsForKey/RangeCommands are where they're separate, so this should be ok

belliottsmith: I was mistaken too though, I don’t think this is sufficient to propagate maxApplied reliably, as the range transaction that is used for bootstrap may be less than a propagated command. We probably want an additional register to store a timestamp to propagate. Perhaps I’ll fold that into my truncation patch as we also need one for tracking timestamps that have been truncated.

belliottsmith: Hmm, actually no. In this case we would not be "safe to read". So we're fine. Forget I said anything.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -461,6 +461,20 @@ void update(Ranges add)
         }
     }
 
+    @Override
+    public AsyncChain<Timestamp> maxAppliedFor(Seekables<?, ?> keysOrRanges, Ranges slice)
```
dcapwell: This implementation doesn't look consistent with the Cassandra version...

1) we don't filter for write txn, so include non-write txn, can add `if (command.txnId().rw() != Txn.Kind.Write) continue;`
2) we do `command.executeAt`, but Cassandra does `byId` aka `command.txnId()`

dcapwell: looking at the bootstrap code, I think we only care about write txn, so #1 makes sense to change in this class... but executeAt I think is more correct in this context, so prob need C* to be changed...

One thing that bugs me, the docs implies that the returned timestamp must be from the dependencies list, but I don't see any logic to enforce or care about that, so is it just a wording issue?

dcapwell (approved):
approving, though there is one open comment about in-memory/C* maxAppliedfor that needs to be addressed

PR #53, opened 2023-07-28
wip

jacek-lewandowski:


accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
```diff
@@ -253,7 +259,13 @@ void onNewEpoch(Topologies prevTopologies, Timestamp executeAt, List<PreAcceptOk
     public void accept(T success, Throwable failure)
     {
         if (failure instanceof CoordinationFailed)
+        {
             ((CoordinationFailed) failure).set(txnId, route.homeKey());
+            if (failure instanceof Preempted)
+                node.agent().metricsEventsListener().onPreempted(txnId);
+            else if (failure instanceof Timeout)
+                node.agent().metricsEventsListener().onTimeout(txnId);
+        }
```
henrikingo, line 284: This if-else construct seems to be recurring quite many times. Wouldn't it make sense to move this into the `metricsEventsListener` and here just call some generic method:

    node.agent().metricsEventListener().onFailure(failure);

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -343,7 +343,7 @@ static <T extends Timestamp> T merge(Timestamp a, Timestamp b, Constructor<T> co
     @Override
     public String toString()
     {
-        return "[" + epoch() + ',' + hlc() + ',' + flags() + ',' + node + ']';
+        return "[" + epoch() + ',' + hlc() + ',' + Integer.toBinaryString(flags()) + ',' + node + ']';
```
henrikingo, line 364: Nice :+1:

henrikingo (approved):

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
jacek-lewandowski: I didn't want to make the events listener be responsible for interpretation of the exception. This is cross-project API and the way how we learn about preemption or timeout may change over the time - for example, it may not be an exception at all, or we may move the reporting to the place where the exception is not thrown yet...

maedhroz: Makes sense. The only thing I'm wondering is if we want to know about invalidations as well...

jacek-lewandowski: Such a metric was not explicitly requested by it makes sense to add it

maedhroz (approved):

PR #54, opened 2023-07-31
Eviction purger

aweisberg:


accord-core/src/test/java/accord/impl/AbstractConfigurationServiceTest.java
```diff
@@ -43,6 +45,7 @@
 
 public class AbstractConfigurationServiceTest
 {
+    // C* uses this, so we can't use the jupiter assertions
```
dcapwell: does this mean C* excludes the junit 5 dependency?  If so, we might want to rethink this module setup to make this more explicit

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -459,45 +408,63 @@ void update(Ranges add)
         }
     }
 
-    class CFKLoader implements CommandLoader<TxnId>
+    static class CFKEntry extends TxnId
```
dcapwell, line 411: ill split this into a different PR once evict is in, but to make sure we actually test the cache-miss case I added a LRU~ish "cache" and dropped this class as its broken... we are pulling a globally modifiable reference within an immutable object (CommandsForKeys)?

accord-core/src/main/java/accord/local/CommonAttributes.java
```diff
@@ -84,6 +80,7 @@ public TxnId txnId()
 
         public Mutable txnId(TxnId txnId)
         {
+            Invariants.checkState(this != EMPTY_ATTRS && this != EMPTY_ATTRS_NONE_PARTIAL_DEPS, "Don't mutate the global empty instances");
```
dcapwell, line 83: in this case I kinda feel extending `CommonAttributes.Mutable` might be best... though its more code


```
private static class ImmutableAttributes extends CommonAttributes.Mutable
{
    priviate final String name; // to know if this was empty or empty w/ deps.none
```

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -685,25 +819,29 @@ public Ballot accepted()
 
     public static class Committed extends Accepted
     {
-        private final ImmutableSortedSet<TxnId> waitingOnCommit;
-        private final ImmutableSortedMap<Timestamp, TxnId> waitingOnApply;
+        public static final Committed EMPTY = new Committed(SaveStatus.Committed);
```
dcapwell: spoke to you about this in Slack, but don't think we should have this.  This is a C* concern for memory size tracking, but as this updates the public API for accord it means that this can't be changed once a release happens (as it would be a breaking change).

I feel C* should handle this by calling the constructor, or SerializerSupport in C* code, as this is only a concern there

accord-core/src/main/java/accord/primitives/AbstractKeys.java
```diff
@@ -229,42 +236,64 @@ public final long foldl(IndexedFoldToLong<? super K> fold, long param, long init
...
+            return toRoutingKeysFactory.apply(result, insertPos, false);
         }
         else
         {
-            result = new RoutingKey[1 + keys.length];
-            resultCount = copyToRoutingKeys(keys, 0, result, 0, insertPos);
-            if (resultCount == 0 || !withKey.equals(result[resultCount - 1]))
-                result[resultCount++] = withKey;
-            resultCount += copyToRoutingKeys(keys, insertPos, result, resultCount, keys.length - insertPos);
+            // TODO review this is a pretty terrible work around for the fact that you can't binarySearch these
```
dcapwell:
```suggestion
            // TODO (now): review this is a pretty terrible work around for the fact that you can't binarySearch these
```

this project uses a custom TODO syntax that tags comments... I have a script to let me keep track of these tags...

```
$ ./remaming-todos.sh | head
## efficiency: 95
| TODO (required, efficiency): callbacks should be associated with a commandStore for processing to avoid this                  | modules/accord/accord-core/src/main/java/accord/impl/SimpleProgressLog.java              |
| Set<Id> notAwareOfDurability; // TODO (easy, efficiency): use Agrona's IntHashSet as soon as Node.Id switches from long to... | modules/accord/accord-core/src/main/java/accord/impl/SimpleProgressLog.java              |
| Set<Id> notPersisted;         // TODO (easy, efficiency): use Agrona's IntHashSet as soon as Node.Id switches from long to... | modules/accord/accord-core/src/main/java/accord/impl/SimpleProgressLog.java              |
| TODO (desirable, efficiency): we can probably simplify things by requiring (empty) Apply messages to be sent also to the c... | modules/accord/accord-core/src/main/java/accord/impl/SimpleProgressLog.java              |
| TODO (desirable, efficiency): forward to local progress shard for processing (if known)                                       | modules/accord/accord-core/src/main/java/accord/impl/SimpleProgressLog.java              |
| TODO (desirable, efficiency): if we are co-located with the home shard, don't need to do anything unless we're in a           | modules/accord/accord-core/src/main/java/accord/impl/SimpleProgressLog.java              |
| TODO (find library, efficiency): this is obviously super inefficient, need some range map                                     | modules/accord/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java           |
| TODO (desired, efficiency): this can be made more efficient by batching by epoch                                              | modules/accord/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java           |
| TODO (find lib, efficiency): this is super inefficient, need to store Command in something queryable                          | modules/accord/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java           |
$ ./remaming-todos.sh -m tags
  95 efficiency
  66 low priority
  53 desired
  52 expected
  39 required
  25 testing
  24 consider
  14 easy
  13 now
  13 clarity
   8 desirable
   4 safety
   3 correctness
   3 API
   2 semantics
   2 liveness
   2 exceptions
   2 cleanup
   2 API consistency
   2 ?
   1 validation
   1 urgent
   1 topology
   1 soon
   1 simple
   1 progress
   1 performance
   1 later
   1 find library
   1 find lib
   1 exepected
   1 error handling
   1 behaviour
```

accord-core/src/main/java/accord/primitives/AbstractKeys.java
```diff
@@ -229,42 +236,64 @@ public final long foldl(IndexedFoldToLong<? super K> fold, long param, long init
...
+            // two incompatible types
+            RoutingKey[] result = new RoutingKey[keys.length];
+            for (int i = 0; i < keys.length; i++)
+            {
```
dcapwell: style: `{}` not needed, can remove

dcapwell (approved):
+1 to the changes on-top of eviction, leaving that patch out of review of this as thats a different PR

accord-core/src/test/java/accord/impl/AbstractConfigurationServiceTest.java
aweisberg: Yes C* uses it, it was added as part of TCM/Bootstrap. This commit already merged upstream in `trunk` and I am not sure why the PR is showing it? It should end up being an a no-op when merged.

accord-core/src/main/java/accord/local/CommonAttributes.java
aweisberg: I'll remove the EMPTY instances so it won't matter

accord-core/src/main/java/accord/local/Command.java
aweisberg: I'll remove them

accord-core/src/main/java/accord/primitives/AbstractKeys.java
aweisberg: I can remove the TODO, Benedict already took a look at it. I'll start using that syntax, thanks.

accord-core/src/main/java/accord/local/Command.java
aweisberg: Why is it a breaking change? The empty object returned is always measurable at runtime so if it changes so will the empty size. Empty size is for in memory footprint, serialization explicitly goes through the fields.

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -351,6 +350,7 @@ private synchronized TopologyUpdate updateTopology(Node node, Snapshot prev, Top
             Ranges removeRanges = subtracted.slice(current, Minimal);
             if (!removeRanges.isEmpty())
             {
+                // TODO (required): This is updating the a non-volatile field in the previous Snapshot, why modify it at all, even with volatile the guaranteed visibility is weak even with mutual exclusion
```
aweisberg, line 361: @belliottsmith Why does this update the previous snapshot in place? We can't guarantee anything in flight reads it so they might as well read it out of the current snapshot?

Can we make a copy to avoid the mutability issue?

belliottsmith: I'm not sure what you mean? This is the snapshot we're updating, inside the snapshot holder. This is safe published because all of the fields are final.

PR #56, opened 2023-08-15
Remote exceptions

aweisberg:


accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -99,12 +99,12 @@ public synchronized void accept(ReadNack reply, Throwable failure)
     {
         if (reply != null)
         {
-            node.reply(replyTo, replyContext, reply);
+            node.reply(replyTo, replyContext, reply, null);
```
dcapwell:
```suggestion
            node.reply(replyTo, replyContext, reply, failure);
```

you didn't check if failure is non-null, so pass that on to the sink to check

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -129,7 +129,9 @@ private void ack(@Nullable Ranges newUnavailable)
         // and prevents races where we respond before dispatching all the required reads (if the reads are
         // completing faster than the reads can be setup on all required shards)
         if (-1 == --waitingOnCount)
-            reply(this.unavailable, data);
+        {
```
dcapwell: style: remove `{}`

accord-core/src/main/java/accord/messages/WaitOnCommit.java
```diff
@@ -141,13 +147,13 @@ public Void reduce(Void o1, Void o2)
...
-        ack();
+        ack(failure);
     }
 
-    private void ack()
+    private void ack(@Nullable Throwable fail)
     {
         if (waitingOnUpdater.decrementAndGet(this) == -1)
-            node.reply(replyTo, replyContext, WaitOnCommitOk.INSTANCE);
+            node.reply(replyTo, replyContext, fail != null ? WaitOnCommitOk.INSTANCE : null, fail);
```
dcapwell: this doesn't look right, if we see a failure will we actually decrement completely?  Also wouldn't the exception be lost if it isn't the last event?

accord-core/src/test/java/accord/burn/BurnTestConfigurationService.java
```diff
@@ -66,7 +70,7 @@ public FetchTopologyRequest(long epoch)
         public void process(Node on, Node.Id from, ReplyContext replyContext)
         {
             Topology topology = on.configService().getTopologyForEpoch(epoch);
```
dcapwell, line 72: this can throw, so prob want a `try/catch`?

dcapwell: posted this in slack, prob want to change `Node.recieve` to have


```
scheduler.now(() -> {
            try
            {
                request.process(this, from, replyContext);
            }
            catch (Throwable t)
            {
                reply(from, replyContext, null, t);
            }
        });
```

that way you don't worry about my above comment

accord-core/src/test/java/accord/impl/list/ListAgent.java
```diff
@@ -46,11 +52,16 @@ public ListAgent(long timeout, Consumer<Throwable> onFailure, Consumer<Runnable>
     @Override
     public void onRecover(Node node, Result success, Throwable fail)
     {
+        if (fail != null)
+        {
+            checkState(success == null, "fail (%s) and success (%s) are both not null", fail, success);
+            // We don't really process errors for Recover here even though it is provided in the interface
```
dcapwell, line 58: we should report back still, else the future can hang forever right?

accord-core/src/test/java/accord/impl/list/ListRequest.java
```diff
@@ -110,38 +110,38 @@ public void accept(Result success, Throwable fail)
             // TODO (desired, testing): error handling
             if (success != null)
             {
-                node.reply(client, replyContext, (ListResult) success);
+                node.reply(client, replyContext, (ListResult) success, null);
```
dcapwell:
```suggestion
                node.reply(client, replyContext, (ListResult) success, fail);
```

didn't check `fail` yet

accord-core/src/test/java/accord/utils/MessageTask.java
```diff
@@ -88,7 +97,8 @@ public TaskRequest(NodeProcess process, String desc)
         @Override
         public void process(Node on, Node.Id from, ReplyContext replyContext)
         {
-            process.process(on, from, success -> on.reply(from, replyContext, success ? SUCCESS : FAILURE));
+            // TODO (review): This is kind of odd, no error to propagate?
+            process.process(on, from, success -> on.reply(from, replyContext, success ? SUCCESS : FAILURE, null));
```
dcapwell, line 100: only happens in `accord.burn.TopologyUpdates#notify` ```long nodeEpoch = node.epoch();
            if (nodeEpoch + 1 < update.epoch())
                onDone.accept(false);```
                
                in this single case, an exception at least gives more detail; which is nice... hit this condition earlier this morning so a human readable string would be nice

accord-core/src/main/java/accord/messages/WaitOnCommit.java
aweisberg: Good catch. I will refactor it to check in `accept` and only send the failure response if it hasn't already sent a regular response and log if somehow we have failure but already sent a response.

accord-core/src/test/java/accord/impl/list/ListAgent.java
aweisberg: We can't report back because without the `success` there is no `requestId` to respond with. We can't use uncaught exception handler because the exception isn't typically uncaught tit's actually been passed by the caller of `onRecover`. `onRecover` is an odd one because it's kind of informative, but it isn't the callback for recover it's just some notification the agent gets.

accord-core/src/main/java/accord/messages/ReadData.java
aweisberg: Yeah makes sense to have it assert.

accord-core/src/test/java/accord/impl/list/ListAgent.java
aweisberg: We can't report back because without the `success` there is no `requestId` to respond with. We can't use uncaught exception handler because the exception isn't typically uncaught tit's actually been passed by to the callback of `Recover`. `onRecover` is an odd one because it's kind of informative, but it isn't the callback for recover it's just some notification the agent gets.

I need to look at why `onRecover` is involved in sending the reply at all. Why doesn't it get done in the callback for `Recover`? Why does it exist at all?

belliottsmith: This is only for the ListAgent, which is used only for testing purposes. We keep enough information to issue a response directly to clients for testing so we get more responses to analyse the correctness of.

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -475,10 +477,15 @@ public void send(Id to, Request send)
...
     }
 
-    public void reply(Id replyingToNode, ReplyContext replyContext, Reply send)
+    public void reply(Id replyingToNode, ReplyContext replyContext, Reply send, Throwable failure)
     {
-        // TODO (usability, now): add Throwable as an argument so the error check is here, every single message gets this wrong causing a NPE here
-        if (send == null)
+        if (failure != null)
+        {
+            checkState(send == null, "fail (%s) and send (%s) are both not null", failure, send);
```
dcapwell:
```suggestion
            if (send != null)
                agent().onUncaughtException(new IllegalArgumentException(String.format("fail (%s) and send (%s) are both not null", failure, send)));
```

this situation would currently cause us to loose the real exception

accord-core/src/main/java/accord/messages/WaitOnCommit.java
```diff
@@ -141,13 +146,30 @@ public Void reduce(Void o1, Void o2)
...
     {
-        ack();
+        if (failure != null)
+        {
+            while (true)
+            {
+                int initialValue = waitingOnUpdater.get(this);
+                if (initialValue == -1)
+                {
+                    logger.error("Had error in WaitOnCommit, but already replied so can't send failure response", failure);
```
dcapwell:
```suggestion
                    node.agent().onUncaughtException(new IllegalStateException("Had error in WaitOnCommit, but already replied so can't send failure response", failure));
```

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -149,14 +149,16 @@ void read(SafeCommandStore safeStore, Timestamp executeAt, PartialTxn txn)
         Ranges unavailable = safeStore.ranges().unsafeToReadAt(executeAt);
 
         txn.read(safeStore, executeAt).begin((next, throwable) -> {
-            if (throwable != null)
+            // TODO (expected, exceptions): should send exception to client, and consistency handle/propagate locally
+            logger.trace("{}: read failed for {}: {}", txnId, unsafeStore, throwable);
+            synchronized (ReadData.this)
```
dcapwell: you are not checking that `throwable` is non-null, so this could lead to an exception at `fail.addSuppressed(throwable);`

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -149,14 +149,16 @@ void read(SafeCommandStore safeStore, Timestamp executeAt, PartialTxn txn)
         Ranges unavailable = safeStore.ranges().unsafeToReadAt(executeAt);
 
         txn.read(safeStore, executeAt).begin((next, throwable) -> {
-            if (throwable != null)
+            // TODO (expected, exceptions): should send exception to client, and consistency handle/propagate locally
+            logger.trace("{}: read failed for {}: {}", txnId, unsafeStore, throwable);
```
dcapwell: we don't actually know there is a read failure, we are a callback, so we see both success and failure!  we have to first check which one we are

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -149,14 +149,16 @@ void read(SafeCommandStore safeStore, Timestamp executeAt, PartialTxn txn)
...
-                logger.trace("{}: read failed for {}: {}", txnId, unsafeStore, throwable);
-                node.reply(replyTo, replyContext, ReadNack.Error);
+                if (fail == null)
+                    fail = throwable;
+                else
+                    fail.addSuppressed(throwable);
             }
-            else
-                readComplete(unsafeStore, next, unavailable);
+            readComplete(unsafeStore, next, unavailable);
```
dcapwell: if there is a failure do we really want to call `readComplete`?  Feel like we should try to halt right away as the coordinator has already seen our failure

accord-core/src/main/java/accord/messages/ReadData.java
aweisberg: Ouch yes.

accord-core/src/main/java/accord/utils/async/AsyncResults.java
```diff
@@ -333,14 +333,19 @@ public RunnableResult(Callable<V> callable)
...
-                trySetResult(callable.call(), null);
+                result = callable.call();
+                success = true;
             }
             catch (Throwable t)
             {
                 trySetResult(null, t);
             }
+            if (success)
+                trySetResult(result, null);
```
dcapwell: it would be good to document why the code is written this way, and we can avoid the boolean `success` if we rely on `return` when a failure is thrown

```
// There are two different type of exceptions: user function throws, listener throws.  To make sure this is clear,
            // make sure to catch the exception from the user function and set as failed, and let the listener exceptions bubble up.
            V call;
            try
            {
                call = callable.call();
            }
            catch (Throwable t)
            {
                setResult(null, t);
                return;
            }
            setResult(call, null);
```

accord-core/src/test/java/accord/impl/basic/TaskExecutorService.java
```diff
@@ -44,14 +44,19 @@ public Task(Callable<T> fn)
         @Override
         public void run()
         {
+            T result = null;
```
dcapwell: to avoid copy/paste maybe we should `extends AsyncResults.RunnableResult`?  

```
public static class Task<T> extends AsyncResults.RunnableResult<T> implements Pending, RunnableFuture<T>
    {
        public Task(Callable<T> fn)
        {
            super(fn);
        }
```

we can avoid the run method in this case completely

accord-core/src/main/java/accord/messages/ReadData.java
aweisberg: I left it calling `readComplete` (exceptionally or otherwise) so that it would eventually send the response once all command stores responded with `fail` containing all the errors that occurred if there are more than one.

I am not married to that decision though. Sending back the first error is fine.

`readComplete` manipulates the `waitingOn` bit set so I left it at that level. `readComplete` does have one override that actually does something different, but since everything always goes through `ack` and `ack` checks for failures I was going to call it good enough.

It might be cleaner to put the `waitingOn` update into `ack` and call `ack` directly. WDYT?

dcapwell: ill look closer before replying directly to your comments, but one thing I am thinking about; when we lived with everyone needing to handle exceptions, no one did... this patch is trying to improve that.... so for this to hope all extending classes "do the right thing" might be good much to ask.... but ill reread this code and reflect on your comment

aweisberg: The derived classes aren't the ones that send back the errors. `ack` does and that is `private`.

The derived classes can (and do) drop errors as part of their execution such as `FetchRequest.readComplete` which starts a new `AsyncChain` and adds a callback which passes the failure to `uncaughtExceptionHandler`.

I think the interface for listeners, task submission, and callbacks errors is possibly what is wrong. Every execution should occur with a context that specifies what the uncaught error handling should be, but you don't have to provide that context along with the thing you are executing. Or you do, but it's bespoke to that one thing you want to run (parameter to callback) rather than common across a variety of things that might be done to execute a request.

That would help with making it so can't get an error and then drop it or consume it in the wrong place because you would never handle it in the first place unless you specifically provided your own error handler rather than the one in the context. And when you do specify it should still hand off errors you don't want to deal with to the parent error handling context.

dcapwell: before this patch this is what I see

* accord.messages.ReadData#read would reply `Error` on the first exception and not call `readComplete`.  This would not update `waitingOn`, but more importantly it would not call `ack`; which means `if (-1 == --waitingOnCount)` will never be true!
* `accord.impl.AbstractFetchCoordinator.FetchRequest#readComplete` makes the read 2 operations:`txn.read()` and `store.maxAppliedFor(...).onSuccess(a -> {this.maxApplied = a; super.readComplete(...)})` (I overly simplified).  Even if a fetch had 2 reads (C* only has 1 atm), then we would still not see the ACK so would never complete the operation....
* `accord.messages.ReadTxnData#readComplete` documents that it should unregister its listener... but because we call `cancel` we will schedule removal of the listener anyways... so thats prob fine.

> I am not married to that decision though. Sending back the first error is fine.

I guess I would ask, what will the coordinator do differently with a list of failures?  If a single one exists we will do the same as if 100 existed...  so by only sending the first we have a smaller msg payload and can start handling the exception faster!  We could detect this case and just log locally that these exceptions are getting ignored....

So, in short, I am in favor of failing fast and not pushing handling exceptions properly to each impl.

dcapwell: > I think the interface for listeners, task submission, and callbacks errors is possibly what is wrong. Every execution should occur with a context that specifies what the uncaught error handling should be,

I agree with you on this point... AsyncChain and Futures are best when we "chain" them and centralize the error handling, but when we keep mixing and matching it gets hard to get right...

the common pattern is

```
// this is actually async, but to simplify the code I didn't do that
A value = null;
for store in stores.intersects(request):
  A local = store.process(request.map)
  value = request.reduce(value, local);
request.accept(value)
```

and in this case we try to add more state to make sure all `map` are called and `accept` is called (which is why we use `-1` in `-1 == --waitingOnCount`)...  

We could change this pattern to be

```
List<AsyncChain<A>> chains;
for store in stores.intersects(request):
  chains.add(store.process(request.map));
AsyncChains.reduce(chains, request.reduce)
```

this would allow us to start composing and chaining, rather than stoping chains, and forking new ones in different places...

Now, given how that is a structural change, I would prefer to do that as its own standalone work; adds far too much risk to this patch...

dcapwell (commented):
I am +1 once `ReadData` is resolved

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -153,7 +153,11 @@ void read(SafeCommandStore safeStore, Timestamp executeAt, PartialTxn txn)
             {
                 // TODO (expected, exceptions): should send exception to client, and consistency handle/propagate locally
                 logger.trace("{}: read failed for {}: {}", txnId, unsafeStore, throwable);
-                node.reply(replyTo, replyContext, ReadNack.Error);
+                synchronized (ReadData.this)
```
dcapwell: why this lock?  all impl delegate to `CommandStores` so wouldn't need locking.

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -53,6 +52,7 @@ public abstract class ReadData extends AbstractEpochRequest<ReadData.ReadNack>
     transient BitSet waitingOn;
     transient int waitingOnCount;
     transient Ranges unavailable;
+    transient Throwable fail;
```
dcapwell: looks like dead code now?

accord-core/src/main/java/accord/messages/ReadTxnData.java
```diff
@@ -231,8 +231,14 @@ protected synchronized void readComplete(CommandStore commandStore, @Nullable Da
     }
 
     @Override
-    protected void reply(@Nullable Ranges unavailable, @Nullable Data data)
+    protected void reply(@Nullable Ranges unavailable, @Nullable Data data, @Nullable Throwable fail)
     {
+        if (fail != null)
+        {
+            node.reply(replyTo, replyContext, null, fail);
```
dcapwell:
```suggestion
            state = State.RETURNED;
            node.reply(replyTo, replyContext, null, fail);
```

accord-core/src/main/java/accord/messages/ReadData.java
aweisberg: You are right it probably isn't necessary. I reflexively added it because the existing call to `cancel` is from a `synchronized` method, but the actual implementations of cancel don't require it to be `synchronized` and are probably also idempotent so it's fine if multiple responses and cancellations happen at once. I'll remove it.

accord-core/src/main/java/accord/messages/ReadTxnData.java
aweisberg: Shouldn't this also have the same illegal state logic as well? Maybe the fail handling should be in the switch below and handle obsolete as well.

accord-core/src/test/java/accord/impl/basic/NodeSink.java
```diff
@@ -82,4 +83,10 @@ public void reply(Id replyToNode, ReplyContext replyContext, Reply reply)
     {
         parent.add(self, replyToNode, Packet.getMessageId(replyContext), reply);
     }
+
+    @Override
+    public void replyWithFailure(Id replyingToNode, ReplyContext replyContext, Throwable failure)
+    {
+        reply(replyingToNode, replyContext, new FailureReply(failure));
```
dcapwell, line 90:
```suggestion
        lookup.apply(self).agent().onUncaughtException(failure);
        reply(replyingToNode, replyContext, new FailureReply(failure));
```

dcapwell (approved):

PR #57, opened 2023-08-15
cassandra 18451

dcapwell:


accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -269,6 +269,13 @@ public synchronized void reportTopology(Topology topology, boolean startSync)
         }
 
         long lastAcked = epochs.lastAcknowledged;
+        // TODO (now, review): lastAcked == 0, lastReceived = 2
+        // if we wait for epoch=1.acknowledge the test seems to wait forever... looks like burn test doesn't ack epoch=1
+        if (lastAcked == 0 && lastReceived > 0)
```
dcapwell, line 274: have to think about this.. when starting up with the topology we can reach received right away, but we don't see an ack until

```
public void acknowledgeEpoch(EpochReady ready)
    {
        ready.metadata.addCallback(() -> {
            synchronized (AbstractConfigurationService.this)
            {
                epochs.acknowledge(ready);
            }
        });
```

so `accord.local.CommandStore#bootstrapper` is the main thing blocking the metadata... so it would be possible that a new node joins the ring, has to bootstrap its state, then receive topology updates.... so waiting on `minEpoch` still makes sense to me...

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -172,7 +172,10 @@ public Node(Id id, MessageSink messageSink, ConfigurationService configService,
     // TODO (cleanup, testing): remove, only used by Maelstrom
     public AsyncResult<Void> start()
     {
-        return onTopologyUpdateInternal(configService.currentTopology(), false).metadata;
+        EpochReady ready = onTopologyUpdateInternal(configService.currentTopology(), false);
+        ready.coordination.addCallback(() -> this.topology.onEpochSyncComplete(id, topology.epoch()));
+        configService.acknowledgeEpoch(ready);
```
dcapwell: before the initial topology would not get an ack, but now it will

accord-core/src/main/java/accord/utils/async/AsyncResults.java
```diff
@@ -201,6 +201,12 @@ public Throwable failure()
                 throw new IllegalStateException("Result succeeded");
             return result.failure;
         }
+
+        @Override
+        public String toString()
+        {
+            return getClass().getSimpleName() + "{status=" + (isDone() ? isSuccess() ? "success" : "failure" : "pending") + "}";
```
dcapwell: im in the debugger all the time, so this is to make my life easier... I don't know if we log any results, but in a debugger there are many to look at so saves times

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -200,13 +202,17 @@ static void burn(RandomSource random, TopologyFactory topologyFactory, List<Id>
         });
 
         Supplier<LongSupplier> nowSupplier = () -> {
-            RandomSource jitter = random.fork();
-            // TODO (expected): jitter should be random walk with intermittent long periods
-            return () -> Math.max(0, delayQueue.nowInMillis() + (jitter.nextInt(10) - 5));
+            RandomSource forked = random.fork();
+            Gen.LongGen jitter = FrequentLargeRange.builder(forked)
+                                                   .raitio(1, 5)
```
dcapwell: every node will have a different `ratio`, all between 1 and 5%

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -200,13 +202,17 @@ static void burn(RandomSource random, TopologyFactory topologyFactory, List<Id>
...
+                                                   .small(50, 5000, TimeUnit.MICROSECONDS)
+                                                   .large(1, 10, TimeUnit.MILLISECONDS)
+                                                   .build();
+            return () -> Math.max(0, queue.nowInMillis() + jitter.nextLong(forked));
         };
 
         StrictSerializabilityVerifier strictSerializable = new StrictSerializabilityVerifier(keyCount);
-        SimulatedDelayedExecutorService globalExecutor = new SimulatedDelayedExecutorService(queue, agent, random.fork());
+        SimulatedDelayedExecutorService globalExecutor = new SimulatedDelayedExecutorService(queue, agent);
```
dcapwell, line 216: the delay logic was pushed to the queue, so no longer needs

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -230,37 +236,50 @@ static void burn(RandomSource random, TopologyFactory topologyFactory, List<Id>
...
-                return;
-
+        Runnable maybeTriggerNextRequest = () -> {
             if (requestIndex.get() < requests.length)
             {
                 int i = requestIndex.getAndIncrement();
                 starts[i] = clock.incrementAndGet();
-                queue.add(requests[i]);
+                // TODO (review): why did i add this again?
+                queue.addNoDelay(requests[i]);
```
dcapwell, line 251: there was a reason I added this, but its been too long I forgot... I can always revert and add back w/e the reason is detected again...

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -357,7 +376,8 @@ public static void main(String[] args)
     @Test
     public void testOne()
     {
-        run(ThreadLocalRandom.current().nextLong(), 1000);
+        run(-2862327792444379209L, 1000);
```
dcapwell: this seed was showing the ack race condition where epoch=3 ack before epoch=2; will revert

accord-core/src/test/java/accord/burn/random/FrequentLargeRange.java
```diff
@@ -18,59 +18,121 @@
...
+    public FrequentLargeRange(LongGen small, LongGen large, double ratio)
     {
-        Invariants.checkArgument(ratio > 0 && ratio <= 1);
         this.small = small;
         this.large = large;
-        this.ratio = ratio;
-        this.steps = (int) (1 / ratio);
-        this.lower = ratio * .8;
-        this.upper = ratio * 1.2;
+        this.runs = Gens.bools().runs(ratio);
```
dcapwell: thought about this after re-reviewing the code; `steps` is really `maxRuns` and because its `1 / ratio` it means that less frequent events have much higher maxRuns... it wouldn't be hard to expose this as a config so we can have more consistent maxRuns.... `1 / .01 == 100`, which feels like too many runs....

The behavior is not changed in this patch, but just something I thought about while reviewing.

accord-core/src/test/java/accord/impl/basic/Cluster.java
```diff
@@ -157,18 +151,6 @@ private void processNext(Object next)
             Packet deliver = (Packet) next;
             Node on = lookup.apply(deliver.dst);
 
-            // TODO (required, testing): random drop chance independent of partition; also port flaky connections etc. from simulator
-            // Drop the message if it goes across the partition
-            boolean drop = ((Packet) next).src.id >= 0 &&
```
dcapwell, line 162: logic now lives in NodeSink

accord-core/src/test/java/accord/impl/basic/Cluster.java
```diff
@@ -190,6 +172,12 @@ private void processNext(Object next)
         }
     }
 
+    public void notifyDropped(Node.Id from, Node.Id to, long id, Message message)
```
dcapwell, line 181: only here to keep the trace logging from before... im cool dropping.  We now have more detailed debug logs for messaging

accord-core/src/test/java/accord/impl/basic/NodeSink.java
```diff
@@ -33,11 +44,20 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static accord.impl.basic.Packet.SENTINEL_MESSAGE_ID;
+import static java.util.concurrent.TimeUnit.SECONDS;
 
 public class NodeSink implements MessageSink
 {
+    private static final boolean DEBUG = false;
+    private enum Action {DELIVER, DROP, DROP_PARTITIONED, FAILURE}
```
dcapwell: in follow up patches I add client actions as well, so we get a more complete view

dcapwell: hit a fault where we have 2 txn that got lost, so added what I was talking about locally to improve debugging...

belliottsmith:
I've rebased on top of this branch for https://github.com/apache/cassandra-accord/pull/58 which also integrates my own related changes and includes some suggestions to rationalise some of our approaches to randomisation. It also addresses a number of bugs (unrelated to this patch). I suggest merging the two pieces of work together, if that works for you?

dcapwell:
> I suggest merging the two pieces of work together, if that works for you?

@belliottsmith I would strongly prefer we keep these decoupled and not couple merge of this work with #58.  This patch has been waiting for over 4 months and I have an active backlog that depends on this work.

I took a quick skim of your changes in #58 and saw that its a mostly clean (though stale) rebase with 2 very specific changes (remove run bias in favor of uniform bias for `FrequentLargeRange`, and extract `NetworkSink.Action` out into `Cluster`); those changes should not be in conflict with follow up work, so it should be low effort for you to maintain.

belliottsmith:
Ok, fastest way to merge in that case is to revert the usage of Gen<X> to XSupplier, and to move `Gen.bools().run()` to `RandomSource` and give it a name like `biasedRepeatedRunBools()` such as here: https://github.com/apache/cassandra-accord/pull/58/files#diff-d9920227d4d5d5d5f60bedd83ceab1d1af0c19999c5eae606be008d4dc00bfc3R57

I would prefer we addressed the bugs alongside increasing their frequency, but this should at least hopefully avoid us bikeshedding indefinitely.

dcapwell:
@belliottsmith pushed the feedback; 2 commits

```
commit b9c8feb80442157bcd3d2d6470e33bd35811ccaa (HEAD -> eviction_CASSANDRA-18451, origin/eviction_CASSANDRA-18451)
Author: David Capwell <dcapwell@gmail.com>
Date:   Wed Aug 16 14:56:22 2023 -0700

    feedback: renamed runs to biasedRepeatingRuns

commit 49da37997ab4b1406c0069c9f387ab025d8990bb
Author: David Capwell <dcapwell@gmail.com>
Date:   Wed Aug 16 14:54:19 2023 -0700

    feedback: switched to suppliers to unblock the patch
```

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
belliottsmith: Do you have the SHA that caused this? It doesn't quite track to me what we're doing here. If the burn test wasn't ACKing epoch 1, I don't see how this change would help either, so would be great to debug an example so can understand the cause/fix.

accord-core/src/main/java/accord/local/Node.java
belliottsmith: Should we be doing this on every node start? If this is only used for test purposes, perhaps we should extend `Node` for testing, or at least mark VisibleForTesting. But if we start simulating node stop/start I'm not sure if this makes sense for testing either.

accord-core/src/main/java/accord/local/PreLoadContext.java
```diff
@@ -52,6 +54,17 @@ public interface PreLoadContext
      */
     default Collection<TxnId> additionalTxnIds() { return Collections.emptyList(); }
 
+    default Collection<TxnId> txnIds()
+    {
+        TxnId primaryTxnId = primaryTxnId();
+        Collection<TxnId> additional = additionalTxnIds();
```
belliottsmith, line 61: If we're adding a default method here, would prefer not to copy all contents to a list when we know it's either additional or additional +1. Why not extend `AbstractCollection`? Or `AbstractList` would be cleaner, if we can restrict `additionalTxnIds()` to `List`

accord-core/src/test/java/accord/burn/random/FrequentLargeRange.java
```diff
@@ -18,59 +18,121 @@
...
+    public FrequentLargeRange(LongGen small, LongGen large, double ratio)
     {
-        Invariants.checkArgument(ratio > 0 && ratio <= 1);
         this.small = small;
         this.large = large;
-        this.ratio = ratio;
-        this.steps = (int) (1 / ratio);
-        this.lower = ratio * .8;
-        this.upper = ratio * 1.2;
+        this.runs = Gens.bools().biasedRepeatingRuns(ratio);
```
belliottsmith, line 38: Didn't we discus moving this to `RandomSource`?

accord-core/src/test/java/accord/utils/Gen.java
```diff
@@ -69,6 +89,16 @@ default Gen<A> filter(Predicate<A> fn)
...
+    {
+        return () -> next(rs);
+    }
+
+    default Stream<A> asStream(RandomSource rs)
+    {
+        return Stream.generate(() -> next(rs));
+    }
+
     interface IntGen extends Gen<Integer>
```
belliottsmith, line 102: Why does this extend `Gen<Integer>`? This leads to much uglier mappings, and confusion about whether there is any boxing. If this utility is to be more broadly used, it would be nice to simplify the reading and writing of test cases with it. If we want to box or unbox, a `boxed()` method is preferable.

accord-core/src/test/java/accord/burn/random/FrequentLargeRange.java
dcapwell: no, you want that and I strongly disagree with that.  If we keep pushing more and more generators into the `RandomSource` it gets too large and harder to discover things.  It also leads to a coupling I strongly feel is bad that everything is "owned" by a `RandomSource`, making it harder for reuse

accord-core/src/test/java/accord/utils/Gen.java
dcapwell: > Why does this extend Gen<Integer>?

Its so you can still use it in all contexts, by having disjoint interfaces that have the same shape, it gets harder to compose together.

>  If this utility is to be more broadly used, it would be nice to simplify the reading and writing of test cases with it

Can you expand on this?  Several tests use this and the `LongGen` version.

> If we want to box or unbox, a boxed() method is preferable.

I am +0 to `.boxed()` vs extends; they solve the same problem.  I can test this out but think I would need to add more methods to this type and gets easier for these primitive types to miss methods added to the base.

accord-core/src/main/java/accord/local/PreLoadContext.java
dcapwell: > Why not extend AbstractCollection? 

what extends?  I don't follow this comment.

belliottsmith: We are performing a costly array copy, when we know we only have one additional element. We can extend AbstractCollection to return the merge of these two items.

accord-core/src/test/java/accord/burn/random/FrequentLargeRange.java
belliottsmith: I didn't recall seeing any reply to my earlier suggestion, sorry if I missed it.

Is there a reason some (e.g. `pick`) methods warrant entry to `RandomSource` but not others?

I think we're going to continue to have friction here. I will continue to embellish `RandomSource` as this is the natural place for me and how I create tests, and is why I created the abstraction. You will continue to put generators into `Gen` because this is how you prefer to write tests, and why you created that abstraction. This is fine while the approaches do not mix; we already have plenty of redundancy in test authorship. But we either have to accept that we have different approaches and stick to the approach already in use when editing tests, or we have to standardise. I have generally tried to avoid imposing my views on your approach, as I don't see any reason to debate abstractions I don't intend to use. But in this case we're editing my work done with my approach, and so the approaches are clashing, which is why I asked not to use the `Gen` abstraction. We seem to have as a result slipped into assuming we need to come to a conclusion about the one true approach, but I fear on this we will ever reach agreement. If we can agree not to impose our approaches on the other, and when editing tests to try to obey the existing style (as we do elsewhere in the codebase), I do not see why both (and more!) approaches, and generators, cannot co-exist.

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
dcapwell: I don't remember a sha/seed, would have to splunk through slack to see if I could find that.  It happened very often with this patch without this change though, not sure if other changes made it less helpful or not though.

So, the history was the following

```
T1: burn test starts with epoch=1, but does not setup the ack; this was a difference with C* which would ack
T2: epoch=2, but ack is slow due to timeouts/failures
T3: epoch=3, ack is slightly faster than epoch=2, so it wins the race and fails
```

I made a change to `start` so we ack the first topology, so if you call start (which we don't in C*) then `lastAcked !=0` and you wouldn't hit this issue.  In the C* case we stored the state locally and on reboot we delay ack until we can replay the events again, so it could be possible to hit this there if the oldest topology had not seen all replies yet.

accord-core/src/main/java/accord/local/Node.java
dcapwell: > Should we be doing this on every node start?

so, the code documents that this is for maelstrom and is used in burn test, but is not used in C*... so I am treading it as a test method.  The main reason is that we need to know the state before we start, and the fact a epoch=1 exists is more a product of our test infra than a real impl.

> If this is only used for test purposes, perhaps we should extend Node for testing

I have 0 issue with a `TestNode` that adds this for maelstrom / burn test

> But if we start simulating node stop/start I'm not sure if this makes sense for testing either.

I can sync with @iamaleksey on this issue so he can flesh this out with his work.  In C* we have a store of what replies we have seen and what are in-flight, so on restart we know what subset to pick form; in accord-core main/test we don't as everything is in-memory so we need to simulate that.  If we are implementing start/stop we need to flesh that out, so feel it would be best to leave there.

accord-core/src/main/java/accord/local/PreLoadContext.java
dcapwell: let me make sure I am clear

```
if (primaryTxnId == null) return additional
if (additional.isEmpty) return Collections.single(primaryTxnId)
return new LooksLikeAArray(primaryTxnId, additionalTxnIds)
```

if so, I don't mind making that change, if not would need to better get where you are coming from

belliottsmith: Pretty much

accord-core/src/test/java/accord/burn/random/FrequentLargeRange.java
belliottsmith: (I should be clear this is a more general comment about the debate we’re having on multiple fronts, and how we might move forwards. I don’t much mind if this method lives in Gens if that’s where you want it and we can otherwise agree on the above principle)

dcapwell: > Is there a reason some (e.g. pick) methods warrant entry to RandomSource but not others?

some picks have state, some do not; some are light weight, some are not.  I have been moving the picks with state outside as they need some extra state and a type to hold it.  so `pick(list)` can stay in the source as its stateless, but `pickWithWeight(list)` "should" stay outside as it needs to hold some state (the weight).  I actually don't mind moving `pick(list)` out into a static function, I just saw it as a very common pattern so felt it worked best in the interface in the current usage.

```
rs.pick(a, b, c)
// vs
pick(rs, a, b, c)
```

I am personally fine with both, but feel that the below is bad as it imposed a cost at each invoke, so makes more sense to push outside as it "should" return a holder of this state (Supplier, Gen, w/e).

```
rs.pickWithWeight(a, b, c)
```

feel free to call me out for my inconsistent stance with `pick(Set)`, this one violates what I stated above, and prob shouldn't be there; I mostly put it there out of laziness and to fix bugs I found with picking form sets in a non-deterministic manner... atm the function is dead code as its only used in my follow up work, so I am also cool with dropping it for now till its used.

> is why I created the abstraction

I am the author, I think you mean why you did similar in Simulator?  When I created this interface I wanted to make sure to have what you needed in simulator, what I needed in property tests, and perfectly match the JDK interface to make migration easier.  When I created this interface, I wanted to make sure it worked in all worlds and didn't impose anything on anyone...

> I think we're going to continue to have friction here. I will continue to embellish RandomSource as this is the natural place for me and how I create tests, and is why I created the abstraction. You will continue to put generators into Gen because this is how you prefer to write tests, and why you created that abstraction. This is fine while the approaches do not mix; we already have plenty of redundancy in test authorship. But we either have to accept that we have different approaches and stick to the approach already in use when editing tests, or we have to standardise.

This feels like a better conversation for dev@ or slack with a wider group of opinions; you prefer `Supplier`, I prefer `Gen`, but both work from a technical point of view so we need a tie breaking to move forward IMO.

>  I have generally tried to avoid imposing my views on your approach, as I don't see any reason to debate abstractions I don't intend to use.

We can choose to live in different worlds, but to be frank I am not the one imposing my views.  In the eviction branch I pointed out your property tests had issues with them and showed how to fix them but also mentioned you didn't need to write as the code existed; I let you choose which way to go.  In this patch your published stance is that I must conform to your view else I am blocked... 

> But in this case we're editing my work done with my approach

how are we editing *your* work when I am the author of this class?  This argument holds in https://github.com/apache/cassandra-accord/pull/58 as you are refactoring my work; this patch does not impose a `Gen` on *your* work; your patch does impose your views on me though.

> We seem to have as a result slipped into assuming we need to come to a conclusion about the one true approach, but I fear on this we will ever reach agreement. If we can agree not to impose our approaches on the other, and when editing tests to try to obey the existing style (as we do elsewhere in the codebase), I do not see why both (and more!) approaches, and generators, cannot co-exist.

I would prefer this, we even had a similar conversation in dev@ when SAI wanted to add a new random testing library to the project; the stance we moved forward with is allow all to live.

But to restate what I said above, I am not forcing my stuff on you, you are forcing yours on me.

belliottsmith: `RandomSource` is a port of my work for the simulator, and copied over at my suggestion.

`RandomDelayQueue`, `BurnTest` and `Cluster` are all my creations and these are where you introduced your new abstractions.

When I pushed back against using `Gen` there, this seemed to turn into a general question about how we might rationalise the approaches, as you seem to prefer that I and others adopt your abstractions and generators more generally.

dcapwell: > where you introduced your new abstractions.

* `RandomDelayQueue`/ `BurnTest` uses `accord.burn.random.FrequentLargeRange` which has existed for a long time at this point, and to follow the style you are imposing on me I convert to a `Supplier`, so in this patch the stored interface is a `Supplier` (again, the interface you are imposing), but all we do is reusing existing work.
* `Cluster` doesn't use `Gen` in this patch, all the random selection is delegated to the objects, which isn't a change to how this class works as this class never used `RandomSource` or anything directly, it delegates all randomness to its other elements; which again is unchanged in this patch.

belliottsmith: Cluster has been refactored to NodeSink, which alongside BurnTest and RandomDelayQueue were modified in this way prior to your reverting them, which happened after this debate was initiated.

Anyway, there’s no point debating this further if we’re in agreement about how to proceed for now. Though I fear we’re going to continue to revisit this.

dcapwell: > Cluster has been refactored to NodeSink,

do you mean the moving the `partition` logic to `NodeSink`?  That doesn't have anything to do with `Gen` vs `Supplier` argument and was done to keep the logic clear and easier for debug logging; I added logging that shows if we chose to drop the msg vs we dropped due to it being partitioned.

> which alongside BurnTest and RandomDelayQueue were modified in this way prior to your reverting them, which happened after this debate was initiated.

* `BurnTest` always returned a Supplier, it was just a lambda now its a function call

```
return () -> time.next(rs);
// vs now
return time.asSupplier(rs);
```

The change in this patch was to reusing existing logic to implement a behavior we wanted, to add jitter for time... 

* `RandomDelayQueue` never exposed `Gen`, it just used an existing type `accord.burn.random.FrequentLargeRange` and called it using `.next`.  A better argument in this case is that the code was cleaner to extract the mapping logic to the constructor and use a `LongSupplier` in this case, so in this case I would agree the `LongSupplier` is cleaner code than the previous, hence why I didn't bother to push back here.

```
// before
TimeUnit.NANOSECONDS.toMillis(jitterNanos.next(rs))
// after
jitterMillis.getAsLong()
```

> Though I fear we’re going to continue to revisit this.

I agree I feel this is going to keep coming up.  The examples you gave never exposed `Gen`, they just reusing existing code... so the pushback to me was that when using existing code that happens to have a `Gen` interface, and intersects classes that you feel must be a given way, then must use a style you want, that is not documented, fleshed out, or voted on...  so as a contributor to this project I have no idea when the next argument will pop up.

What's also unexpected to me about this is that Simulator has a similar style to `Gen` (it has both styles); you supply the `RandomSource`[1].  We even spoke months back about one day unifying these so that Simulator extends the logic in Accord.  So from my view this feedback came out of no where and became a blocking comment... so when will the next one come up?

[1]

```
// see org.apache.cassandra.simulator.ClusterSimulation.Builder#schedulerFactory
static SchedulerFactory schedulerFactory(RunnableActionScheduler.Kind... kinds)
        {
            return (random) -> {
                switch (Choices.random(random, kinds).choose(random))
                {
                    default: throw new AssertionError();
                    case SEQUENTIAL: return new RunnableActionScheduler.Sequential();
                    case UNIFORM: return new RunnableActionScheduler.RandomUniform(random);
                    case RANDOM_WALK: return new RunnableActionScheduler.RandomWalk(random);
                }
            };
        }
```

belliottsmith: Perhaps we are talking past each other? Now we no longer utilise the `Gen` API within the top-level declarations of these classes, I am suggesting we move forwards. Your examples are comparing this patch in its current form with the patch in its form when this disagreement arose, which is to say where `Gen.LongGen` etc was peppered around the implementation of the aforementioned classes.

https://github.com/apache/cassandra-accord/pull/57/commits/49da37997ab4b1406c0069c9f387ab025d8990bb

This is mixing styles in my book. In fact, there remain examples I would prefer we removed, e.g. https://github.com/apache/cassandra-accord/pull/57/commits/49da37997ab4b1406c0069c9f387ab025d8990bb#diff-d7eab055034d6123ac548c3da83300a6ab4aa050e057d3d2cf03b059ab6df502R197 - this is very different to my style of authorship, which would directly compose primitive actions on `RandomSource` since that is entirely sufficient here. I haven't asked for that, though; just to isolate the usages to as close to the instantiation as possible, and not confuse the API/top level of the classes.

My style of code does not use `Gen` at all, and introducing it - as demonstrated above - leads to a very different style of code being written. This is totally fine when authoring your own classes, but it's less well received by authors of their own classes that they're still maintaining, in which consistency of style will be harmed. I don't think this is the first time we've discussed our conflicting approaches to test authorship, so this shouldn't be a huge surprise? If your style and its components begin to weave their way amongst my style of authorship, then suddenly I have to have opinions about how your API is structured, it's inescapable (but that's its own can of worms, neither of us have time for right now).

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -230,37 +237,50 @@ static void burn(RandomSource random, TopologyFactory topologyFactory, List<Id>
 
         // not used for atomicity, just for encapsulation
         AtomicReference<Runnable> onSubmitted = new AtomicReference<>();
-        Consumer<Packet> responseSink = packet -> {
-            ListResult reply = (ListResult) packet.message;
-            if (replies[(int)packet.replyId] != null)
-                return;
-
+        Runnable maybeTriggerNextRequest = () -> {
```
belliottsmith: Why is this refactor necessary? The explanation doesn't quite make sense to me, as we were always enqueueing the next message before, unless we had already done so by receiving a later response first. 

I think in both version we potentially breach our concurrency restrictions, as we will enqueue for both the heartbeat and any future response.

accord-core/src/test/java/accord/burn/BurnTest.java
belliottsmith: I'm sure this is fine

accord-core/src/test/java/accord/impl/basic/PendingQueue.java
```diff
@@ -23,9 +23,11 @@
 public interface PendingQueue
 {
     void add(Pending item);
+    void addNoDelay(Pending item);
     void add(Pending item, long delay, TimeUnit units);
     boolean remove(Pending item);
     Pending poll();
     int size();
+    void checkFailures();
```
belliottsmith: This method only makes sense for the `PropagatingPendingQueue`. Does it really need to be exposed?

accord-core/src/test/java/accord/impl/list/ListRequest.java
```diff
@@ -162,7 +209,11 @@ public ListRequest(Txn txn)
     @Override
     public void process(Node node, Id client, ReplyContext replyContext)
     {
-        node.coordinate(txn).addCallback(new ResultCallback(node, client, replyContext, txn));
+        if (id != null)
+            throw new IllegalStateException("Called process multiple times");
+        id = node.nextTxnId(txn.kind(), txn.keys().domain());
+        ((NodeSink) node.messageSink()).debugClient(id, txn, NodeSink.ClientAction.SUBMIT);
```
belliottsmith: Can we e.g. have some utility method for this to hide the explicit casts?

accord-core/src/test/java/accord/impl/basic/Cluster.java
```diff
@@ -289,8 +278,9 @@ public static EnumMap<MessageType, Stats> run(Id[] nodes, Supplier<PendingQueue>
 
             while (!sinks.onDone.isEmpty())
             {
-                sinks.onDone.forEach(Runnable::run);
+                List<Runnable> onDone = new ArrayList<>(sinks.onDone);
```
belliottsmith, line 286: What issue is this addressing? We loop on processing `onDone`, so presumably we have some dependency for `onDone` work cessation on some `processPending` work?

accord-core/src/test/java/accord/impl/list/ListRequest.java
```diff
@@ -110,49 +122,84 @@ public void accept(Result success, Throwable fail)
...
+            }
+            else
+            {
+                ((NodeSink) node.messageSink()).debugClient(id, fail, NodeSink.ClientAction.FAILURE);
+                node.agent().onUncaughtException(fail);
+            }
+        }
+
+        private void checkOnResult(@Nullable RoutingKey homeKey, TxnId txnId, int attempt, Throwable t) {
+            if (attempt == 42)
```
belliottsmith: I think this was left in erroneously

accord-core/src/test/java/accord/burn/random/FrequentLargeRange.java
dcapwell: > Your examples are comparing this patch in its current form with the patch in its form when this disagreement arose, 

No, I am not.  I am replying to your comments and arguing they are not really valid; I give such examples as `return () -> time.next(rs);` vs `return time.asSupplier(rs);` which actively compares pre-feedback and post-feedback.  I also reply to your comment about `Cluster` and `NodeSink` where the change doesn't have anything to do with `Gen` vs `Supplier`, the change was mostly for debug logging reasons.

>  I am suggesting we move forwards.

What are the open issues before a +1 is given?  It is not clear what feedback is required vs optional as the only active threads I see is https://github.com/apache/cassandra-accord/pull/57#discussion_r1301845319 (small change to a single method) and this thread, so I want to make sure I understand the scope of work remaining.

> This is mixing styles in my book.

This is not at all true though.  `trunk` branch is currently 8c7a3c9ef4209d635b186189e17a2d9e728e9871, which has the following call paths that this PR touches that are under conflict

* `BurnTest` `nowSupplier`:  https://github.com/apache/cassandra-accord/blob/8c7a3c9ef4209d635b186189e17a2d9e728e9871/accord-core/src/test/java/accord/burn/BurnTest.java#L202-L206.  In this example we use a `Supplier<LongSupplier>` as we create one per node, and `Node` uses `LongSupplier` for time.  Within this supplier we do `return () -> Math.max(0, delayQueue.nowInMillis() + (jitter.nextInt(10) - 5));`.  Before applying the feedback you asked for, we still returned a `Supplier<LongSupplier>` and replaced the body to be `return () -> Math.max(0, queue.nowInMillis() + jitter.nextLong(forked));`. So, the only change before was

```
// before
jitter.nextInt(10) - 5
// after
jitter.nextLong(forked)
```

Before the logic was stateless, but we moved to a stateful model to express "runs" of long jitter, so moved from calling `RandomSource` to `FrequentLargeRange`.  I don't see how this logic was not within the style before and how it is now, or how future changes are any different.

* `RandomDelayQueue` `add`: https://github.com/apache/cassandra-accord/blob/8c7a3c9ef4209d635b186189e17a2d9e728e9871/accord-core/src/test/java/accord/impl/basic/RandomDelayQueue.java#L102

```
// before
random.nextInt(500)
// after
TimeUnit.NANOSECONDS.toMillis(jitterInNano.nextLong(random))
// after cleanup
jitterInMillis.nextLong(random)
```

Again, this moved from a stateless source of random to a stateful one and the local style is not really changed.

* `NodeSink` `send`: https://github.com/apache/cassandra-accord/blob/8c7a3c9ef4209d635b186189e17a2d9e728e9871/accord-core/src/test/java/accord/impl/basic/NodeSink.java#L65-L79. This class would delegate delivery and timeouts to the `PendingQueue` and only used random to change the delay; which is mostly unchanged in my patch (before/after).  The core change to this class is making it so we don't `DELIVER` every time and conditionally deliver; which is the feature request that this JIRA is scoped to do.  So

```
// before 
parent.add(self, to, messageId, send); // deliver
// after
Action action = partitioned(to) ? Action.DROP_PARTITIONED
                                        // call actions() per node so each one has different "runs" state
                                        : nodeActions.computeIfAbsent(to, ignore -> actions()).next(rs);
// can be a DELIVER, DELIVER_WITH_FAILURE, FAILURE, DROP, or DROP_PARTITIONED
```

The other change was that we don't use "cpu scheduling delay" (as we talked about) and to use a "network latency delay", so again

```
// before
parent.add(self, to, messageId, send); // deliver
// after
parent.add(packet, networkJitterNanos(to), TimeUnit.NANOSECONDS);
```

I struggle to see how I change the "style" (outside of `NodeSink` which went from nothing to something...).  If this is the argument imposed in this PR I have no clue how to deal with https://github.com/apache/cassandra-accord/pull/58 which actually changes the style (moving away from `RandomSource` in favor of `Supplier` and interfaces that extend it.  This moves away from the top level classes *owning* the random and how to do things in favor of a new abstraction)!

> this is very different to my style of authorship

I have to push back on this... Apache Accord style guide is the Apache Cassandra style guide, which does not document what you are imposing; that is your personal style.  Our current rules state that you have to have `clearly expressed and reasonable technical grounds` for such a pushback; we are supposed to work as a team.  If we randomly start changing the rules without voting and documenting, then we are not able to work as a team.

> since that is entirely sufficient here

Your patch exposes a new interface and move away from `RandomSource` within the context of this argument (since we are talking about the code calling `.next`)... you use `RandomSource` to create a `Supplier`, which is a new style you wish to move forward with in https://github.com/apache/cassandra-accord/pull/58.  In this patch I use `FrequentLargeRange`, so is your argument here that `FrequentLargeRange` should be merged into a function call within `RandomSource`?  If so this is the first time I have seen such an argument imposed here, the closest was the comment about `Gens.bools().biasedRepeatingRuns(ratio)` vs `rs.biasedRepeatingRuns(ratio)`; to me this feels like a very different conversation than `Supplier` vs `Gen`.

> This is totally fine when authoring your own classes, but it's less well received by authors of their own classes that they're still maintaining, in which consistency of style will be harmed. 

The code is owned by Apache and we all are expected to work together on it; neither of us *own* any file.  

I still hold by my pushback above, I don't see how I change the style of the code on trunk.  We can not use `RandomSource` directly at each callback as part of the requirement for this JIRA is to add some set of state to improve the randomness, that state is currently stored in a `Gen` and in your other PR you want to introduce `Supplier`.  I would argue that my style is closer to the current style than ours as the *owner* of the `RandomSource` is the top level class.

> I don't think this is the first time we've discussed our conflicting approaches to test authorship, so this shouldn't be a huge surprise? 

You have told me before you don't like `Gen` but have not given actual reasons.  The signature is also present (just different named) in Simulator, so it's hard to see what your argument actually is.

We talked about this in eviction, where I did not force you to match my style, I pointed out that you were duplicating code and had bugs, so you could fix your code (and I showed you how) or reuse.  You had expressed there again that you don't like `Gen` and would prefer to hand roll your own logic, which I was fine with.

> If your style and its components begin to weave their way amongst my style of authorship, then suddenly I have to have opinions about how your API is structured, it's inescapable.

And the same seems to be true for me it seems.  I had expressed my feedback in your patch and am +0 for you to introduce `Supplier`; I don't want to block you because of my preferences, but you actively choose to block me due to yours.  I don't agree with your choice, but I will not block you on it as it is still "correct".  I will ask you to fix or remove code that isn't (which I pointed out in your PR), but since `Supplier` works, I won't block it.  

So yes, we are different here.  I do not try to force you into my style, but you wish to force me, then argue that we should live in peace doing our own things...

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: The issue I faced was that so many txn failed and got delayed to `onDone` that we stoped enqueuing the next packets.  We detected that this happened and would do `return;` and not start the next packets.

> I think in both version we potentially breach our concurrency restrictions, as we will enqueue for both the heartbeat and any future response.

I saw this code as limiting the number of packets and not in-flight `Txn`, so we see a client exception (timeout / fault) and would enqueue the next.

accord-core/src/test/java/accord/impl/basic/PendingQueue.java
dcapwell: we use `PendingQueue` and not `PropagatingPendingQueue`, so by changing the signature we use I can move it there; this is leaking details of one impl.

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -200,13 +202,18 @@ static void burn(RandomSource random, TopologyFactory topologyFactory, List<Id>
         });
 
         Supplier<LongSupplier> nowSupplier = () -> {
-            RandomSource jitter = random.fork();
-            // TODO (expected): jitter should be random walk with intermittent long periods
-            return () -> Math.max(0, delayQueue.nowInMillis() + (jitter.nextInt(10) - 5));
+            RandomSource forked = random.fork();
+            return FrequentLargeRange.builder(forked)
+                                                   .raitio(1, 5)
```
belliottsmith: ratio

accord-core/src/test/java/accord/impl/list/ListRequest.java
dcapwell: we construct this within the `generate` method, so we could always pass in something, sure

accord-core/src/test/java/accord/burn/random/FrequentLargeRange.java
```diff
@@ -18,59 +18,121 @@
...
+            this.ratio = ratio;
+            return this;
+        }
+
+        public Builder raitio(int min, int max)
         {
-            run--;
-            largeCount++;
-            return large.getLong(randomSource);
+            this.ratio = ratio = random.nextInt(min, max) / 100.0D;
```
belliottsmith: redundant assignment?

accord-core/src/test/java/accord/impl/basic/Cluster.java
dcapwell: `new ArrayList<>(sinks.onDone);`  was to fix a `ConcurretModification` issue, the `onDone` method threw so we enqueue in `sinks.onDone` then failed.

accord-core/src/test/java/accord/impl/list/ListRequest.java
dcapwell: I think I used to have 3 or something, `42` was to see if the issue was ephemeral or not (the issue was not); I can lower the retry limit

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
dcapwell: >  In the C* case we stored the state locally and on reboot we delay ack until we can replay the events again

Correction, the very first topology we load from disk we ack

So yeah, with the change to `start` this change isn't 100% needed, mostly left just in case

accord-core/src/test/java/accord/impl/basic/PendingQueue.java
dcapwell: removed from this interface and pass a new `Runnable checkFailures` to `Cluster` which calls this function.

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: > as we will enqueue for both the heartbeat and any future response.

"should" there be any "future responses"?  We send the heart beat IFF the client saw a failure (other than Invalidated) and enqueue trying to figure out the state in `onDone`.  `onDone` is only triggered once we finished processing all the inflight work, which implies the response sink stopped enqueueing work.  So the "future responses" should only be after we have finished enqueueing all work

The issue I saw was that we stopped enqueuing work, so we fail with `throw new AssertionError("Incomplete set of responses; clock=" + clock.get() + ", expected operations=" + (operations * 2));`.  The root cause I found was that if we don't enqueue on the heart beat there are conditions where we stop progressing the backlog and complete early and fail as the expected amount of operations didn't happen.

accord-core/src/test/java/accord/impl/list/ListRequest.java
dcapwell: created a new `MessageListener` and have everything use that

accord-core/src/test/java/accord/burn/random/FrequentLargeRange.java
belliottsmith: Ideally this is something we would randomise for each run, but let's look at that in follow-up.

accord-core/src/test/java/accord/burn/BurnTest.java
belliottsmith: > "should" there be any "future responses"? We send the heart beat IFF the client saw a failure

Yes, the client will later receive a message if the transaction is recovered - at least to notify of its successful application (I think we do not notify the client on invalidation). 

> I saw this code as limiting the number of packets and not in-flight Txn, so we see a client exception (timeout / fault) and would enqueue the next.

Yes, but the client will (hopefully) receive the later _valid_ response that informs of the outcome of the transaction, at which point it will trigger _another_ pack send, breaching our limits.

> The issue I saw was that we stopped enqueuing work

I don't think that error meant that we had stopped enqueueing work that was still pending - it means we didn't get a complete set of responses to the work we had successfully enqueued. Reaching that message should mean we *must have* successfully enqueued every client packet. We don't stop the simulation until we have sent all packets and processed all ensuing work.

accord-core/src/main/java/accord/local/Node.java
belliottsmith: That documentation appears to be wrong as it claims it's only used by `Maelstrom`. So, let's document it as for testing only and perhaps call it `unsafeStart` or `testStart` - otherwise this is cool by me.

dcapwell: went with `unsafeStart` and documented this behavior

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: Looking at the git history I think this is a by product of writing this patch months ago

```
f7fc6e18aacdeb902f4482ae9fe9429c7e5b8bf8	Fri May 12 01:53:04 PDT 2023	Benedict Elliott Smith	Permit nodes to join a cluster without the full transaction history
```

That patch made sure we notify on all `CoordinationFailure`, so this logic isn't needed anymore

dcapwell: reverted this change

belliottsmith (approved):

PR #62, opened 2023-08-22
CEP-15: Minimize transaction state kept in system tables

iamaleksey:
patch by Aleksey Yeschenko; reviewed by Ariel Weisberg for CASSANDRA-18573

### DIFF
--- a/accord-core/src/main/java/accord/messages/Propagate.java
+++ b/accord-core/src/main/java/accord/messages/Propagate.java
@@ -1,0 +1,358 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package accord.messages;
+
+import accord.api.Result;
+import accord.api.RoutingKey;
+import accord.coordinate.Infer;
+import accord.local.Command;
+import accord.local.Commands;
+import accord.local.Node;
+import accord.local.SafeCommand;
+import accord.local.SafeCommandStore;
+import accord.local.SaveStatus;
+import accord.local.Status;
+import accord.primitives.EpochSupplier;
+import accord.primitives.Keys;
+import accord.primitives.PartialDeps;
+import accord.primitives.PartialTxn;
+import accord.primitives.Participants;
+import accord.primitives.Ranges;
+import accord.primitives.Route;
+import accord.primitives.Seekables;
+import accord.primitives.Timestamp;
+import accord.primitives.TxnId;
+import accord.primitives.Writes;
+import accord.utils.Invariants;
+import accord.utils.MapReduceConsume;
+
+import javax.annotation.Nullable;
+import java.util.function.BiConsumer;
+
+import static accord.local.Status.NotDefined;
+import static accord.local.Status.Phase.Cleanup;
+import static accord.local.Status.PreApplied;
+import static accord.primitives.Routables.Slice.Minimal;
+
+public class Propagate implements MapReduceConsume<SafeCommandStore, Void>, EpochSupplier, LocalMessage
+{
+    public static class SerializerSupport
+    {
+        public static Propagate create(TxnId txnId, Route<?> route, SaveStatus saveStatus, SaveStatus maxSaveStatus, Status.Durability durability, RoutingKey homeKey, RoutingKey progressKey, Status.Known achieved, PartialTxn partialTxn, PartialDeps partialDeps, long toEpoch, Timestamp executeAt, Writes writes, Result result)
+        {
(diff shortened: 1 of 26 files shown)


accord-core/src/main/java/accord/primitives/PartialTxn.java
```diff
@@ -55,6 +57,8 @@ default boolean covers(Participants<?> participants)
         return covering().containsAll(participants);
     }
 
+    boolean covers(PartialTxn txn);
```
aweisberg: `covers` is already used for an indication that the state of the transaction covers some keys/ranges and it's a different concept from the sufficient equivalence being tested by the message `AccordStateCache` validation. Could we use a different name? `equivalentOrSuperceding`?

accord-core/src/main/java/accord/messages/Propagate.java
```diff
@@ -0,0 +1,358 @@
...
+
+    @Override
+    public MessageType type()
+    {
+        switch (achieved.propagate())
+        {
+            case Applied:
+            case PreApplied:
+                if (toEpoch >= executeAt.epoch())
+                    return MessageType.PROPAGATE_APPLY_MSG;
```
aweisberg, line 314: Huh, so a lot of these cases are kind of black magic without explanation. If the status is `Applied` or `PreApplied` why does `toEpoch` matter? How can `toEpoch` be > `executeAt.epoch` when it's either assigned `srcEpoch` or `executeAt.epoch`?

What even is `toEpoch`? The highest `epoch` the transaction might execute in?

accord-core/src/main/java/accord/messages/Propagate.java
```diff
@@ -0,0 +1,358 @@
...
+            case Committed:
+            case ReadyToExecute:
+                return MessageType.PROPAGATE_COMMIT_MSG;
+            case PreCommitted:
+                if (!achieved.definition.isKnown())
+                    return MessageType.PROPAGATE_OTHER_MSG;
```
aweisberg, line 320: Precommitted but no definition known? So we glimpsed a transaction that is still in flight and didn't contact anyone that actually knew the definition? And this is after running `FetchData` and/or `RecoverWithRoute`?

accord-core/src/main/java/accord/messages/Propagate.java
```diff
@@ -0,0 +1,358 @@
...
+            case PreAccepted:
+                if (Route.isFullRoute(route))
```
aweisberg: I thought we always send the full route now so how can it be `PreAccepted`, but not have the route?

aweisberg (changes requested):
I could stare a little longer at `SerializerSupport` to try an ascertain whether it round trips things correctly, but really it will be a moving target and if it's tested well that is the best validation.

accord-core/src/main/java/accord/messages/Propagate.java
belliottsmith: We might not contact a quorum. We also might not contact any replicas of our range, so they may not have our portion of the definition. Or we might contact a replica that only partially replicates our local ranges that participate in the transaction. In some cases all we want to know is the `executeAt`, at which point we can stop asking for more responses.

accord-core/src/main/java/accord/messages/Propagate.java
belliottsmith: If it's what it was before, it's the epoch we can successfully propagate for. If for instance we adopt an additional range in the execution epoch, we may not have requested enough data to propagate Applied or PreApplied.

accord-core/src/main/java/accord/messages/Propagate.java
belliottsmith: It might be impossible to reach this branch now.

accord-core/src/main/java/accord/primitives/PartialTxn.java
iamaleksey: Coming up with a name for it was a struggle, because all these terms are overloaded, including 'superceding'. I scouted thesaurus to no satisfactory outcome.

accord-core/src/main/java/accord/messages/Propagate.java
iamaleksey: This was extracted from `OnDone#apply()` switch statement verbatim, except with calls to `Commands#{apply, commit, preaccept}` replaced with message type enum.

accord-core/src/main/java/accord/messages/Propagate.java
iamaleksey: @belliottsmith do you mean impossible *not* to, once reached `PreAccepted` label? Could also have fallen through `PreCommitted` but should have `route` in that case too anyway.

belliottsmith: Right, I mean this condition might well be redundant. If it isn’t, perhaps we should make sure that it is.

accord-core/src/main/java/accord/primitives/PartialTxn.java
iamaleksey: Renamed to `isEqualOrFuller()`, cheers.

accord-core/src/main/java/accord/messages/Propagate.java
iamaleksey: Removed this branch, for it indeed seems unreachable now.

accord-core/src/main/java/accord/primitives/PartialTxn.java
```diff
@@ -24,6 +24,10 @@
 import accord.api.Read;
 import accord.api.Update;
 
+import java.util.Objects;
+
+//import java.util.Objects;
```
aweisberg: Cleanup import

aweisberg (changes requested):
One minor nit for the import

accord-core/src/main/java/accord/primitives/PartialTxn.java
iamaleksey: Cleaned up, cheers.

aweisberg (approved):
+1 TY

PR #64, opened 2023-08-23
Compaction purger fix

aweisberg:
shouldCleanup can't enforce invariants because some of the time we are checking an invalid slice of command state

### DIFF
--- a/accord-core/src/main/java/accord/local/Commands.java
+++ b/accord-core/src/main/java/accord/local/Commands.java
@@ -886,7 +886,12 @@ public class Commands
                              safeStore.commandStore().redundantBefore(), safeStore.commandStore().durableBefore());
     }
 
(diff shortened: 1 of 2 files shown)


accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -895,6 +900,7 @@ public static Cleanup shouldCleanup(TxnId txnId, Status status, Durability durab
             return Cleanup.ERASE;
         }
 
+        // TODO (review): If we always send the full route when will this happen?
```
dcapwell: in C* side I saw that `route` is `null` before, so this can happen when compaction purger interacts with it; not seen your new C* side though

dcapwell: `Route.isFullRoute` allows `null`... but at this point it "shouldn't" be null?  Im in favor of keeping it as is

dcapwell (approved):

accord-core/src/main/java/accord/local/Commands.java
aweisberg: Yeah you are right it can be null

PR #65, opened 2023-08-29
CASSANDRA-18804: (Accord): Bug fixes from CASSANDRA-18675 to better support adding keyspaces

dcapwell:


accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -92,6 +94,11 @@ public abstract class InMemoryCommandStore extends CommandStore
 
     private final TreeMap<TxnId, RangeCommand> rangeCommands = new TreeMap<>();
     private final TreeMap<TxnId, Ranges> historicalRangeCommands = new TreeMap<>();
+    /**
+     * Since this cache is fully in-memory the store does not hit states that most stores will; that data is not in-memory!
+     * To simulate such behaviors, a "cache" is used to state what is in-memory vs what needs to be loaded.
+     */
+    private final Set<Object> cache = new LinkedHashSet<>();
```
dcapwell: this was unrelated to add/remove keyspace patch, it was mostly due to an observation we had a lot of code checking if things are cached to avoid loading and since this class is 100% in-memory we were at 100% cache hit rate, so went with faking a cache miss to get better coverage.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -408,63 +464,39 @@ void update(Ranges add)
         }
     }
 
-    static class CFKEntry extends TxnId
+    enum Loader implements CommandLoader<Command>
```
dcapwell: CFK is supposed to be immutable, but the old `CFKEntry` used the global backing commands map and would return updated values, making CFK mutable!  Switched away from mutable loader to immutable loader to better match the expected behavior and what C* does...

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -871,31 +903,41 @@ public static class Synchronized extends InMemoryCommandStore
...
         public Synchronized(int id, NodeTimeService time, Agent agent, DataStore store, ProgressLog.Factory progressLogFactory, EpochUpdateHolder epochUpdateHolder)
         {
             super(id, time, agent, store, progressLogFactory, epochUpdateHolder);
         }
 
-        private synchronized void maybeRun()
+        private void maybeRun()
         {
-            if (active != null)
+            if (!running.compareAndSet(false, true))
```
dcapwell: this is not related to this patch but was from a different patch I had... I added more messages tests that used this class and I hit a deadlock where one store (S1) was trying to enqueue work to a different store (S2) and that store was trying to enqueue work in the other store (S1)...

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -871,31 +903,41 @@ public static class Synchronized extends InMemoryCommandStore
...
                 return;
-
-            active = queue.poll();
-            while (active != null)
+            synchronized (this)
```
dcapwell: was not 100% sure this was fully needed given the single threaded nature of the work, but felt best to not try to change the locking too much

accord-core/src/main/java/accord/impl/SimpleProgressLog.java
```diff
@@ -277,7 +277,7 @@ void run(SafeCommandStore safeStore, SafeCommand safeCommand)
 
                                                 updateMax(success.asProgressToken());
                                             }
-                                        });
+                                        }).begin(node.agent());
```
dcapwell: the leak detector found this case.  `CommandStore` returns a `AsyncChain` and not `AsyncResult`!  In Cassandra we use `AsyncChains.Head` so require `begin` to start, but in the in-memory stories we returned a `AsyncResult` which would make code work under tests but not in C*...

accord-core/src/main/java/accord/messages/TxnRequest.java
```diff
@@ -234,7 +234,7 @@ public static <I, O> O computeScope(Node.Id node, Topologies topologies, I keys,
         O scope = null;
         for (int i = startIndex, mi = topologies.size() ; i < mi ; ++i)
         {
-            Topology topology = topologies.get(i);
+            Topology topology = topologies.get(i).global();
```
dcapwell, line 247: this was a bug found with remove prefix.  The subset topologies would leak the `global` values and not the subset values, which caused the scope logic to work as it works off `global`, so once this was fixed I had to be explicit and ask for the `global` values...

accord-core/src/main/java/accord/messages/WaitUntilApplied.java
```diff
@@ -170,6 +170,7 @@ private ReadNack apply(SafeCommandStore safeStore, SafeCommand safeCommand)
                 }
                 else
                 {
+                    // TODO (debug): this will cause us to send a Nack back, so why are we listening?
```
dcapwell: will remove, but in debugger I wasn't sure why we do this... `NotCommitted` causes the process to return a `Nack` which causes the waiting to halt!  In my usage I keep retrying the wait over and over again when a `NotCommitted` or a `Timeout` is observed...

accord-core/src/main/java/accord/topology/Topology.java
```diff
@@ -147,11 +179,6 @@ public boolean isSubset()
         return supersetIndexes.length < shards.length;
     }
 
-    public Topology withEpoch(long epoch)
```
dcapwell, line 169: moved to `TopologyUtils` under `src/test` as the semantics are not clear what to do here... *should* this also update the `global` reference?  Since this was only used by tests and only for the `global` state anyways felt was easier to punt figuring out the semantic and just make it a test method

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -180,6 +182,14 @@ boolean syncCompleteFor(Unseekables<?> intersect)
         {
             return syncComplete.containsAll(intersect);
         }
+
+        @Override
+        public String toString()
+        {
+            return "EpochState{" +
```
dcapwell, line 189: I don't remember if this is logged, but im always in a debugger so this makes my life easier

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -498,7 +508,7 @@ public Topologies withUnsyncedEpochs(Unseekables<?> select, long minEpoch, long
         return withSufficientEpochs(select, minEpoch, maxEpoch, epochState -> epochState.syncComplete);
     }
 
-    public Topologies withOpenEpochs(Unseekables<?> select, Timestamp min, Timestamp max)
+    public Topologies withOpenEpochs(Unseekables<?> select, EpochSupplier min, EpochSupplier max)
```
dcapwell, line 511: this was updated for `ListStore` as it works with historic epochs; so rather than define a full `Timestamp`, using `EpochSupplier` allowed `() -> epoch` as well in this case

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -92,72 +102,101 @@ static List<Packet> generate(RandomSource random, Function<? super CommandStore,
 
         for (int count = 0 ; count < operations ; ++count)
         {
+            int finalCount = count;
             Id client = clients.get(random.nextInt(clients.size()));
             Id node = nodes.get(random.nextInt(nodes.size()));
 
             boolean isRangeQuery = random.nextBoolean();
+            String description;
+            Function<Node, Txn> gen;
```
dcapwell: since we don't know what topologies exist until we run on a node, moved the current logic into a lambda that runs in the node so we know what prefixes could be accessed...

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -243,6 +283,12 @@ static void burn(RandomSource random, TopologyFactory topologyFactory, List<Id>
                 if (i == requests.length - 1)
                     onSubmitted.get().run();
             }
+            if (packet.message instanceof Reply.FailureReply)
```
dcapwell, line 378: hit a NPE in the `List*` code which hit this case... so assuming the `List*` is bug free we should never hit this, but rather than hitting a class cast felt that the original exception was best so adding directly to the failures was better

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -311,6 +373,7 @@ static void burn(RandomSource random, TopologyFactory topologyFactory, List<Id>
 
         logger.info("Received {} acks, {} nacks, {} lost, {} truncated ({} total) to {} operations", acks.get(), nacks.get(), lost.get(), truncated.get(), acks.get() + nacks.get() + lost.get() + truncated.get(), operations);
         logger.info("Message counts: {}", messageStatsMap.entrySet());
+        logger.info("Took {} and in logical time of {}", Duration.ofNanos(System.nanoTime() - startNanos), Duration.ofMillis(queue.nowInMillis() - startLogicalMillis));
```
dcapwell, line 476: this was mostly for me... I wanted to get a better grasp of how much "logical" time has passed as there is a lot of delayed logic

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -357,6 +420,7 @@ public static void main(String[] args)
     @Test
     public void testOne()
     {
+//        run(726880888609333L, 1000); // taking a long time...  gave up after 15 minutes!
```
dcapwell: will revert, this is just a seed that hit a while true loop in recovery. Spoke to Benedict and seems that his branch may fix this

accord-core/src/test/java/accord/coordinate/tracking/TrackerReconcilerTest.java
```diff
@@ -67,7 +67,14 @@ void test(int count, BiFunction<RandomSource, Topologies, ? extends TrackerRecon
...
-        for (TrackerReconciler<?, ?, ?> test : TrackerReconciler.generate(seed, constructor))
-            test.test();
+        try
+        {
+            for (TrackerReconciler<?, ?, ?> test : TrackerReconciler.generate(seed, constructor))
+                test.test();
+        }
+        catch (Throwable t)
+        {
+            throw new AssertionError("Failure for seed " + seed, t);
```
dcapwell, line 77: unrelated to the core work, but was found when doing the topology refactors... the current test relies on logging which means that CI is unlikely to give us the seed we failed with, making theses property tests not reproducible...  by throwing with the seed we should be able to repo issues assuming the property tests are deterministic

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
```diff
@@ -52,6 +59,28 @@ public static CommandStores.Factory factory(PendingQueue pending)
                new DelayedCommandStores(time, agent, store, random, shardDistributor, progressLogFactory, new SimulatedDelayedExecutorService(pending, agent, random));
     }
 
+    @Override
+    protected boolean shouldBootstrap(Node node, Topology previous, Topology updated, Range range)
+    {
+        if (!super.shouldBootstrap(node, previous, updated, range))
+            return false;
+        if (!(range.start() instanceof PrefixedIntHashKey)) return true;
```
dcapwell, line 68: I think this logic could be rewritten to `previous.ranges.containsAll(Ranges.of(range))` which should actually match the C* logic... can switch but wasn't sure if there would be edge cases where we don't want that as the default behavior...

accord-core/src/test/java/accord/topology/TopologyManagerTest.java
```diff
@@ -277,4 +304,280 @@ void truncateTopologyCantTruncateUnsyncedEpochs()
     {
 
     }
+
+    @Test
+    void removeRanges()
```
dcapwell, line 307: BurnTest does not include remove range support, but TM does in this patch, so this test is still relevant.  The remove range in BurnTest has issues with multiple subsystems so was punted, but this subsystem was low effort so kept

accord-core/src/test/java/accord/topology/TopologyManagerTest.java
```diff
@@ -277,4 +304,280 @@ void truncateTopologyCantTruncateUnsyncedEpochs()
...
+                    .topology(1, a -> a.isEmpty())
+                    .topology(2, a -> a.isEmpty())
+                    .topology(3, a -> a.isNotEmpty()
+                                       .isRangesEqualTo(PrefixedIntHashKey.range(0, 0, 100))
+                                       .isHostsEqualTo(dc2Nodes));
+        }
+    }
+
+    @Test
+    void fuzz()
```
dcapwell, line 401: this test was very useful to make sure the different mutations work with add/remove prefix.  I believe I got some of the fixes into earlier commits [1] so you don't see them there... this test is old for me...  I found that it was far faster to hit issues here than in `BurnTest`


[1]

```
commit 7c15f3a6203939bc6cb398e538df1ca3557cbe03
Author: Benedict Elliott Smith <benedict@apache.org>
Date:   Tue Aug 15 16:51:22 2023 +0100

    State Eviction (#50)

    Permit the state machine to erase transactions that are known to be applied across the cluster.
```

accord-core/src/test/java/accord/topology/TopologyRandomizer.java
```diff
@@ -26,64 +26,99 @@
...
+        {
+            @Override
+            public void onUpdate(Topology topology)
+            {
+
+            }
+        }
+    }
+
+    private static class State
```
dcapwell, line 74: `addPrefix`/`removePrefix` needs to know what is the highest prefix seen so far so have a counter for that, but since we are using enums we can't pass this state into `<init>` and instead have to pass this into the call site...  I "could" just pass in the counter to all functions but felt this way was slightly better... more than glad to rework this as I really don't like it

accord-core/src/test/java/accord/topology/TopologyUtils.java
```diff
@@ -0,0 +1,191 @@
...
+import accord.utils.Gens;
+import accord.utils.RandomSource;
+import accord.utils.Utils;
+import accord.utils.WrapAroundList;
+import accord.utils.WrapAroundSet;
+
+import javax.annotation.Nullable;
+import java.util.*;
+
+import static accord.utils.Utils.toArray;
```
dcapwell, line 39: this was moved into `accord.topology` package to get access to package private state (for `withEpoch`); most of the class is not new but git/GH doesn't seem to detect the move

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -354,6 +418,7 @@ public static void main(String[] args)
         }
     }
 
+    @Timeout(value = 3, unit = TimeUnit.MINUTES)
```
dcapwell: this is so the test times out rather than the build... when while true loops are found the build times out due to no progress... felt it was best to have the test timeout...

One issue with this is that the seed will be lost... we don't depend on `org.awaitility` like C* does, so could also add that test dependency and update the `run(seed)` method to do this timeout and log the seed on failure.

accord-core/src/test/java/accord/impl/list/ListStore.java
```diff
@@ -49,28 +147,162 @@ public ListStore(Node.Id node)
         this.node = node;
     }
 
-    public Timestamped<int[]> get(Key key)
+    public synchronized Timestamped<int[]> get(Ranges unavailable, Timestamp executeAt, Key key)
```
dcapwell, line 152: `synchronized` was from a different patch and only kept it here as `write` had it...

accord-core/src/test/java/accord/impl/list/ListStore.java
```diff
@@ -96,4 +328,222 @@ private static boolean isStrictPrefix(int[] a, int[] b)
...
+                runWhenReady(node, epoch, () -> removeLocalWhenReady(node, epoch, finalLocalRemove));
+            }
+            // TODO (correctness, coverage): add cleanup logic for global ranges removed. This must be solved in CASSANDRA-18675
+        }
+        previousTopology = topology;
+    }
+
+    private void runWhenReady(Node node, long epoch, Runnable whenKnown)
+    {
+        if (node.topology().epoch() >= epoch) whenKnown.run();
```
dcapwell, line 393: was finding issues as this logic happens *before* accord sees the epoch (to better mimic TCM) and the wait logic didn't expect gaps... so implemented my own wait logic...

accord-core/src/main/java/accord/coordinate/TopologyMismatch.java
```diff
@@ -0,0 +1,77 @@
...
+                sb.append('\n');
+            sb.append(String.format("HomeKey %s exists for a range that is no longer globally valid (%d -> %s)", homeKey, topology.epoch(), topology.ranges()));
+        }
+        return sb.toString();
+    }
+
+    @Nullable
+    public static TopologyMismatch checkForMismatch(Topology t, TxnId txnId, RoutingKey homeKey, Seekables<?, ?> keysOrRanges)
+    {
+        EnumSet<TopologyMismatch.Reason> reasons = EnumSet.noneOf(TopologyMismatch.Reason.class);
```
bdeggleston: Can we set this to null and only allocate the EnumSet if we encounter a problem?

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
bdeggleston: I think the SafeCommandStore/SafeCommand would mostly address this coverage? It should throw if you attempt to load something that wasn't included in the PreLoadContext

accord-core/src/main/java/accord/messages/WaitUntilApplied.java
bdeggleston: reminder to remove

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -363,6 +477,11 @@ protected void checkNextIsHead()
         Invariants.checkState(next instanceof Head<?>, "Next is not an instance of AsyncChains.Head (it is %s); was map/flatMap called on the same object multiple times?", next.getClass());
     }
 
+    public static AsyncChain<?> detectLeak(Consumer<Throwable> onLeak, Runnable onCall)
```
bdeggleston, line 480: this is cool, WDYT about adding a flag to include this in every chain that can be used in testing?

bdeggleston (changes requested):

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
dcapwell: We have a few of places that try to access data outside of the `PreLoadContext` and acts on the global cache.

`accord.impl.InMemoryCommandStore#ifPresent(accord.primitives.TxnId)`
`accord.impl.InMemoryCommandStore.InMemorySafeStore#getIfLoaded(accord.primitives.TxnId)`
`accord.impl.AbstractSafeCommandStore#getInternalIfLoadedAndInitialised`
`accord.local.SafeCommandStore#ifLoadedAndInitialised`
`accord.local.Commands.NotifyWaitingOn#accept`

```
SafeCommand curSafe = ifLoadedAndInitialised(safeStore, depth);
                Command cur = curSafe != null ? curSafe.current() : null;
                if (cur == null)
                {
                    // need to load; schedule execution for later
                    safeStore.commandStore().execute(this, this).begin(safeStore.agent());
                    return;
                }
```

Cases like this would be 100% cache hit in BurnTest, so the non-cache hit case was never tested.

EDIT: `s/a lot/a few/g`

accord-core/src/main/java/accord/utils/async/AsyncChains.java
dcapwell: > WDYT about adding a flag to include this in every chain that can be used in testing?

I currently use `accord.utils.Invariants#isParanoid` in `accord.impl.basic.DelayedCommandStores.DelayedCommandStore#submit(java.util.concurrent.Callable<T>)`, so `CommandStore` leaks should be detected.  Are there other places you are thinking about?  We have `accord.utils.async.AsyncChains.Head` but that feels tricky as we use `extends`, so would need a costly `finalize` or a big refactor to move away from `Head` directly.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
dcapwell: s/a lot/a few/g

accord-core/src/main/java/accord/coordinate/TopologyMismatch.java
```diff
@@ -0,0 +1,85 @@
...
+    {
+        EnumSet<TopologyMismatch.Reason> reasons = null;
+        if (!t.ranges().contains(homeKey))
+        {
+            if (reasons == null)
+                reasons = EnumSet.noneOf(TopologyMismatch.Reason.class);
+            reasons.add(TopologyMismatch.Reason.HOME_KEY);
+        }
+        Seekables<?, ?> updatedRanges = keysOrRanges.slice(t.ranges(), Routables.Slice.Minimal);
+        if (!updatedRanges.containsAll(keysOrRanges))
```
bdeggleston: Could we use AbstractKeys#any here, or something similar so we don't have to allocate a slice?

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
```diff
@@ -250,6 +250,16 @@ void onNewEpoch(Topologies prevTopologies, Timestamp executeAt, List<PreAcceptOk
         // TODO (desired, efficiency): check if we have already have a valid quorum for the future epoch
         //  (noting that nodes may have adopted new ranges, in which case they should be discounted, and quorums may have changed shape)
         node.withEpoch(executeAt.epoch(), () -> {
+            TopologyMismatch mismatch = TopologyMismatch.checkForMismatch(node.topology().globalForEpoch(executeAt.epoch()), txnId, route.homeKey(), txn.keys());
+            if (mismatch != null)
+            {
+                Invalidate.invalidate(node, txnId, route, (outcome, failure) -> {
```
bdeggleston: wouldn't we want to exclude the now unreplicated keys from the route to prevent invalidate from running into mismatch issues itself?

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
bdeggleston: can you take a look at how we handled this in the prune deps patch? Instead of adding an actual cache, there's a check to determine if we can expose objects not specified in the preload context. Then we can have it return true/false based on some random distribution. Aside from having a lower footprint, I think it would simulate a wider variety of eviction patterns.

See https://github.com/bdeggleston/cassandra-accord/blob/C18784-reduce-deps-v2/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java#L759C51-L759C51

bdeggleston (changes requested):

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
dcapwell: > unreplicated keys from the route to prevent invalidate

Can you say more?  This should include all keys/ranges for the txn

accord-core/src/main/java/accord/coordinate/TopologyMismatch.java
dcapwell: I think we need `t.contains(keysOrRanges)` to avoid allocation, slice/intersect may have non-intersect which would mean incompatible change; ill try to look closer to see if we can avoid allocation here

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
dcapwell: I'll try to understand `commandStore.canExposeUnloaded()` and get back to you.  This looks to be a store property and not based off the requested data...

I added the cache logic as I noticed that the Cassandra logic was drastically different than InMemory logic, and we kept having logic optimized for InMemory logic... so I wanted to get the 2 stores closer as majority of testing is in BurnTest and far less is done with Cassandra...

dcapwell: ill need to look closer at this change, after rebase we are failing and looks like its a cache hit issue... not sure why yet...  if I disable cache logic burn test fails and not sure why yet, so can't say this logic is broken or some other logic is actually broken...

dcapwell: honestly I am fine pushing this into its own stand alone patch, not related to the core changes, so if can get the rest in faster, thats bette

accord-core/src/main/java/accord/coordinate/TopologyMismatch.java
dcapwell: I did ranges.containsAll(keysOrRanges) as that is logically the same.

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
bdeggleston: so presumably we're invalidating this txn because it contains keys that don't intersect with the accord topology, ie: are unreplicated from the perspective of accord. If we try to invalidate with those same keys, won't the invalidation be unable to make progress because it's trying to find replicas for unreplicated keys?

dcapwell: I back ported my test from the drop keyspace branch and in a debugger I see `accord.coordinate.Invalidate#Invalidate`

```
Topologies topologies = node.topology().forEpoch(invalidateWith, txnId.epoch());
txnId = {TxnId@3550} "[1,5,2(KW),1]"
invalidateWith = {FullKeyRoute@3548} "{homeKey:1#51196,[1#26558,1#51196,1#54417]}"
this.invalidateWith = {FullKeyRoute@3548} "{homeKey:1#51196,[1#26558,1#51196,1#54417]}"
```

this uses topology `[Topology{epoch=1, [Shard[1#21843,1#29124]:(-1655615791f, -1170363843f, -942051380f), Shard[1#50967,1#58248]:(1f, 1298436801f, 1702745641f)]}]`
so, rather than using the new epoch which doesn't include the range, we use the range from the txn_id.epoch (which is still valid)

dcapwell: Take a look at `accord.local.CommandsTest#removeRangesValidate` to see that this works just fine

accord-core/src/main/java/accord/utils/RandomSource.java
```diff
@@ -233,6 +238,93 @@ default double nextDouble(double minInclusive, double maxExclusive)
 
     double nextGaussian();
 
+    default int pickInt(int first, int second, int... rest)
```
dcapwell, line 241: This was removed in CASSANDRA-19045, adding it back

accord-core/src/main/java/accord/utils/Timestamped.java
```diff
@@ -68,6 +71,6 @@ public static <T> Timestamped<T> mergeEqual(Timestamped<T> a, Timestamped<T> b,
     @Override
     public String toString()
     {
-        return data.toString();
+        return toString.apply(data);
```
dcapwell, line 74: `data` is almost always `int[]` which is useless... so logs on error didn't make sense without this change

accord-core/src/test/java/accord/Utils.java
```diff
@@ -118,6 +123,16 @@ public static Txn writeTxn(Ranges ranges)
         return new Txn.InMemory(ranges, MockStore.read(ranges), MockStore.QUERY, MockStore.update(ranges));
     }
 
+    public static Txn listWriteTxn(Node.Id client, Keys keys)
+    {
+        ListUpdate update = new ListUpdate(Function.identity());
+        for (Key k : keys)
+            update.put(k, 1);
```
dcapwell, line 130: this isn't 100% correct... if you run this txn twice the second one will freak out as ListRequest won't like this... it works just fine for single run cases (which is what I do)...  I guess I could add a `int...` or something to allow overriding... but didn't want to complicate things too much

accord-core/src/test/java/accord/impl/basic/Cluster.java
```diff
@@ -496,7 +506,7 @@ private Function<List<Id>, BiFunction<Id, Id, Link>> overrideLinks(RandomSource
                 case RANDOM_BIDIRECTIONAL:
                 case RANDOM_UNIDIRECTIONAL:
                     boolean bidirectional = kind == RANDOM_BIDIRECTIONAL;
-                    int count = random.nextInt(bidirectional || random.nextBoolean() ? nodesList.size() : (nodesList.size() * nodesList.size())/2);
+                    int count = random.nextInt(bidirectional || random.nextBoolean() ? nodesList.size() : Math.max(1, (nodesList.size() * nodesList.size())/2));
```
dcapwell, line 509: when nodesList size = 1, this fails (`1/2 = 0), hence the `Math.max(1` to keep it safe

accord-core/src/test/java/accord/utils/RandomSourceTest.java
```diff
@@ -19,25 +19,22 @@
...
 
     @Test
     public void testBiasedInts()
     {
-        RandomSource random = RandomSource.wrap(new Random());
-        long seed = random.nextLong();
-        logger.info("Seed: {}", seed);
-        random.setSeed(seed);
-        testBiasedInts(random, 1000, 100000, 0.01, 0.1);
+        test().check(random -> testBiasedInts(random, 1000, 100000, 0.01, 0.1));
```
dcapwell: this is also in https://issues.apache.org/jira/browse/CASSANDRA-19114, its a small change so hope that can go in quickly... I just want stable CI so had to fix this flakey test

accord-core/src/test/java/accord/utils/RandomTestRunner.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * The main difference between this class and {@link Property} is that this class does not reference {@link Gen} and some
+ * authors wish to avoid that interface, but need to correctly work with randomness in a test that is reproducable when
+ * issues are found from CI.
+ *
+ * {@link Builder#check(Consumer)} is functinally the same as the following from {@link Property}
+ *
+ * {@code qt().withExamples(1).check(testcase); }
+ */
+public class RandomTestRunner
```
dcapwell: this is in https://issues.apache.org/jira/browse/CASSANDRA-19114

buildSrc/src/main/groovy/accord.java-conventions.gradle
```diff
@@ -49,10 +49,6 @@ test {
     jvmArgs += ['-XX:+HeapDumpOnOutOfMemoryError', "-XX:HeapDumpPath=${buildDir}"]
 }
 
-tasks.withType(Test) {
```
dcapwell, line 52: looks like rebase issue that got merged... I added this, Benedict rebased onto my branch and added similar logic (see line 49), but didn't detect the logical conflict so this got merged.

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
bdeggleston: yep that makes sense, thanks

accord-core/src/test/java/accord/Utils.java
bdeggleston: I don't understand the problem, wouldn't ListUpdate just append another 1 if this was run twice?

dcapwell: possible... I really didn't bother testing... I think `Timestamped` doesn't like this... its been awhile...

bdeggleston: that's fine. If it has sharp edges that make it unsuitable outside of a narrow use case though, it shouldn't go in Utils.

dcapwell: ill test and push this to the test if it has such a sharp edge (think it does)

dcapwell: tested, we do just append, so this is actually safe...

bdeggleston (approved):
+1

accord-core/src/main/java/accord/coordinate/Barrier.java
```diff
@@ -66,7 +66,7 @@ public class Barrier<S extends Seekables<?, ?>> extends AsyncResults.AbstractRes
     private final BarrierType barrierType;
 
     @VisibleForTesting
-    CoordinateSyncPoint<S> coordinateSyncPoint;
+    AsyncResult<SyncPoint<S>> coordinateSyncPoint;
```
belliottsmith, line 69: Do we need to make this type less specific?

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
belliottsmith: This should invoke `Propose.proposeInvalidate` not `Invalidate.invalidate`

belliottsmith: I think we should also set `initialPreAcceptIsDone=true` else we might yet complete successfully as well as propose invalidation, which would be a serious fault.

accord-core/src/main/java/accord/coordinate/ReadCoordinator.java
```diff
@@ -251,7 +251,7 @@ protected void start(Set<Id> to)
 
     public void start()
     {
-        Set<Id> contact = newHashSetWithExpectedSize(maxShardsPerEpoch());
+        Set<Id> contact = newLinkedHashSetWithExpectedSize(maxShardsPerEpoch());
```
belliottsmith: I think it is a vestigial requirement that this be a set. We can just allocate a `List`, and modify the `start` method to accept a `List`.

accord-core/src/main/java/accord/impl/CommandTimeseries.java
```diff
@@ -128,14 +128,17 @@ public Timestamp minTimestamp()
         Timestamp result = null;
         for (D data : commands.values())
         {
+            TxnId txnId = loader.txnId(data);
+            Timestamp executeAt = loader.executeAt(data);
             if (result == null)
             {
-                result = Timestamp.min(loader.txnId(data), loader.executeAt(data));
+                result = executeAt == null ? txnId : Timestamp.min(txnId, executeAt);
```
belliottsmith: `Timestamp.nonNullOrMin`?

accord-core/src/main/java/accord/impl/CommandTimeseries.java
```diff
@@ -128,14 +128,17 @@ public Timestamp minTimestamp()
...
             }
             else
             {
-                result = Timestamp.min(result, loader.txnId(data));
-                result = Timestamp.min(result, loader.executeAt(data));
+                result = Timestamp.min(result, txnId);
+                if (executeAt != null)
+                    result = Timestamp.min(result, executeAt);
```
belliottsmith: `Timestamp.nonNullOrMin`?

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -479,9 +486,19 @@ public <T> void send(Collection<Id> to, Function<Id, Request> requestFactory, Ca
...
+        checkIterationSafe(to);
         to.forEach(dst -> messageSink.send(dst, requestFactory.apply(dst), executor, callback));
     }
 
+    private static void checkIterationSafe(Collection<?> collection)
+    {
+        if (!Invariants.isParanoid())
+            return;
+        if (collection instanceof List) return;
+        if (collection instanceof NavigableSet || collection instanceof LinkedHashSet || "java.util.LinkedHashMap.LinkedKeySet".equals(collection.getClass().getCanonicalName())) return;
```
belliottsmith: `|| collection instanceof DeterministicSet`

accord-core/src/main/java/accord/messages/Commit.java
```diff
@@ -275,6 +275,22 @@ public static void commitInvalidate(Node node, TxnId txnId, Unseekables<?> infor
             // TODO (expected, safety): this kind of check needs to be inserted in all equivalent methods
             Invariants.checkState(untilEpoch >= txnId.epoch());
             Invariants.checkState(node.topology().hasEpoch(untilEpoch));
+            // Possible that the invalidation is due to the range no longer existing in untilEpoch, which means that
```
belliottsmith: So a couple of thoughts here:

1) How can this happen? We should only invoke with an `untilEpoch` we attempted to `Accept` with, and we shouldn't attempt to `Accept` in an epoch where the keys don't exist
2) If we have somehow done this, I think it is fine to contact `untilEpoch` even if some keys are missing? The problem with going to the earliest epoch we can fine that contains them is that if we have somehow gone into a future epoch, we might have left traces we should ideally invalidate for other keys that did still exist, and have moved replica

My preference would be to first figure out (1), as if it's happening it may indicate some other issue that we can address. Then we might anyway want to support a version of `preciseEpochs` that permits topology mismatches.

accord-core/src/main/java/accord/primitives/AbstractRanges.java
```diff
@@ -56,7 +56,10 @@ public int indexOf(RoutableKey key)
     @Override
     public int indexOf(Range find)
```
belliottsmith: I think we should rename `indexOf` to `ceilIndexOf` just so callers are aware it's not necessarily unique. Or have it require the `Search` parameter.

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -70,101 +72,133 @@
...
+                gen = n -> {
+                    int[] prefixes = prefixes(n.topology().current());
+
+                    int rangeCount = 1 + random.nextInt(2);
+                    List<Range> requestRanges = new ArrayList<>();
+                    while (--rangeCount >= 0)
+                    {
+                        int prefix = random.pickInt(prefixes);
+                        int i = random.nextInt(HASH_RANGE_START, HASH_RANGE_END);
+                        int j = 1 + random.nextInt(i, HASH_RANGE_END);
```
belliottsmith: This doesn't cap the size of the range involved in the transaction, so we'll tend towards transactions that cover half the cluster. The prior code wasn't perfect, of course - ideally we would probably vary the maximum range size for each run.

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -70,101 +72,133 @@
...
 
         for (int count = 0 ; count < operations ; ++count)
         {
+            int finalCount = count;
             Id client = clients.get(random.nextInt(clients.size()));
             Id node = nodes.get(random.nextInt(nodes.size()));
 
             boolean isRangeQuery = random.nextBoolean();
+            String description;
+            Function<Node, Txn> gen;
```
belliottsmith: more descriptive name?

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -217,21 +254,41 @@ static void burn(RandomSource random, TopologyFactory topologyFactory, List<Id>
...
+            throw new IllegalAccessError("Global executor should enver get a stale event");
+        }));
+        Int2ObjectHashMap<Verifier> validators = new Int2ObjectHashMap<>();
         Function<CommandStore, AsyncExecutor> executor = ignore -> globalExecutor;
 
         MessageListener listener = MessageListener.get();
 
-        Packet[] requests = toArray(generate(random, listener, executor, clients, nodes, keyCount, operations), Packet[]::new);
+        if (keyCount > ((long) HASH_RANGE_END - (long) HASH_RANGE_START + 1L))
+            throw new AssertionError(String.format("Attempted to create %d keys which is larger than the range (%d, %d]", keyCount, HASH_RANGE_START, HASH_RANGE_END));
```
belliottsmith: Why is that a bug? Multiple keys can map to a single hash. It should be fine to run with a hash range of [0..1) an 1000 keys if we want.

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -217,21 +254,41 @@ static void burn(RandomSource random, TopologyFactory topologyFactory, List<Id>
...
+            for (int i = 0; i < keyCount; i++)
+            {
+                int hash;
+                do
+                {
+                    // start is exclusive, and end is inclusive; which is the oposite of nextInt, so +1 to account for this
+                    hash = 1 + random.nextInt(HASH_RANGE_START, HASH_RANGE_END);
+                }
+                while (!seen.add(hash));
+                keys[i] = CRCUtils.reverseCRC32LittleEnding(hash);
```
belliottsmith: What is it we're achieving here, with this refactor? This appears to give up the possibility of hash collisions for reverse lookups, but what value do we derive from reverse lookups?

belliottsmith (commented):
Looks pretty good to me, some minor feedback/queries to discuss.

accord-core/src/main/java/accord/coordinate/Barrier.java
dcapwell: yes, as we may return a `accord.utils.async.AsyncResults.Immediate` with a failure.

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
dcapwell: looking into this now

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: > but what value do we derive from reverse lookups

In `BurnTest` no value, but in other tests it makes it so we can be very explicit what ranges we are interacting with by generating keys within those ranges.  I use this property to create keys within a given range and outside that range so we can see how we behave under different conditions.

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
dcapwell: made this change

accord-core/src/main/java/accord/primitives/AbstractRanges.java
dcapwell: I am not a fan of this as this method is coming from `accord.primitives.Routables#indexOf` and that doesn't really make sense to me with the other 2 implementations (finding a key in a set of keys).

One possibility is to add a *new* method on `AbstractRanges` that lets you define the `Search` param and have `indexOf` return the *first* match.  The only code path I see that might use this new API is `accord.primitives.AbstractRanges#intersects(accord.primitives.Range)` as it could possible do less checks as it can return after the first match

belliottsmith: I would prefer to remove `indexOf` from `Routables` and introduce a new `indexOf` method only for `RoutableKey` types (e.g. RoutingKeys and Keys). We shouldn't really _have_ an `indexOf` method that accepts a `Range` for `AbstractRanges` as its poorly defined.

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: pushed a change that I hope solves this problem.  I bound ranges to be of the following sizes

```
[655, 1310, 2620, 5240, 10480, 20960] // if 20960 is too big its a config change to shrink
```

`655 = (HASH_RANGE_END - HASH_RANGE_START + 1) / 100`
and the rest of the values are `* 2` the previous.

I added a new `pickZipf` function to select the range size and made sure this logic respects it.  

I looked closer at the previous logic (on trunk) and it selects from the full range of values (64k), then again from 8k, and if this wraps around we default to `0`.  This was effectively selecting a range of 1-8k uniformly (not exactly, simplifying and ignoring wrap around case); figured changing the distribution away from uniform would also be a welcome change.

accord-core/src/test/java/accord/burn/BurnTest.java
belliottsmith: I don't think we need to use the same hashing behaviour between tests though? It would be nicer to support hash collisions IMO, as this tests additional functionality. Though we could force collisions if we wanted as well, by reducing the hash range rather than hard coding it.

accord-core/src/main/java/accord/primitives/AbstractRanges.java
dcapwell: pushed.  `indexOf` is now in `AbstractKeys` and `AbstractRanges` but they don't share the same signature; ranges asks for a search op.  I added a new `firstIndexOf` that uses `CEIL` with the new method in the single place that called this before.

accord-core/src/test/java/accord/burn/BurnTest.java
belliottsmith: If we're changing distribution, we should randomise the distribution for each run also, as my experience of zipf is that it actually is often less useful for this kind of test as most values are the same, and they cluster very small. Good distribution behaviour is annoyingly hard. Uniform is not perfect either, but I would ordinarily prefer it over zipf for this kind of input, so we should at least switch between the two.

dcapwell: if you prefer uniform I can do a simple `int rangeSize = rs.nextInt(0, 1 << 13) + 1`.  If you feel there isn't value in zipf then the complexity of switching between them doesn't seem worth it

dcapwell: my main thinking was to add a bias either to small or large ranges (I pushed bias towards small)

dcapwell: got back from lunch and implemented logic that each run does 1 of 4 things

50% chance to select uniform
50% chance to select zipf

When in zipf you have a uniform chance to select the following bias: small, large, random.

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: I went with `txnGenerator`, not sure a better name... it generates `Txn`

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: ill get back to this comment later, you have a different comment this depends on so best to address things there.

https://github.com/apache/cassandra-accord/pull/65#discussion_r1442889538

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: > I don't think we need to use the same hashing behaviour between tests though?

Agree

> It would be nicer to support hash collisions IMO

Let me take a look at this again.  The reason for this line `keys[i] = CRCUtils.reverseCRC32LittleEnding(hash);` was to try to respect `keyCount` (I interpret as number of unique keys to test again) and the limited hash range you had desired.  Since `PrefixedIntHashKey` uses crc32 for the hashing and doesn't mod like the other hash impl, the only way to limit the hash range would be to reverse the hash to get the keys.

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: `BurnTest` now owns the hash function, so able to remove this check.  It was here because a key outside of that range wouldn't be known in the `Cluster`, so fail fast.  Now that we rely on a truncated hash range for keys this safety check doesn't matter.

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: `BurnTest` now owns the hash function, so supports collisions.  I kept the previous keys rather than randomly selecting keys like my previous patch.

accord-core/src/main/java/accord/messages/Commit.java
dcapwell: > How can this happen? We should only invoke with an untilEpoch we attempted to Accept with, and we shouldn't attempt to Accept in an epoch where the keys don't exist

This is the stack trace that hits it

```
commitInvalidate:287, Commit$Invalidate (accord.messages)
commitInvalidate:270, Commit$Invalidate (accord.messages)
lambda$commitInvalidate$12:365, Recover (accord.coordinate)
run:-1, 1373949107 (accord.coordinate.Recover$$Lambda$689)
withEpoch:297, Node (accord.local)
commitInvalidate:365, Recover (accord.coordinate)
lambda$invalidate$10:358, Recover (accord.coordinate)
accept:-1, 73688220 (accord.coordinate.Recover$$Lambda$686)
onSuccess:202, Propose$Invalidate (accord.coordinate)
onSuccess:137, Propose$Invalidate (accord.coordinate)
accept:-1, 353023213 (accord.messages.SafeCallback$$Lambda$555)
lambda$safeCall$1:74, SafeCallback (accord.messages)
```

this epoch is coming from `accord.coordinate.Recover#commitInvalidate`

```
Timestamp invalidateUntil = recoverOks.stream().map(ok -> ok.executeAt).reduce(txnId, Timestamp::max);
```

the peers that are sending the `executeAt` have the following for `Command`

```
Command@1850137076{[1,5,2(KW),7]:AcceptedInvalidateWithDefinition}
```

This looks to be triggered by `accord.messages.Accept.Invalidate#apply`


>  I think it is fine to contact untilEpoch even if some keys are missing?

I believe my original logic that Blake approved allowed this, but since you asked for TM to reject these cases, we had to handle here.

belliottsmith: Ok, so the mistake is that we use a later epoch that is voted on (and recorded as such) on preaccept that wasn't proposed. So I guess we really want to replicate the original epoch we used for sending, which means for any Accept we see, we use the highest timestamp. But for PreAccept we only use TxnId. In this case we *should* never use a future epoch, and we should still invalidate/clear any records we might have created. 

So, something like
`Timestamp invalidateUntil = recoverOks.stream().map(ok -> ok.saveStatus.hasBeen(Accepted) ? ok.executeAt : ok.txnId).reduce(txnId, Timestamp::max);`

This isn't perfect, as an `AcceptInvalidate` could mask an `Accept` with a higher timestamp, and cause us to not clear some state. But I'm not sure it's worth agonising over, as eventually it all gets cleaned up. Better to keep things simple.

accord-core/src/main/java/accord/primitives/AbstractRanges.java
belliottsmith: I think it would be cleaner to either call it `ceilIndexOf` or just supply/accept `CEIL` parameter, so we don't mix descriptors of functionality.

accord-core/src/main/java/accord/coordinate/TxnExecute.java
```diff
@@ -74,8 +75,9 @@ private TxnExecute(Node node, TxnId txnId, Txn txn, FullRoute<?> route, Particip
     }
 
     @Override
-    protected void start(Set<Id> readSet)
+    protected void start(Iterable<Id> to)
     {
+        Set<Id> readSet = Sets.newHashSet(to);
```
belliottsmith: maybe switch to `IntHashSet` and accept it in `Commit`?

accord-core/src/main/java/accord/messages/Commit.java
dcapwell: I pushed this change and see that `CommandsTest` is stable with it, also tried to document the "why" for next authors who review

dcapwell:
Here are `BurnTest` seeds that fail for me (in C* IntelliJ) that I feel are not caused by this patch

```
        Suppressed: java.lang.IllegalStateException: Unexpected status for Command@1920466142{[12,29534,1(RR),1]:ReadyToExecute}
               at accord.utils.Invariants.illegalState(Invariants.java:44)
               at accord.utils.Invariants.checkState(Invariants.java:127)
               at accord.coordinate.Infer$InvalidateAndCallback.apply(Infer.java:265)
               at accord.coordinate.Infer$CleanupAndCallback.apply(Infer.java:229)
               at accord.coordinate.Infer$CleanupAndCallback.apply(Infer.java:201)
         */
        run(4016471194711250L, 1000);
```

```
/*
        Suppressed: java.lang.IllegalStateException: [1, 2, 3, 4] != [1, 2, 3]
		at accord.utils.Invariants.illegalState(Invariants.java:44)
		at accord.utils.Invariants.checkState(Invariants.java:133)
		at accord.utils.Timestamped.mergeEqual(Timestamped.java:67)
		at accord.impl.list.ListData.lambda$merge$0(ListData.java:37)
		at java.base/java.util.Map.merge(Map.java:1270)
		at accord.impl.list.ListData.lambda$merge$1(ListData.java:37)
		at java.base/java.util.TreeMap.forEach(TreeMap.java:1002)
		at accord.impl.list.ListData.merge(ListData.java:36)
		at accord.coordinate.TxnExecute.process(TxnExecute.java:104)
		at accord.coordinate.TxnExecute.process(TxnExecute.java:49)
		at accord.coordinate.ReadCoordinator.onSuccess(ReadCoordinator.java:121)
		at accord.coordinate.ReadCoordinator.onSuccess(ReadCoordinator.java:40)
         */
        run(4016642298723958L, 1000);
```

accord-core/src/main/java/accord/primitives/AbstractRanges.java
dcapwell: I used `firstIndexOf` to be similar to `java.util.List#lastIndexOf`, but switched to `ceilIndexOf` for now

accord-core/src/main/java/accord/coordinate/TxnExecute.java
dcapwell: ill make the change but I wonder if its not best to just do `List.contains`; I went with `Iterable` to make sure I didn't break anything... I feel like in practice the read set will be small as its 1 node per non-overlapping range.

belliottsmith (approved):

PR #67, opened 2023-09-21
CASSANDRA-18874: Add Jepson's Elle to Accord and Paxos validation

dcapwell:


accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -313,6 +316,7 @@ static void burn(RandomSource random, TopologyFactory topologyFactory, List<Id>
                                           topologyFactory, initialRequests::poll,
                                           onSubmitted::set
             );
+            verifier.close();
```
dcapwell, line 319: if the simulation throws, we don't care about the verifier also checking...

henrikingo:
Didn't review im depth, but I superficially checked a few things and wanted to note for the record I have no objections or comments.

As an example I went and checked whether ee would have implemented these workloads in some tool we already use for other testing. In particular nosqlbench. The answer is no. So using the official elle client is good.

accord-core/src/test/java/accord/verify/ElleVerifier.java
```diff
@@ -0,0 +1,411 @@
...
+        {
+            super(Type.append, key, value);
+        }
+    }
+
+    private static class Event extends ObjectPersistentMap
+    {
+        enum Type
+        {
+            ok, fail; // invoke, info
```
kijanowski: To increase coverage, isn't `info` also required as a possible response type ([as per docs](https://github.com/jepsen-io/elle/blob/main/src/elle/list_append.clj#L582)), where we (on client side) do not know, if the transaction has been applied or if it failed?
I'm asking because in the [clojure based test](https://github.com/datastax/accordclient/blob/main/src/accordclient/core.clj#L85-L93) I've used for the algorithm verification I've mapped:
- `UnavailableException` and `NoHostAvailableException` to `:fail`
- `ReadTimeoutException`, `WriteTimeoutException` and `OperationTimedOutException` to `:info`

Wanted to make sure, that I'm not misinterpreting response types and `info` has been skipped deliberately.

belliottsmith: I think we probably also want to issue `:invoke` with the start time as well?

belliottsmith: I'm very surprised that the `:invoke` tag is ignored, as it seems impossible to distinguish between some valid and invalid traces without knowing whether two operations overlapped their execution or not.

dcapwell: hmmm maybe I was confused reading the code... thought invoke is what we sent and info was random stuff like "added network partition"... ill try to look closer

dcapwell: reading the code I see `dirty-update-cases` uses `:info` so will look to add that.  I did see `verify-unique-appends` check for `invoke`, this looks to make sure only a single `:invoke` wrote `X`... I don't see any other usage...

ill try adding both; this will add more memory but is more "correct"... still not sure that either really provides value for Elle, but I don't know Clojure well so maybe im missing something.

belliottsmith: The comments linked above suggest invoke isn’t used; I’m just very surprised by this if it’s true, as I can’t see how you would detect some invalid traces.

dcapwell: joys... `invoke` means I must change the process ids... they don't allow concurrent process... so history#pair-index-fold fails with 

```
{:type :jepsen.history/double-invoke,
 :op
 {:process 0,
  :type :invoke,
  :f nil,
  :value [[:r 4]],
  :index 1,
  :time 60},
 :running
 {:process 0,
  :type :invoke,
  :f nil,
  :value [[:r 6]],
  :index 0,
  :time 3}}
```

dcapwell: @belliottsmith , @kijanowski pushed a change to add `:invoke` and did see that it is used (it gets mad if you have concurrent invoke from the same `process`... so had to make every `process` our `start`; was planning to make it client but that would break the model)

accord-core/src/test/java/accord/verify/ElleVerifier.java
```diff
@@ -0,0 +1,384 @@
...
+            else
+                throw new UnsupportedOperationException("Unable to update key " + key);
+            return this;
+        }
+    }
+
+    private static class Keys
+    {
+        // event keys
+        private static Keyword process = RT.keyword(null, "process");
```
maedhroz: nit: Everything from `process` to `emptyTransactionGraph ` looks like it could be `final`

accord-core/src/test/java/accord/verify/ElleVerifier.java
```diff
@@ -0,0 +1,384 @@
...
+                    for (Object type : seq)
+                    {
+                        if (type == Keys.emptyTransactionGraph)
+                        {
+                            empty = true;
+                            continue; // nothing to see here
+                        }
+                        throw new AssertionError("Unexpected anomaly type detected: " + type);
+                    }
+                    if (empty)
```
maedhroz: nit: `empty` is always `true` here (I guess because we've guarded w/ `!seq.isEmpty()`)

accord-core/src/test/java/accord/verify/ElleVerifier.java
```diff
@@ -0,0 +1,384 @@
...
+
+        private ObjectPersistentMap(Set<Keyword> keys)
+        {
+            this.keys = keys;
+        }
+
+        @Override
+        public boolean containsKey(Object key)
+        {
+            return keys.contains(key);
```
maedhroz, line 308: nit: might be worth an assertion that `key instanceof Keyword`

accord-core/src/test/java/accord/verify/ElleVerifier.java
```diff
@@ -0,0 +1,384 @@
...
+            return value == null ? 2 : 3;
+        }
+    }
+
+    private static class Read extends Action
+    {
+        protected Read(int key, int[] seq)
+        {
+            // TODO (optimization): rather than vector of boxed int, can we use the interfaces so we can stay primitive array?
+            super(Type.r, key, seq == null ? null : PersistentVector.create(IntStream.of(seq).mapToObj(Integer::valueOf).collect(Collectors.toList())));
```
maedhroz:
```suggestion
            super(Type.r, key, seq == null ? null : PersistentVector.create(IntStream.of(seq).boxed().collect(Collectors.toList())));
```
...shorter I guess, if you don't get to the TODO for now

maedhroz (approved):

PR #68, opened 2023-09-29
CASSANDRA-18816: backport changes from repair patch

dcapwell:


accord-core/src/main/java/accord/utils/DefaultRandom.java
```diff
@@ -20,25 +20,20 @@
 
 import java.util.Random;
 
-public class DefaultRandom extends Random implements RandomSource
+public class DefaultRandom extends WrappedRandomSource
```
dcapwell, line 23: needed for jdk17 support as `Random` conflicts with methods in `RandomSource`

maedhroz (approved):

accord-core/src/test/java/accord/burn/random/FrequentLargeRangeTest.java
```diff
@@ -30,7 +30,7 @@
 
 import static accord.utils.Property.qt;
 
-class SegmentedRandomRangeTest
+class FrequentLargeRangeTest
```
dcapwell, line 33: looks like I forgot to rename this class when I renamed `SegmentedRandomRange ` to `FrequentLargeRange`

PR #70, opened 2023-10-12
Minor fixes in the burn

stonewhitener:
1. Fixed a problem that `writeCount` might be `0` even though `isWrite` is true.
2. Since the number of nodes is in `[rf, 3 * rf)`, `Math.max(nodes.size() + 1, rf)` always returns `nodes.size() + 1` when determining the number of ranges in a cluster. Therefore, the unnecessary `max` has been removed.

## Note

The following is not included in this commit for now.
Furthermore, I would suggest the number of ranges should include `nodes.size()` therefore the following line:
```java
IntHashKey.ranges(random.nextInt(nodes.size() + 1, nodes.size() * 3))
```
should be:
```java
IntHashKey.ranges(random.nextInt(nodes.size(), nodes.size() * 3))
```

### DIFF
--- a/accord-core/src/test/java/accord/burn/BurnTest.java
+++ b/accord-core/src/test/java/accord/burn/BurnTest.java
@@ -119,7 +119,7 @@ public class BurnTest
             {
                 boolean isWrite = random.nextBoolean();
                 int readCount = 1 + random.nextInt(2);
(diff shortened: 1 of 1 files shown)


belliottsmith:
Thanks @stonewhitener

PR #73, opened 2023-11-28
Fixed flakey test from RandomSourceTest as the binary search needed a ceil to find the first match in the list.  Fixed a bug where the seed can get clost or put into the wrong test

dcapwell:


accord-core/src/test/java/accord/utils/RandomSourceTest.java
```diff
@@ -111,18 +104,101 @@ private double testOneBiasedLongs(RandomSource random, int min, int median, int
...
-        int i = Arrays.binarySearch(results, median);
+        int i = firstBinarySearch(results, median);
         if (i < 0) i = -1 - i;
-        int j = Arrays.binarySearch(results, median + 1);
+        int j = firstBinarySearch(results, median + 1);
         if (j < 0) j = -2 - j;
         else --j;
-        i -= results.length/2;
-        j -= results.length/2;
+        i -= Math.abs(results.length/2);
```
belliottsmith: I don't think this is necessary? It's not possible for `results.length/2` to be negative?

accord-core/src/test/java/accord/utils/RandomSourceTest.java
```diff
@@ -111,18 +104,101 @@ private double testOneBiasedLongs(RandomSource random, int min, int median, int
             results[i] = random.nextBiasedInt(min, median, max);
 
         Arrays.sort(results);
-        int i = Arrays.binarySearch(results, median);
+        int i = firstBinarySearch(results, median);
         if (i < 0) i = -1 - i;
-        int j = Arrays.binarySearch(results, median + 1);
+        int j = firstBinarySearch(results, median + 1);
```
belliottsmith: If we're copying `ceil`, we could just use `ceil` directly?

dcapwell: renamed the method to `ceil`. I also simplified it to call the forked `ceil` and only exists to simplify the signature.

belliottsmith:
+1

PR #74, opened 2023-11-29
Pre-requisite changes for CASSANDRA-18888

iamaleksey:
patch by Aleksey Yeschenko; reviewed by Ariel Weisberg for CASSANDRA-18888

### DIFF
--- a/accord-core/src/main/java/accord/messages/LocalRequest.java
+++ b/accord-core/src/main/java/accord/messages/LocalRequest.java
@@ -19,13 +19,19 @@ package accord.messages;
 
 import accord.local.Node;
 import accord.local.PreLoadContext;
+import accord.local.SafeCommandStore;
+import accord.utils.MapReduceConsume;
 
-public interface LocalMessage extends Message, PreLoadContext
+import java.util.function.BiConsumer;
+
+public interface LocalRequest<R> extends Request, PreLoadContext, MapReduceConsume<SafeCommandStore, Void>
 {
+    void process(Node on, BiConsumer<R, Throwable> callback);
+
+    BiConsumer<R, Throwable> callback();
+
     interface Handler
     {
-        void handle(LocalMessage message, Node node);
+        void handle(LocalRequest<?> message, Node node);
     }
-
-    void process(Node node);
 }
--- a/accord-core/src/main/java/accord/messages/LocalRequest.java
+++ b/accord-core/src/main/java/accord/messages/LocalRequest.java
@@ -26,6 +26,11 @@ import java.util.function.BiConsumer;
 
 public interface LocalRequest<R> extends Request, PreLoadContext, MapReduceConsume<SafeCommandStore, Void>
 {
+    /**
+     * Process the request without executing the callback
+     */
+    void process(Node on);
+
     void process(Node on, BiConsumer<R, Throwable> callback);
 
     BiConsumer<R, Throwable> callback();
--- a/accord-core/src/main/java/accord/messages/MessageType.java
+++ b/accord-core/src/main/java/accord/messages/MessageType.java
@@ -31,49 +31,52 @@ import static accord.messages.MessageType.Kind.LOCAL;
  */
 public class MessageType
 {
-    public static final MessageType SIMPLE_RSP                       = mt(REMOTE, false);
-    public static final MessageType FAILURE_RSP                      = mt(REMOTE, false);
-    public static final MessageType PRE_ACCEPT_REQ                   = mt(REMOTE, true );
-    public static final MessageType PRE_ACCEPT_RSP                   = mt(REMOTE, false);
-    public static final MessageType ACCEPT_REQ                       = mt(REMOTE, true );
-    public static final MessageType ACCEPT_RSP                       = mt(REMOTE, false);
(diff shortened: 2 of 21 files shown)


accord-core/src/main/java/accord/messages/Request.java
```diff
@@ -23,7 +23,16 @@
...
-    void process(Node on, Id from, ReplyContext replyContext);
     default long waitForEpoch() { return 0; }
     default long knownEpoch() { return waitForEpoch(); }
+    void preProcess(Node on, Id from, ReplyContext replyContext);
+    void process(Node on, Id from, ReplyContext replyContext);
+
+    /**
+     * Process the request without replying back
+     */
+    default void process(Node on)
```
aweisberg: So many processes makes this a little messy and it's spreading to other classes implementing `no-op` methods.

Since propagate is the only class that does `process(Node on)` and it goes to a dedicated `LocalRequest.Hander` could that not be part of `Request`?

I don't know what `preProcess` and `process` being split is for, I imagine it is clear on the integration side.

accord-core/src/main/java/accord/messages/LocalRequest.java
```diff
@@ -19,13 +19,19 @@
...
+
+    BiConsumer<R, Throwable> callback();
+
     interface Handler
     {
-        void handle(LocalMessage message, Node node);
+        void handle(LocalRequest<?> message, Node node);
     }
-
-    void process(Node node);
```
aweisberg, line 30: This is weird because we call this now from `process(Node on, BiConsumer<Status.Known, Throwable> callback)` so why have both?

accord-core/src/main/java/accord/messages/MessageType.java
```diff
@@ -98,25 +101,40 @@ public enum Kind { LOCAL, REMOTE }
...
 
-    /**
-     * LOCAL messages are not sent to remote nodes.
-     */
+    protected static MessageType remote(String name, boolean hasSideEffects)
+    {
+        return new MessageType(name, REMOTE, hasSideEffects);
+    }
+
+    private final String name;
```
aweisberg, line 114: FYI Benedict wanted to change this back to an enum again have multiple enums that share a common interface so the integrations can add more types.

accord-core/src/main/java/accord/messages/Propagate.java
```diff
@@ -212,6 +224,18 @@ else if (achieved.deps.hasProposedOrDecidedDeps())
...
+    @Override
+    public void preProcess(Node on, Node.Id from, ReplyContext replyContext)
+    {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void process(Node on, Node.Id from, ReplyContext replyContext)
+    {
+        throw new UnsupportedOperationException();
```
aweisberg, line 236: Heh, some dance is going on here with the various versions of process I just haven't figured it out yet.

aweisberg (changes requested):
Sorry I took so longer for what should have been a quick review. Had some questions about the proliferation of `Request.process` methods.

accord-core/src/main/java/accord/messages/Request.java
```diff
@@ -23,7 +23,16 @@
...
+    {
+        throw new UnsupportedOperationException();
```
aweisberg: I think this belongs in the `LocalRequest` interface not the `Request` interface.

accord-core/src/main/java/accord/messages/Request.java
```diff
@@ -23,7 +23,16 @@
 
 public interface Request extends Message
 {
-    void process(Node on, Id from, ReplyContext replyContext);
     default long waitForEpoch() { return 0; }
     default long knownEpoch() { return waitForEpoch(); }
+    void preProcess(Node on, Id from, ReplyContext replyContext);
```
aweisberg, line 27: This is used to facilitate an external error handling path that combines with `MapReduceConsume`.

Seems like it would be better interface wise to add a method that takes all these parameters and the error, but I can see how that would lead to duplication and wouldn't be a ton clearer.

accord-core/src/main/java/accord/messages/LocalRequest.java
aweisberg: I don't understand what I was asking here...

accord-core/src/main/java/accord/messages/Request.java
iamaleksey: It's for replay on startup, when the request needs to be processed, but a response is no longer expected.

accord-core/src/main/java/accord/messages/Request.java
iamaleksey: Yes. I've looked into alternatives and they seemed no better ¯\_(ツ)_/¯

accord-core/src/main/java/accord/messages/Request.java
iamaleksey: But to avoid confusion, I moved it to `LocalRequest` for now, as suggested. Will put the required change for playback on startup in the next PR, where its purpose will be more clear. Cheers.

aweisberg (approved):
+1 TY

PR #77, opened 2023-12-05
Add isSafeToPurge(TxnId) to DurableBefore

aweisberg:


accord-core/src/main/java/accord/local/DurableBefore.java
```diff
@@ -191,6 +191,11 @@ public Durability get(TxnId txnId, RoutingKey participant)
         return entry == null ? NotDurable : entry.get(txnId);
     }
 
+    public boolean isUniversal(TxnId txnId)
+    {
+        return min(txnId) == UniversalOrInvalidated;
```
aweisberg: @belliottsmith should this be `UniversalOrInvalidated` or both `Universal` and `UniversalOrInvalidated`. We use this from the accord compaction purger code to determine if something is universally durable and can be dropped.

belliottsmith: This class never returns `Universal` because it cannot make such a guarantee. It only has summary information, so it only knows that one or the other happened.

belliottsmith (approved):

PR #78, opened 2024-01-08
CASSANDRA-19253: (Accord) NPE while trying to serialize FoundKnownMap as value is null half the time but unexpected while serializing

dcapwell:


accord-core/src/main/java/accord/impl/CommandTimeseries.java
```diff
@@ -130,12 +130,12 @@ public Timestamp minTimestamp()
         {
             if (result == null)
             {
-                result = Timestamp.min(loader.txnId(data), loader.executeAt(data));
+                result = Timestamp.nonNullOrMin(loader.txnId(data), loader.executeAt(data));
```
dcapwell, line 133: also fixed this in [`https://issues.apache.org/jira/browse/CASSANDRA-18804](https://issues.apache.org/jira/browse/CASSANDRA-18804) but hit it in burn tests so fixed in this patch as well.

bdeggleston (approved):
+1

PR #80, opened 2024-01-31
CASSANDRA-19305: Ephemeral reads

belliottsmith:
Introduce faster per-key linearizable reads (suitable for any single key transactional read, or weaker multi-key reads). These guarantee 1 WAN RT, require minimal state, and offer single replica consistent execution. This is achieved by only calculating dependencies for execution, no execution timestamp.


accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
```diff
@@ -210,92 +115,32 @@ public synchronized void onSuccess(Id from, PreAcceptReply reply)
...
-            if (equivalent)
-            {
-                onPreAccepted(topologies, executeAt, successes);
-            }
-            else
-            {
-                extraPreAccept = new ExtraPreAccept(prevTopologies.currentEpoch() + 1, executeAt.epoch());
-                extraPreAccept.start();
-            }
+        Propose.Invalidate.proposeInvalidate(node, new Ballot(node.uniqueNow()), txnId, route.someParticipatingKey(), (outcome, failure) -> {
```
aweisberg, line 135: Really needs a comment explaining why in this mismatch case we invalidate, and it looks like we return an error the client?

accord-core/src/main/java/accord/coordinate/CoordinateEphemeralRead.java
```diff
@@ -0,0 +1,124 @@
...
+import accord.primitives.Txn;
+import accord.primitives.TxnId;
+import accord.topology.Topologies;
+import accord.utils.async.AsyncResult;
+import accord.utils.async.AsyncResults;
+
+import static accord.coordinate.tracking.RequestStatus.Failed;
+import static accord.coordinate.tracking.RequestStatus.Success;
+
+public class CoordinateEphemeralRead extends AbstractCoordinatePreAccept<Result, GetEphemeralReadDepsOk>
```
aweisberg, line 56: Needs a comment explaining how an ephemeral read is coordinated and what it does differently.

When we talked about it some key differences that came our were:
1. The ephemeral read isn't witnessed by other transactions and doesn't create persistent transaction state
2. Only does the preaccept round (and extra rounds), not accept because...
3. Doesn't pick an `executeAt` per se it just uses its original transaction id.
4. It waits for the deps it finds to execute and then executes
5. It does not run Apply since no one is waiting for apply and there is no persistent state

accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
```diff
@@ -0,0 +1,252 @@
...
+            for (long epoch = topologies.currentEpoch() ; equivalent && epoch > prevTopologies.currentEpoch() ; --epoch)
+                equivalent = topologies.forEpoch(epoch).shards().equals(prevTopologies.current().shards());
+
+            if (equivalent)
+            {
+                onPreAccepted(topologies);
+            }
+            else
+            {
+                extraEpochs = new ExtraEpochs(prevTopologies.currentEpoch() + 1, latestEpoch);
```
aweisberg, line 224: Just so I understand, we might create multiple `ExtraEpochs` if the `executeAt` keeps moving forward each time we attempt `PreAccept`?

accord-core/src/main/java/accord/coordinate/CoordinateEphemeralRead.java
```diff
@@ -0,0 +1,124 @@
...
+    void onNewEpochTopologyMismatch(TopologyMismatch mismatch)
+    {
+        accept(null, mismatch);
+    }
+
+    @Override
+    void onPreAccepted(Topologies topologies)
+    {
+        Deps deps = Deps.merge(oks, ok -> ok.deps);
+        new ExecuteEphemeralRead(node, topologies, route, txnId, txn, executeAtEpoch, deps, this).start();
```
aweisberg, line 135: Can we bump the Epoch in the `txnid` for the read to the `executeAtEpoch` so that when the read happens it can check if the epoch has a migration happening?

I think that would address the issue of doing a paxos write with async commit, then doing an Accord read at a different node that doesn't know about migrating to Paxos and having Accord accept the read. With the later Epoch Cassandra will reject the read in `TxnQuery`.

I think this gives you read your own writes in this base case. There might be more compleax cases where it still doesn't hold because the read can still occur in a later epoch, but it at least puts in a floor.

aweisberg (changes requested):
Close to +1 just one concern about execution not really taking place in a specific epoch determined during preaccept/accept. It executes in an arbitrary later epoch where migration could ping back and forth for the key (in theory).

accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
belliottsmith: Correct.

belliottsmith: Note that this can only happen under ownership changes, so the new owners would need to have agreed some newer transaction. I think in practice this is almost impossible, as the new epochs would have to have initiated an `ExclusiveSyncPoint` to be able to begin coordinating the ranges, so the `TxnId` we're coordinating would likely reject it - but it is possible the coordinator proposed a future `TxnId`.

At some point it will be worth elaborating a control flow diagram for all of this stuff.

accord-core/src/main/java/accord/coordinate/CoordinateEphemeralRead.java
belliottsmith: Hmm, maybe what we want then is to use `TxnId.withEpoch(executeAtEpoch)` as the `executeAt`?

belliottsmith: Could you point me to where this check happens?

aweisberg: It's here https://github.com/apache/cassandra/blob/00172b5fa9a935897a30b918964c2eaa4ccfd556/src/java/org/apache/cassandra/service/accord/txn/TxnQuery.java#L155

Thinking through this again. An ephemeral read needs to not read something already migrated back to Paxos, or read something not yet migrated to Accord. In Paxos -> Accord I think this works because there are no ephemeral reads so the interop checks at prepare/propose should catch it. 

### Accord -> Paxos 
In Accord -> Paxos a coordinator that doesn't know about the migration can create an ephemeral read and get it accepted because Accord doesn't know about migration only the integration does and the integration only checks when the read actually tries to run. The current approach just assumes that all keys are no longer readable/writable in the migration epoch regardless of whether the key has been migrated.

Because ephemeral reads aren't visible to other transactions (such as the barrier migrating the key to Paxos) the ephemeral read could run after the key is migrated to Paxos. It could run at any time really so it could end up executing in a later epoch even if the `executeAt` is in an earlier epoch.

However since it is a single key and single replica read maybe we can validate that the ephemeral read is safe in the integration by checking if a Paxos write occurred? It would need to block other Paxos writes while it reads the key. This is basically substituting for the ephemeral read not blocking other txns inside Accord.

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
aweisberg: My concern here returning an error to the client is that we surface an error for a read that should still be able to succeed if retried? I don't think a topology change is a good reason for that?

aweisberg: Another question that comes to mind is that the actual execution isn't guaranteed to take place in the `executeAt` epoch so presumably when we read a given key or range there is validation that the node still actually has that data?

Or are we relying on the node not having done any cleanup so it might have stale data, but that still meets the per key linearizability bar?

This brings up back to cleanup and how it will integrate with Accord. Since ephemeral reads aren't visible to other transactions it's not like you can run a barrier to know what data is safe to drop?

belliottsmith: It cannot succeed because the data involved no longer exists. A topology mismatch means the ranges involved are partially not present anymore.

Regarding reading data that is no longer valid, replicas already have a concept of stale ranges. When we drop data we can mark the range as stale for the command stores that previously owned them, so any ephemeral read that contacts them will understand the replica was unable to satisfy the request (this is already plumbed in).

We don’t currently have any support for dropping data really, I will leave a TODO related to this and file a ticket.

We also need to have ephemeral reads detect that they need to execute in a later epoch, but I planned to address that later. But I might look at that next, followed by asynchronous dependency handling.

accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
aweisberg: What do you mean the TxnId we are coordinating would likely reject it? 

Seems like for ephemeral reads we technically only have to do this once because we don't care about performing a stale read from Accord's perspective it just has to gather more dependencies if the nodes it originally contacted might not have known them all.

accord-core/src/main/java/accord/primitives/Txn.java
```diff
@@ -96,34 +108,64 @@ enum Kind
...
+        Kind(char shortName)
+        {
+            this.shortName = shortName;
+        }
 
         public enum Kinds implements Predicate<Kind>
         {
-            Ws, RorWs, WsOrSyncPoint, SyncPoints, AnyGlobal, Any;
+            Nothing,
+            Ws,
```
aweisberg, line 132: Ws is not a durable write? Do they witness ephemeral reads?

belliottsmith: I don't follow. Ws just means writes. Nobody witnesses ephemeral reads.

accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
belliottsmith: > Seems like for ephemeral reads we technically only have to do this once

In theory there might be a chain of epochs we haven't witnessed that were created before we started, and each might have registered transactions we must witness. For blind writes it might even be possible for them to reach agreement without waiting for earlier epochs to finish, so they might have been acknowledged before we started and need to be witnessed.

> What do you mean the TxnId we are coordinating would likely reject it?

Before we "close" an earlier epoch (stop routing queries to both it and the newer epoch) we coordinate an exclusive sync point and register all of its dependencies as historical transactions. The act of doing this prevents the lower TxnId from being agreed (its coordinator will propose an invalidation).

Though, in fairness, we don't need to use exclusive sync points for this, and probably we should use normal sync points.

accord-core/src/main/java/accord/primitives/Txn.java
aweisberg: This just seems like it should be included in the comment below (and be below the comment).

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
aweisberg: Ah I get it. The `ReadCoordinator` would then send additional reads until it finds a replica or has to fail.

accord-core/src/main/java/accord/primitives/Txn.java
belliottsmith: I’m confused. Ws is the set of write operations - that doesn’t include any reads, only writes. RorWs needs to specify it doesn’t see one type of read.

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -430,6 +430,22 @@ private static void stableRecipientLocalSyncPoint(SafeCommandStore safeStore, Tx
...
+        Command command = safeCommand.current();
+        if (command.hasBeen(Stable))
+            return;
+
+        Ranges coordinateRanges = coordinateRanges(safeStore, txnId);
+        // TODO (desired, consider): in the case of sync points, the coordinator is unlikely to be a home shard, do we mind this? should document at least
+        ProgressShard progressShard = No;
+        Invariants.checkState(validate(command.status(), command, Ranges.EMPTY, coordinateRanges, progressShard, route, Set, partialTxn, Set, partialDeps, Set));
+        CommonAttributes attrs = set(safeStore, command, command, Ranges.EMPTY, coordinateRanges, progressShard, route, partialTxn, Set, partialDeps, Set);
+        safeCommand.stable(attrs, Ballot.ZERO, txnId, initialiseWaitingOn(safeStore, txnId, Timestamp.MAX, attrs.partialDeps(), route));
```
aweisberg, line 445: I think here is where the `executeAt` needs to be instead of the txnid.

accord-core/src/main/java/accord/coordinate/CoordinateEphemeralRead.java
aweisberg: To close the loop on this. Ephemeral reads *do* execute in their `executeAt` epoch, it's validated according to Benedict. I haven't found where yet though. So that means the code in `TxnQuery` that validates things will be able to work.

accord-core/src/main/java/accord/primitives/Txn.java
aweisberg: OK, I was also confused because I was thinking that comment was intended to apply to all the `Kinds` below the comment, but it only applies to `RorWs`.

The positioning of the comment makes it seem like everything above it witnesses ephemeral reads and everything below does not. In reality everything doesn't witness them.

aweisberg (approved):
+1 TY!

PR #83, opened 2024-03-04
CASSANDRA-19355: Refactort CommandsForRange to work based off disk state rather than in-memory state

dcapwell:


accord-core/src/test/java/accord/utils/Gens.java
```diff
@@ -116,6 +125,253 @@ public static Gen.IntGen pickZipf(int[] array)
...
+        int i = Arrays.binarySearch(weights, rs.nextFloat());
+        if (i < 0) i = -1 - i;
+        return i;
+    }
+
+    public static Gen<Gen.IntGen> mixedDistribution(int minInclusive, int maxExclusive)
+    {
+        int domainSize = (maxExclusive - minInclusive + 1);
+        if (domainSize < 0)
+            throw new IllegalArgumentException("Range is too large; min=" + minInclusive + ", max=" + maxExclusive);
```
dcapwell, line 242: I could support but figured its best to keep this logic simpler until thats needed... if you want `mixDistribution(Long.MIN_VALUE, Long.MAX_VALUE)` we need `BigInteger` to handle this...

accord-core/src/test/java/accord/utils/Gens.java
```diff
@@ -116,6 +125,253 @@ public static Gen.IntGen pickZipf(int[] array)
...
+        int[] array, indexes;
+        if (domainSize > 200) // randomly selected
+        {
+            int numBuckets = 10;
```
dcapwell, line 246: when it comes to `randomWeight` and `zipF` having larger arrays doesn't make too much sense, so this puts a bias into a smaller array and generates values within that section of the range.  

I could make this a config, but I didn't need so can refactor later if needed

accord-core/src/test/java/accord/utils/SearchableRangeListTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import accord.impl.IntKey;
+import accord.primitives.Range;
+import org.assertj.core.api.Assertions;
+
+import static accord.utils.Property.qt;
+
+class SearchableRangeListTest
```
dcapwell, line 34: this is kinda duplicated in `org.apache.cassandra.utils.RTreeTest#test` but has better coverage... I wrote this class first and feel its always best to have tests at this level / project rather than depending on the C* test...

buildSrc/src/main/groovy/accord.java-conventions.gradle
```diff
@@ -29,7 +29,7 @@ repositories {
 }
 
 compileJava {
-    sourceCompatibility = JavaVersion.VERSION_1_8
+    sourceCompatibility = JavaVersion.VERSION_11
```
dcapwell, line 32: we actually don't compile on jdk8... so this just makes it so we can use JDK 11 features again

PR #84, opened 2024-03-29
Cassandra 19503

dcapwell:


accord-core/src/main/java/accord/coordinate/FetchMaxConflict.java
```diff
@@ -85,7 +85,7 @@ public static AsyncResult<Timestamp> fetchMaxConflict(Node node, Seekables<?, ?>
     @Override
     void contact(Set<Node.Id> nodes, Topologies topologies, Callback<GetMaxConflictOk> callback)
     {
-        node.send(nodes, to -> new GetMaxConflict(to, topologies, route, keysOrRanges, executionEpoch));
+        node.send(nodes, to -> new GetMaxConflict(to, topologies, route, keysOrRanges, executionEpoch), callback);
```
dcapwell, line 88: the fact this happened implies that burn test doesn't hit this code path, so should be something we think about adding... C* hits this all the time as we don't have a max applied, so we would block forever

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -231,16 +234,37 @@ void abort(Ranges abort)
         // TODO (expected): implement abort
     }
 
-    public static class FetchRequest extends WaitUntilAppliedAndReadData
```
dcapwell: there was a TODO to merge these 2 so I went ahead and did that

accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -231,16 +234,37 @@ void abort(Ranges abort)
...
+    public static class FetchRequest extends ReadData
     {
+        private static final ExecuteOn EXECUTE_ON = new ExecuteOn(Applied, Applied);
+        public final PartialTxn read;
```
dcapwell, line 240: honestly I kept this `PartialTxn` because it didn't impact anything and I needed to get green before really thinking about it... We only actually care about `accord.api.Read` so would be a trivial refactor to switch to that

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -587,7 +590,9 @@ public Topologies preciseEpochs(Unseekables<?> select, long minEpoch, long maxEp
     {
         Epochs snapshot = epochs;
 
-        TopologyMismatch tm = TopologyMismatch.checkForMismatch(snapshot.get(maxEpoch).global(), select);
+        EpochState maxState = snapshot.get(maxEpoch);
+        Invariants.checkState(maxState != null, "Unable to find epoch %d; known epochs are %d -> %d", maxEpoch, snapshot.minEpoch(), snapshot.currentEpoch);
```
dcapwell, line 594: in C* we bootstrap with epoch=9 but there are pending txn in epoch=8... this means that we are failing to handle those txn and SimpleProgressLog starts spamming us... I tried to look at how to add historic epochs but it was getting a bit annoying and didn't actually impact the bootstrap tests... so figured I could delay that problem to another patch

accord-core/src/test/java/accord/impl/list/ListRead.java
```diff
@@ -78,13 +78,17 @@ public AsyncChain<Data> read(Seekable key, SafeCommandStore safeStore, Timestamp
         {
             default: throw new AssertionError();
             case Key:
+                if (!keys.contains((Key)key))
```
dcapwell, line 81: the main reason bootstrap regression failed to be detected in burn test was that it switched to using the sync point's emptyTxn, which has 0 keys... since ListRead supports ranges it was able to return the data requested from bootstrap even though its inputs were outside of the range of `keys`... This check helps protect that from happening again.

PR #85, opened 2024-04-19
Accord IR support

bdeggleston:


aweisberg (approved):
Is it actually necessary to have the `executeAt` for `ApplyThenWaitUntilApplied` or is this just for hygiene where we don't discard whatever the adapter is supplied?

bdeggleston:
> Is it actually necessary to have the executeAt for ApplyThenWaitUntilApplied or is this just for hygiene where we don't discard whatever the adapter is supplied?

It's just for hygiene. Otherwise we start throwing inconsistent timestamp exceptions

PR #87, opened 2024-05-03
Cassandra 19618

dcapwell:
The following errors exist in `BurnTest`

```
//        run(516192490847500L); // Incomplete txn ({read:[]}) provided; does not cover [(857#0,857#13107], (857#13107,857#22936]]
//        run(517010705664541L); // Incomplete txn ({read:[]}) provided; does not cover [(0#54964,0#56246]]
//        run(517027263147375L); // Incomplete txn ({read:[]}) provided; does not cover [(739#19659,739#22932]]
//        run(517209858698166L); // Inconsistent execution timestamp detected for command Command@781010128{[2,3004,6(KS),16]:Applied}: [2,3004,6(KS),16] != [3,3013,10,2]
```

### DIFF
--- a/accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
+++ b/accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
@@ -96,6 +96,7 @@ abstract class AbstractCoordinatePreAccept<T, R> extends SettableResult<T> imple
     }
 
     final Node node;
+    @Nullable
     final TxnId txnId;
     final FullRoute<?> route;
 
@@ -219,7 +220,7 @@ abstract class AbstractCoordinatePreAccept<T, R> extends SettableResult<T> imple
                 onNewEpochTopologyMismatch(mismatch);
                 return;
             }
-            topologies = node.topology().withUnsyncedEpochs(route, txnId.epoch(), latestEpoch);
+            topologies = node.topology().withUnsyncedEpochs(route, txnId == null ? executeAtEpoch() : txnId.epoch(), latestEpoch);
             boolean equivalent = topologies.oldestEpoch() <= prevTopologies.currentEpoch();
             for (long epoch = topologies.currentEpoch() ; equivalent && epoch > prevTopologies.currentEpoch() ; --epoch)
                 equivalent = topologies.forEpoch(epoch).shards().equals(prevTopologies.current().shards());
--- a/accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
+++ b/accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
@@ -220,7 +220,7 @@ abstract class AbstractCoordinatePreAccept<T, R> extends SettableResult<T> imple
                 onNewEpochTopologyMismatch(mismatch);
                 return;
             }
-            topologies = node.topology().withUnsyncedEpochs(route, txnId == null ? executeAtEpoch() : txnId.epoch(), latestEpoch);
+            topologies = node.topology().withUnsyncedEpochs(route, earliestEpoch(), latestEpoch);
             boolean equivalent = topologies.oldestEpoch() <= prevTopologies.currentEpoch();
             for (long epoch = topologies.currentEpoch() ; equivalent && epoch > prevTopologies.currentEpoch() ; --epoch)
                 equivalent = topologies.forEpoch(epoch).shards().equals(prevTopologies.current().shards());
@@ -235,6 +235,11 @@ abstract class AbstractCoordinatePreAccept<T, R> extends SettableResult<T> imple
                 extraEpochs.start();
             }
         });
+    }
+
+    protected long earliestEpoch()
+    {
+        return txnId == null ? executeAtEpoch() : txnId.epoch();
     }
 
     @Override
--- a/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
+++ b/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
@@ -650,7 +650,7 @@ public abstract class InMemoryCommandStore extends CommandStore
     public static class InMemorySafeStore extends AbstractSafeCommandStore<InMemorySafeCommand, InMemorySafeTimestampsForKey, InMemorySafeCommandsForKey>
     {
         private final InMemoryCommandStore commandStore;
-        private final Map<TxnId, InMemorySafeCommand> commands;
+        protected final Map<TxnId, InMemorySafeCommand> commands;
         private final Map<RoutableKey, InMemorySafeTimestampsForKey> timestampsForKey;
         private final Map<RoutableKey, InMemorySafeCommandsForKey> commandsForKey;
         private final RangesForEpoch ranges;
--- a/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
+++ b/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
@@ -29,6 +29,7 @@ import java.util.NavigableMap;
 import java.util.Queue;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.TreeSet;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
@@ -99,6 +100,7 @@ public abstract class InMemoryCommandStore extends CommandStore
 
     private final TreeMap<TxnId, RangeCommand> rangeCommands = new TreeMap<>();
     private final TreeMap<TxnId, Ranges> historicalRangeCommands = new TreeMap<>();
+    private final TreeSet<TxnId> historicalTxnIds = new TreeSet<>();
     protected Timestamp maxRedundant = Timestamp.NONE;
 
     private InMemorySafeStore current;
@@ -106,6 +108,11 @@ public abstract class InMemoryCommandStore extends CommandStore
     public InMemoryCommandStore(int id, NodeTimeService time, Agent agent, DataStore store, ProgressLog.Factory progressLogFactory, EpochUpdateHolder epochUpdateHolder)
     {
         super(id, time, agent, store, progressLogFactory, epochUpdateHolder);
+    }
+
+    public Set<TxnId> historicalTxnIds()
+    {
+        return historicalTxnIds;
     }
 
     protected boolean canExposeUnloaded()
@@ -797,6 +804,7 @@ public abstract class InMemoryCommandStore extends CommandStore
                         return;
 
                     get(key).registerHistorical(this, txnId);
+                    commandStore.historicalTxnIds.add(txnId);
                 });
 
             });
@@ -815,6 +823,7 @@ public abstract class InMemoryCommandStore extends CommandStore
                     return;
 
                 historicalRangeCommands.merge(txnId, ranges.slice(allRanges), Ranges::with);
+                commandStore.historicalTxnIds.add(txnId);
             });
         }
 
--- a/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
+++ b/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
@@ -388,6 +388,8 @@ public abstract class InMemoryCommandStore extends CommandStore
         timestampsForKey.put(key, timestampsForKey((Key) key).createSafeReference());
     }
 
+    protected void validateRead(Command current) {}
+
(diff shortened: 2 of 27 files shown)


accord-core/src/main/java/accord/primitives/Routables.java
```diff
@@ -46,6 +48,10 @@ enum Slice
     int size();
 
     boolean isEmpty();
+    default Stream<K> stream()
```
dcapwell, line 51: this is dead code in the patch, but I was using it in debugger often; I can remove, just figured its useful...

accord-core/src/test/java/accord/impl/MessageListener.java
```diff
@@ -117,6 +127,18 @@ public void onClientAction(ClientAction action, Node.Id from, TxnId id, Object m
...
         }
 
+        private Topology previous = null;
+
+        @Override
+        public void onTopologyChange(Topology topology)
+        {
+            if (previous != null)
+            {
+                logger.debug("Topology Change {} -> {}: {}", previous.epoch(), topology.epoch(), TopologyUtils.diff(previous, topology));
```
dcapwell: honestly I can drop the diff... I mostly did it to confirm where things happened but personally found `ListStore.history` easier as it shows the local history (including bootstrap) for a single key... I did find the epoch notification helpful though to know when this changes

accord-core/src/main/java/accord/messages/Propagate.java
```diff
@@ -513,6 +513,7 @@ public MessageType type()
                 if (toEpoch >= committedExecuteAt.epoch())
                     return MessageType.PROPAGATE_APPLY_MSG;
             case Committed:
```
dcapwell, line 515: spoke with Benedict in slack, `Committed` might not make complete sense here...

accord-core/src/test/java/accord/impl/basic/Journal.java
```diff
@@ -0,0 +1,361 @@
...
+            requests.addAll(delayed);
+        }
+        waitForEpochs.removeIfLong(epoch -> !delayedRequests.containsKey(epoch));
+        
+        // for anything queued, put into the pending epochs or schedule
+        RequestContext request;
+        while (null != (request = unframedRequests.poll()))
+        {
+            long waitForEpoch = request.waitForEpoch;
+            if (waitForEpoch != 0 && !node.topology().hasEpoch(waitForEpoch))
```
dcapwell, line 256: `waitForEpoch != 0 && ` was also fixed in the `AccordJournal` though not sure we ever hit that... we did in burn test and know `epoch=0` isn't valid (its the empty state)

accord-core/src/main/java/accord/local/SerializerSupport.java
```diff
@@ -162,15 +167,15 @@ private static Command.Truncated truncated(Mutable attrs, SaveStatus status, Tim
             case TruncatedApplyWithDeps:
                 Set<MessageType> witnessed = messageProvider.test(APPLY_TYPES);
                 checkState(!witnessed.isEmpty());
-                if (witnessed.contains(APPLY_MINIMAL_REQ))
+                if (witnessed.contains(APPLY_MAXIMAL_REQ))
```
dcapwell, line 178: there was a bug where the `partialTxn` exists in `APPLY_MINIMAL_REQ` but because the next section is an `if` and not an `else if`  the `PROPAGATE_APPLY_MSG` was seen, but that doesn't have the `partialTxn`; this lead to a NPE...

I changed to use else for each section and go from max -> min -> propagate

accord-core/src/main/java/accord/local/SerializerSupport.java
```diff
@@ -231,6 +236,7 @@ private static <I, O> O extract(RangesForEpoch rangesForEpoch, SaveStatus status
             case AcceptedInvalidate:
             case Accepted:
             case PreCommitted:
+            {
```
dcapwell, line 257: txn/deps are used later on, so added so javac doesn't get mad

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -539,11 +544,65 @@ public static void main(String[] args)
...
+
+//        run(321453140939208L); // ISE in accord.local.SerializerSupport.truncated(SerializerSupport.java:171)
+//        run(325079839240208L); // ISE in accord.local.SerializerSupport.truncated(SerializerSupport.java:171)
+
+//        run(325076065173541L); // Invalid state: waiting for execution of command that is not owned at the execution time
+
+//        run(325459494008416L); // Writes is null
+    }
+
+    private static void run(long seed)
```
dcapwell, line 552: I never use `main` but could migrate this to include the operation so `main` has the same timeout logic.  I pulled this from `Property` class as its used to avoid loosing the seed when tests take longer than expected...

accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
```diff
@@ -219,7 +220,7 @@ final void onNewEpoch(Topologies prevTopologies, long latestEpoch)
                 onNewEpochTopologyMismatch(mismatch);
                 return;
             }
-            topologies = node.topology().withUnsyncedEpochs(route, txnId.epoch(), latestEpoch);
+            topologies = node.topology().withUnsyncedEpochs(route, txnId == null ? executeAtEpoch() : txnId.epoch(), latestEpoch);
```
belliottsmith: Perhaps introduce a helper method `earliestEpoch` or something (that could in theory be overridden) so this semantic is clearer?

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -99,6 +100,7 @@ public abstract class InMemoryCommandStore extends CommandStore
 
     private final TreeMap<TxnId, RangeCommand> rangeCommands = new TreeMap<>();
     private final TreeMap<TxnId, Ranges> historicalRangeCommands = new TreeMap<>();
+    private final TreeSet<TxnId> historicalTxnIds = new TreeSet<>();
```
belliottsmith: switch on using the `Invariants.DEBUG` flag?

accord-core/src/main/java/accord/local/Bootstrap.java
```diff
@@ -261,7 +261,11 @@ private void safeToReadCallback(SafeToRead state, Timestamp executeAt, Throwable
             else
             {
                 // TODO (expected): first check to see if we are still relevant
-                node.agent().onFailedBootstrap("SafeToRead", state.ranges, () -> started(state, null), failure);
+                CommandStore store = CommandStore.current();
+                node.agent().onFailedBootstrap("SafeToRead", state.ranges, () -> {
+                    if (CommandStore.maybeCurrent() == store) started(state, null);
```
belliottsmith: maybe we should have a method on `CommandStore` that accepts a `self` parameter and either executes the `Runnable` or `submits` it as necessary

accord-core/src/main/java/accord/local/SerializerSupport.java
```diff
@@ -285,14 +308,25 @@ else if (witnessed.contains(STABLE_FAST_PATH_REQ))
...
+                    {
+                        commit = messageProvider.commitMaximal();
+                    }
+                    else if (witnessed.contains(COMMIT_SLOW_PATH_REQ))
                     {
                         commit = messageProvider.commitSlowPath();
                     }
+                    else if (witnessed.contains(PRE_ACCEPT_REQ) || witnessed.contains(BEGIN_RECOVER_REQ) || witnessed.contains(PROPAGATE_PRE_ACCEPT_MSG))
+                    {
+                        Commit slowPath = messageProvider.stableSlowPath();
```
belliottsmith, line 322: should be `commitSlowPath`? We aren't meant to be stable here....

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
dcapwell: I don't actually plan to merge this, its just something I needed for trying to cover more states... I might be able to push this to the delayed store (which is test only)

accord-core/src/main/java/accord/local/SerializerSupport.java
dcapwell: That change hits NPE in `BurnTest`.  4/4 tests failed this way

```
java.lang.NullPointerException
		at accord.local.SerializerSupport.extract(SerializerSupport.java:323)
		at accord.local.SerializerSupport.committed(SerializerSupport.java:151)
		at accord.local.SerializerSupport.reconstruct(SerializerSupport.java:111)
```

this is due to `slowPath` not being present.  The test journal returns `null`, not sure what `AccordJournal` does when you ask for something not found... can always make it explicit `null` isn't allowed?

dcapwell: Also are we not in `case Stable:`?  So we *should* be stable?

belliottsmith: Oh, sorry - GitHub UX folding confused me

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -388,20 +395,32 @@ private void loadCommandsForKey(RoutableKey key,
...
+            validateRead(current);
+        }
 
         for (Seekable seekable : context.keys())
         {
             switch (seekable.domain())
             {
                 case Key:
                     RoutableKey key = (RoutableKey) seekable;
+                    if (context.keyHistory() == KeyHistory.NONE)
```
belliottsmith: Currently in C* we only load _either_ timestamps _or_ commands. So we should apply the same logic here too.

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -963,7 +970,7 @@ public boolean isEqualOrFuller(Command command)
 
     public static class PreAccepted extends AbstractCommand
     {
-        private final Timestamp executeAt;
+        private final @Nullable Timestamp executeAt;
```
belliottsmith, line 973: I don't think this should ever be null? If we're `AcceptedInvalidated` (or, logically, `PreAcceptedInvalidated` that doesn't currently exist as a reified state but is a side-state of `NotDefined`) then we shouldn't be a `PreAccepted` object.

accord-core/src/main/java/accord/primitives/Routables.java
belliottsmith: seems harmless enough

belliottsmith (approved):
I've only taken a cursory look at the journal changes and some of the additional validation, but the bug fixes all LGTM and should be merged once last comment addressed

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
dcapwell: removed.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
dcapwell: switched to

```
switch (context.keyHistory())
                    {
                        case NONE:
                            continue;
                        case COMMANDS:
                            commandsForKey.put(key, commandsForKey((Key) key).createSafeReference());
                            break;
                        case TIMESTAMPS:
                            timestampsForKey.put(key, timestampsForKey((Key) key).createSafeReference());
                            break;
                        default: throw new UnsupportedOperationException("Unknown key history: " + context.keyHistory());
                    }
```

accord-core/src/main/java/accord/local/Command.java
dcapwell: I wish I documented why I made this change.  There was a NPE I think in `equals` as it didn't handle `null`... let me see if I can find the code path.

dcapwell: here is a stack trace

```
<init>:985, Command$PreAccepted (accord.local)
<init>:1079, Command$Accepted (accord.local)
acceptInvalidated:1702, Command (accord.local)
acceptInvalidated:157, SafeCommand (accord.local)
acceptInvalidate:297, Commands (accord.local)
apply:263, Accept$Invalidate (accord.messages)
apply:240, Accept$Invalidate (accord.messages)
```

```
static Command.Accepted acceptInvalidated(Command command, Ballot ballot)
    {
        SaveStatus saveStatus = SaveStatus.get(Status.AcceptedInvalidate, command.known());
        return validate(new Command.Accepted(command, saveStatus, ballot, command.executeAt(), command.partialTxn(), null, ballot));
    }
```

PR #88, opened 2024-05-13
Move preaccept expiration logic away from Agent

iamaleksey:


belliottsmith (approved):

accord-core/src/main/java/accord/coordinate/CoordinateTransaction.java
```diff
@@ -84,7 +86,7 @@ void onPreAccepted(Topologies topologies, Timestamp executeAt, List<PreAcceptOk>
             //                                  but by sending accept we rule out hybrid fast-path
             // TODO (low priority, efficiency): if we receive an expired response, perhaps defer to permit at least one other
             //                                  node to respond before invalidating
-            if (executeAt.isRejected() || node.agent().isExpired(txnId, executeAt.hlc()))
+            if (executeAt.isRejected() || executeAt.hlc() - txnId.hlc() >= node.agent().preAcceptTimeout(MICROSECONDS))
```
dcapwell: this is true for C* but is this true for accord?  I thought `hlc` was kinda impl specific...

dcapwell: `accord.primitives.Timestamp#hlc`

```
    /**
     * A hybrid logical clock with implementation-defined resolution
     */
    public long hlc()
```

dcapwell (changes requested):
We currently define `hlc` resolution as being implementation specific and this patch assumes microsecond... this works for Cassandra but I don't think this works for accord as it breaks this contract...

accord-core/src/main/java/accord/coordinate/CoordinateTransaction.java
belliottsmith: Good catch

iamaleksey: Good point indeed. And it's actually trivially fixed and looks a bit simpler once addressed. Pushed and update.

ifesdjeen (approved):

ifesdjeen:
My approach to this was slightly different: I have exposed the time resolution via `Node`, but maybe we can just do this later if we see fit.

PR #89, opened 2024-05-14
CASSANDRA-18732 Baseline Diagnostic vtables for Accord

maedhroz:


accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -590,6 +592,20 @@ public <O> AsyncChain<O> mapReduce(PreLoadContext context, Routables<?> keys, lo
         return chain == null ? AsyncChains.success(null) : chain;
     }
 
+    public <O> Map<Integer, O> map(PreLoadContext context, Function<? super SafeCommandStore, O> mapper)
```
dcapwell: this adds a blocking call, something we can not do.  This *must* be `AsyncChain<Map<Integer, O>>`.

Also, not too much of a fan of `Map` rather than `List`...  the user function has access to the store so can include it in its output if needed... map just feels heavy weight and could reproduce the same results by using `accord.local.CommandStores#ids` with the list output (as they are the same order)


I would recommend the following instead

```
public <O> AsyncChain<List<O>> map(Function<? super SafeCommandStore, O> mapper)
    {
        return map(PreLoadContext.empty(), mapper);
    }

    public <O> AsyncChain<List<O>> map(PreLoadContext context, Function<? super SafeCommandStore, O> mapper)
    {
        ShardHolder[] shards = current.shards;
        List<AsyncChain<O>> results = new ArrayList<>(shards.length);
        for (ShardHolder shard : shards)
        {
            AsyncChain<O> next = shard.store.submit(context, mapper);
            results.add(next);
        }
        return AsyncChains.all(results);
    }
```

dcapwell (changes requested):

accord-core/src/main/java/accord/local/CommandStores.java
maedhroz: Done, essentially verbatim on the Accord side

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -58,7 +54,10 @@
 import accord.api.Scheduler;
 import accord.api.TopologySorter;
 import accord.config.LocalConfig;
+import accord.coordinate.CoordinateEphemeralRead;
```
dcapwell, line 57: why are you adding imports?  builds should be failing as we don't use these no?

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -589,14 +589,21 @@ protected void start(BiConsumer<? super T, Throwable> callback)
 
     public static <V> AsyncChain<V> ofCallable(Executor executor, Callable<V> callable)
     {
-        return new Head<V>()
+        return ofCallable(executor, callable, AsyncChains::encapsulate);
+    }
+
+    public static <V> AsyncChain<V> ofCallable(Executor executor,
```
dcapwell, line 595: why do we need to expose this method?

dcapwell (changes requested):

accord-core/src/main/java/accord/utils/async/AsyncChains.java
maedhroz: It's called from `TxnNamedRead#performLocalRead()`

accord-core/src/main/java/accord/local/Node.java
maedhroz: I didn't add this. I just moved it to the correct location from above.

dcapwell (approved):

PR #90, opened 2024-05-17
CASSANDRA-19642: IndexOutOfBoundsException while serializing CommandsForKey

dcapwell:


accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
```diff
@@ -266,22 +265,14 @@ public DelayedSafeStore(DelayedCommandStore commandStore, RangesForEpoch ranges,
         @Override
         public void postExecute()
         {
-            if (context instanceof Message)
```
dcapwell, line 269: not related to this patch... I added this logic to detect mutations outside of types with side effects, but really we should check that we can still load the command that we are trying to write...  this makes sure that we don't bias towards only checking for txn we re-load

accord-core/src/main/java/accord/local/CommandsForKey.java
```diff
@@ -161,6 +161,21 @@ public int compareTo(Unmanaged that)
...
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            Unmanaged unmanaged = (Unmanaged) o;
+            return pending == unmanaged.pending && waitingUntil.equals(unmanaged.waitingUntil) && txnId.equals(unmanaged.txnId);
+        }
+
+        @Override
+        public int hashCode()
+        {
+            return Objects.hash(pending, waitingUntil, txnId);
```
dcapwell, line 176: this is needed, tests store into a hash set

bdeggleston (approved):
+1

PR #91, opened 2024-05-20
Accord barrier/inclusive sync point fixes

aweisberg:
Patch by Ariel Weisberg, Benedict Eliott Smith; reviewed by David Capwell for CASSANDRA-19641

### DIFF
--- a/accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
+++ b/accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
@@ -128,7 +128,7 @@ public class CoordinateSyncPoint<S extends Seekables<?, ?>> extends CoordinatePr
             topologies = node.topology().forEpoch(route, txnId.epoch());
             // TODO (required): consider the required semantics of a SyncPoint
             if (tracker.hasFastPathAccepted() && txnId.kind() == Kind.SyncPoint)
-                execute(adapter, node, topologies, route, FAST, txnId, txn, executeAt, deps, this);
+                execute(adapter, node, topologies, route, FAST, txnId, txn, txnId, deps, this);
             else
                 propose(adapter, node, topologies, route, Ballot.ZERO, txnId, txn, executeAt, deps, this);
         }
--- a/accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
+++ b/accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
@@ -20,8 +20,6 @@ package accord.coordinate;
 
 import java.util.function.BiConsumer;
 
-import javax.annotation.Nullable;
-
 import accord.api.Result;
 import accord.coordinate.ExecuteSyncPoint.ExecuteBlocking;
 import accord.local.Node;
@@ -39,6 +37,7 @@ import accord.primitives.TxnId;
 import accord.primitives.Writes;
 import accord.topology.Topologies;
 import accord.utils.Faults;
+import javax.annotation.Nullable;
 
 import static accord.coordinate.ExecutePath.FAST;
 import static accord.coordinate.ExecutePath.SLOW;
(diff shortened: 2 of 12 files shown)


accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
```diff
@@ -128,7 +128,7 @@ void onPreAccepted(Topologies topologies, Timestamp executeAt, List<PreAcceptOk>
             topologies = node.topology().forEpoch(route, txnId.epoch());
             // TODO (required): consider the required semantics of a SyncPoint
             if (tracker.hasFastPathAccepted() && txnId.kind() == Kind.SyncPoint)
-                execute(adapter, node, topologies, route, FAST, txnId, txn, executeAt, deps, this);
+                execute(adapter, node, topologies, route, FAST, txnId, txn, txnId, deps, this);
```
belliottsmith, line 131: I don't think this is correct - sync points take an actual `executeAt`. If we intend to coordinate them like `ExclusiveSyncPoint` then we need to modify them to execute the same way (i.e. they need to update their kind to signify `awaitsOnlyDeps`, and need to use the same execution path as `ExclusiveSyncPoint` and `EphemeralRead`)

aweisberg: This is the fast path, when we talked about this you said that fast path pre-accept always uses the transaction ID, not executeAt. This is how the fast path works for regular transactions. Why would inclusive sync points need to function differently?

accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
```diff
@@ -261,21 +260,50 @@ public void execute(Node node, Topologies all, FullRoute<?> route, ExecutePath p
...
             @Override
             public void execute(Node node, Topologies all, FullRoute<?> route, ExecutePath path, TxnId txnId, Txn txn, Timestamp executeAt, Deps deps, BiConsumer<? super SyncPoint<S>, Throwable> callback)
             {
                 ExecuteBlocking<S> execute = ExecuteBlocking.atQuorum(node, all, new SyncPoint<>(txnId, deps, (S)txn.keys(), route), executeAt);
-                execute.addCallback(callback);
                 execute.start();
+                // If there are no deps to wait on then everything is already applied
+                // If this is the async adapter then we want to invoke the callback immediately
+                // and the caller can wait on the txn locally if they want
+                if (deps.isEmpty() || this == AsyncInclusiveSyncPointAdapter.INSTANCE)
```
belliottsmith: I don't think `deps.isEmpty()` is a good test to apply as in a real cluster this should almost never be true; also not super keen on testing equality with an instance object, should maybe just have an overridable method?

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
belliottsmith: Oh, yes. My mistake.

accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
aweisberg: If you have a table that isn't written to frequently won't it only take minutes for everything to be truncated so there are empty deps? I was just thinking of making repair a little faster, but yeah I could see that the usage is infrequent enough we don't even really care.

I reduced constructor visibility to go with the `==` but yeah I'll make it a method.

belliottsmith: It's never empty, it will always have some exclusive sync point that represents the point in time of truncation to ensure no stale replica serves the request.

Perhaps we can issue entirely empty deps if we know no such replica exists, but I am not sure it is a sensible optimisation given the potential additional surface area for bugs.

aweisberg: Updated it.

belliottsmith (approved):

PR #93, opened 2024-05-31
CASSANDRA-19674: CEP-15: (Accord) Bootstraps LocalOnly txn is not written to journal which causes issues once evicted from the cache

dcapwell:


accord-core/src/main/java/accord/local/Bootstrap.java
```diff
@@ -120,6 +125,7 @@ void start(SafeCommandStore safeStore0)
         {
             globalSyncId = node.nextTxnId(ExclusiveSyncPoint, Routable.Domain.Range);
             localSyncId = globalSyncId.as(LocalOnly).withEpoch(epoch);
+            Invariants.checkArgument(epoch <= globalSyncId.epoch(), "Attempting to use local epoch %d which is larger than global epoch %d", epoch, globalSyncId.epoch());
```
dcapwell, line 123: never hit this case, just being defensive

accord-core/src/main/java/accord/local/Bootstrap.java
```diff
@@ -136,19 +142,20 @@ void start(SafeCommandStore safeStore0)
                // TODO (required, correcness) : PreLoadContext only works with Seekables, which doesn't allow mixing Keys and Ranges... But Deps has both Keys AND Ranges!
                // TODO (required): is localSyncId even being used anymore
                // ATM all known implementations store ranges in-memory, but this will not be true soon, so this will need to be addressed
-               .flatMap(syncPoint -> node.withEpoch(epoch, () -> store.submit(contextFor(localSyncId, syncPoint.waitFor.keyDeps.keys(), KeyHistory.COMMANDS), safeStore1 -> {
+               .flatMap(syncPoint -> store.submit(empty(), safeStore1 -> {
+                   //TODO (deterministic reply): this changes due to errors and invalidate, so how do we handle this on reply?
```
dcapwell: example, we see `valid=[1, 2, 3]`, we schedule `CreateBootstrapCompleteMarkerTransaction` and *before* that runs we update `valid=[1]`... we *store* in journal `valid=[1]` so maybe its fine?

accord-core/src/main/java/accord/local/SerializerSupport.java
```diff
@@ -90,11 +93,80 @@ public class SerializerSupport
                 .addAll(APPLY_TYPES)
                 .build();
 
+    private static final Set<MessageType> BOOTSTRAP_TYPES = ImmutableSet.of(MessageType.BOOTSTRAP_ATTEMPT_MARK_BOOTSTRAP_COMPLETE,
+                                                                            MessageType.BOOTSTRAP_ATTEMPT_COMPLETE_MARKER);
+
+    private static Command localOnly(Agent agent, RangesForEpoch rangesForEpoch, Mutable attrs, SaveStatus status, Timestamp executeAt, @Nullable Timestamp executesAtLeast, Ballot promised, Ballot accepted, WaitingOnProvider waitingOnProvider, MessageProvider messageProvider)
```
dcapwell, line 97: I honestly hate this method, so if anyone can think of a better way im all ears... felt LocalOnly was too special cased in the code that trying to handle in multiple places was error prone

accord-core/src/main/java/accord/local/SerializerSupport.java
```diff
@@ -90,11 +93,80 @@ public class SerializerSupport
...
+            var keys = marker.valid;
+            var route = marker.syncPoint.route();
+
+            Txn emptyTxn = agent.emptyTxn(marker.localSyncId.kind(), keys);
+            Ranges coordinateRanges = rangesForEpoch.coordinates(attrs.txnId());
+            attrs.partialTxn(emptyTxn.slice(coordinateRanges, true))
+                 .partialDeps(Deps.NONE.slice(coordinateRanges));
+        }
+
+        switch (status.status)
```
dcapwell, line 112: honestly several of these status are not valid... the one thing is I don't know if `ProgressLog` is expected to ignore `LocalOnly` or not, so not sure if I can assume the 2 live status expected or not... so went with duplication of code to be safer...

accord-core/src/main/java/accord/messages/Propagate.java
```diff
@@ -59,13 +61,13 @@
 import static accord.utils.Invariants.illegalState;
 
 // TODO (required): detect propagate loops where we don't manage to update anything but should
-public class Propagate implements EpochSupplier, LocalRequest<Status.Known>
+public class Propagate implements EpochSupplier, LocalRequest<Status.Known>, PreLoadContext, MapReduceConsume<SafeCommandStore, Void>
```
dcapwell, line 64: this was the only `LocalRequest` and details of it leaked into the `LocalRequest` API and made it harder to add more.  I had to remove `PreLoadContext` and `MapReduceConsume` to do what I wanted for bootstrap steps, so had to pull all those into this class; hence the changes

accord-core/src/test/java/accord/impl/basic/Cluster.java
```diff
@@ -274,6 +278,84 @@ public void now(Runnable run)
         run.run();
     }
 
+    public static Map<MessageType, Cluster.Stats> run(Supplier<RandomSource> randomSupplier,
```
dcapwell, line 281: This was pulled out of `accord-core/src/test/java/accord/local/CommandsTest.java` so another test could use.

belliottsmith:
Given we're abandoning determinism, it might be we don't need to introduce all of this refactor. The only reason we need it today is for loading the transaction itself, but the only info we're missing is essentially just the ranges we're bootstrapping, which I think should be stored fully in the `Route`. So, it might be easier to just specialise reconstruction for now?

dcapwell:
> Given we're abandoning determinism

news to me, didn't know that.

> but the only info we're missing is essentially just the ranges we're bootstrapping, which I think should be stored fully in the Route

We store the `route` in the keyspace, so I can use that.  `localSyncId` is the `txnId` so we have that, so the following should work just fine using the `route`


```
Txn emptyTxn = agent.emptyTxn(marker.localSyncId.kind(), marker.ranges);
            writes = emptyTxn.execute(marker.localSyncId, marker.localSyncId, null);
            results = emptyTxn.result(marker.localSyncId, marker.localSyncId, null);
            Ranges coordinateRanges = rangesForEpoch.coordinates(attrs.txnId());
            attrs.partialTxn(emptyTxn.slice(coordinateRanges, true))
                 .partialDeps(Deps.NONE.slice(coordinateRanges));
```

What I see is that the code doesn't perfectly make the `keys` and `route` a 1-to-1 in the code, so will see if `BurnTest` hits a case with them differing.  Our `route` is `syncPoint.route()` and keys is `Ranges valid` which can be mutated *after* the sync point

dcapwell:
@belliottsmith pushed change so we no longer use `LocalRequest` and kept the existing bootstrap logic.  I updated the reconstruct logic to use `Route` and tests are passing fine.

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -543,9 +544,11 @@ public static void main(String[] args)
         }
     }
 
-    @Test
+//    @Test
+    @RepeatedTest(Integer.MAX_VALUE)
```
dcapwell: ill revert, just doing this to help run in the background and collect broken seeds

accord-core/src/main/java/accord/local/SerializerSupport.java
belliottsmith: Yes, `ProgressLog` should not interact with `LocalOnly` transactions.

accord-core/src/main/java/accord/local/SerializerSupport.java
belliottsmith: Let's leave a `TODO (required)` or `TODO (expected)` to revisit this towards the end of development and clean it up.

belliottsmith (approved):

PR #96, opened 2024-06-06
ApplyThenWaitUntilApplied supplies wrong epoch for executeAtEpoch

aweisberg:
Patch by Ariel Weisberg; Reviewed by Benedict Elliott Smith for CASSANDRA-19687

### DIFF
--- a/accord-core/src/main/java/accord/local/SerializerSupport.java
+++ b/accord-core/src/main/java/accord/local/SerializerSupport.java
@@ -18,7 +18,6 @@
 package accord.local;
 
 import java.util.Set;
-
 import javax.annotation.Nullable;
 
 import com.google.common.collect.ImmutableSet;
@@ -92,6 +91,7 @@ public class SerializerSupport
     ImmutableSet.<MessageType>builder()
                 .addAll(PRE_ACCEPT_STABLE_TYPES)
                 .addAll(APPLY_TYPES)
+                .add(PROPAGATE_OTHER_MSG)
                 .build();
 
     private static Command localOnly(Agent agent, RangesForEpoch rangesForEpoch, Mutable attrs, SaveStatus status, Timestamp executeAt, @Nullable Timestamp executesAtLeast, Ballot promised, Ballot accepted, WaitingOnProvider waitingOnProvider, MessageProvider messageProvider)
--- a/accord-core/src/main/java/accord/messages/ApplyThenWaitUntilApplied.java
+++ b/accord-core/src/main/java/accord/messages/ApplyThenWaitUntilApplied.java
@@ -60,9 +60,9 @@ public class ApplyThenWaitUntilApplied extends WaitUntilApplied
     @SuppressWarnings("unused")
     public static class SerializerSupport
     {
-        public static ApplyThenWaitUntilApplied create(TxnId txnId, Participants<?> readScope, long executeAtEpoch, Timestamp executeAt, FullRoute<?> route, PartialTxn txn, PartialDeps deps, Writes writes, Result result, Seekables<?, ?> notify)
+        public static ApplyThenWaitUntilApplied create(TxnId txnId, Participants<?> readScope, Timestamp executeAt, FullRoute<?> route, PartialTxn txn, PartialDeps deps, Writes writes, Result result, Seekables<?, ?> notify)
         {
-            return new ApplyThenWaitUntilApplied(txnId, readScope, executeAtEpoch, executeAt, route, txn, deps, writes, result, notify);
+            return new ApplyThenWaitUntilApplied(txnId, readScope, executeAt, route, txn, deps, writes, result, notify);
         }
     }
 
@@ -74,9 +74,9 @@ public class ApplyThenWaitUntilApplied extends WaitUntilApplied
     public final Result result;
     public final Seekables<?, ?> notify;
 
-    public ApplyThenWaitUntilApplied(Node.Id to, Topologies topologies, Timestamp executeAt, FullRoute<?> route, TxnId txnId, Txn txn, Deps deps, Participants<?> readScope, long executeAtEpoch, Writes writes, Result result, Seekables<?, ?> notify)
+    public ApplyThenWaitUntilApplied(Node.Id to, Topologies topologies, Timestamp executeAt, FullRoute<?> route, TxnId txnId, Txn txn, Deps deps, Participants<?> readScope, Writes writes, Result result, Seekables<?, ?> notify)
     {
-        super(to, topologies, txnId, readScope, executeAtEpoch);
+        super(to, topologies, txnId, readScope, executeAt.epoch());
         this.executeAt = executeAt;
         Ranges slice = computeScope(to, topologies, null, 0, (i,r)->r, Ranges::with);
         this.route = route;
@@ -87,9 +87,9 @@ public class ApplyThenWaitUntilApplied extends WaitUntilApplied
         this.notify = notify == null ? null : notify.slice(slice);
     }
 
-    protected ApplyThenWaitUntilApplied(TxnId txnId, Participants<?> readScope, long executeAtEpoch, Timestamp executeAt, FullRoute<?> route, PartialTxn txn, PartialDeps deps, Writes writes, Result result, Seekables<?, ?> notify)
+    protected ApplyThenWaitUntilApplied(TxnId txnId, Participants<?> readScope, Timestamp executeAt, FullRoute<?> route, PartialTxn txn, PartialDeps deps, Writes writes, Result result, Seekables<?, ?> notify)
     {
-        super(txnId, readScope, executeAtEpoch);
+        super(txnId, readScope, executeAt.epoch());
         this.executeAt = executeAt;
         this.route = route;
         this.txn = txn;
--- a/accord-core/src/main/java/accord/primitives/RangeDeps.java
+++ b/accord-core/src/main/java/accord/primitives/RangeDeps.java
@@ -18,24 +18,55 @@
 
 package accord.primitives;
 
-import accord.api.Key;
-import accord.api.RoutingKey;
-import accord.utils.*;
-import accord.utils.SortedArrays.SortedArrayList;
-import net.nicoulaj.compilecommand.annotations.DontInline;
-import net.nicoulaj.compilecommand.annotations.Inline;
-
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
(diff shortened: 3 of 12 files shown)


accord-core/src/main/java/accord/local/SerializerSupport.java
```diff
@@ -88,6 +87,7 @@ public class SerializerSupport
     ImmutableSet.<MessageType>builder()
                 .addAll(PRE_ACCEPT_STABLE_TYPES)
                 .addAll(APPLY_TYPES)
+                .add(PROPAGATE_OTHER_MSG)
```
dcapwell, line 94: is this true?  `accord.messages.Propagate#type`

```
    @Override
    public MessageType type()
    {
        // TODO (now): this logic doesn't work now we permit upgrading; need to pick the maximum *possible* status we might propagate
        //     might be better to
        switch (achieved.propagatesStatus())
        {
            case Applied:
            case PreApplied:
                if (toEpoch >= committedExecuteAt.epoch())
                    return MessageType.PROPAGATE_APPLY_MSG;
            case Committed:
            case Stable:
                return MessageType.PROPAGATE_STABLE_MSG;
            case PreCommitted:
                if (!achieved.definition.isKnown())
                    return MessageType.PROPAGATE_OTHER_MSG;
            case PreAccepted:
                return MessageType.PROPAGATE_PRE_ACCEPT_MSG;
            default:
                return MessageType.PROPAGATE_OTHER_MSG;
        }
    }
```
    
 if we reach applied we would be `PROPAGATE_STABLE_MSG` or `PROPAGATE_APPLY_MSG`.  This means that the only cases where this happens are below


Screenshot 2024-06-10 at 1 06 25 PM

accord-core/src/main/java/accord/primitives/RangeDeps.java
```diff
@@ -18,24 +18,55 @@
...
+import accord.utils.RelationMultiMap;
+import accord.utils.SearchableRangeList;
+import accord.utils.SortedArrays;
+import accord.utils.SortedArrays.SortedArrayList;
+import accord.utils.SymmetricComparator;
+import accord.utils.TriFunction;
+import net.nicoulaj.compilecommand.annotations.DontInline;
+import net.nicoulaj.compilecommand.annotations.Inline;
 
-import static accord.utils.ArrayBuffers.*;
```
dcapwell, line 37: silly `.*`!

accord-core/src/main/java/accord/utils/CheckpointIntervalArray.java
```diff
@@ -213,9 +213,6 @@ else if ((ri & BIT29) != 0)
                 forEachScanOrCheckpoint.accept(p1, p2, p3, p4, i);
         }
 
-        if (start == end)
```
dcapwell, line 216: we spoke in slack, just putting it here.  The disk based version needed this for a reason I can't remember, so that would need to be updated to handle this case.

accord-core/src/test/java/accord/impl/list/ListStore.java
```diff
@@ -512,6 +513,8 @@ public static AsyncChain<SyncPoint<Ranges>> coordinate(Node node, long minEpoch,
             Await coordinate = new Await(node, minEpoch, sp);
             coordinate.start();
             return coordinate.recover(t -> {
+                if (t.getClass() == SyncPointErased.class)
```
dcapwell, line 516: TIL: `SyncPointErased`.... fancy...

accord-core/src/test/java/accord/impl/list/ListStore.java
```diff
@@ -512,6 +513,8 @@ public static AsyncChain<SyncPoint<Ranges>> coordinate(Node node, long minEpoch,
...
+                    return AsyncChains.success(null);
```
dcapwell, line 517: this code is consistent with the rest of the function, so its fine... but relooking at it `return AsyncChains.success(null)` feels wrong as the only caller doesn't handle that case!  Its there for debug logging, so if we actually log we will hit a NPE I think...

nothing for you to change as you are consistent... but a bad access will lead to a NPE if we hit this or `Redundant`

dcapwell (commented):
ill leave the real +1 to @belliottsmith as. I am not sure if the core change is correct or not.  Just doing a drive by for the small things I saw

accord-core/src/main/java/accord/local/SerializerSupport.java
belliottsmith: We should only maybe propagate `Truncated` or `Invalidated` out of those

accord-core/src/main/java/accord/messages/ApplyThenWaitUntilApplied.java
```diff
@@ -87,9 +87,9 @@ public ApplyThenWaitUntilApplied(Node.Id to, Topologies topologies, Timestamp ex
         this.notify = notify == null ? null : notify.slice(slice);
     }
 
-    protected ApplyThenWaitUntilApplied(TxnId txnId, Participants<?> readScope, long executeAtEpoch, Timestamp executeAt, FullRoute<?> route, PartialTxn txn, PartialDeps deps, Writes writes, Result result, Seekables<?, ?> notify)
+    protected ApplyThenWaitUntilApplied(TxnId txnId, Participants<?> readScope, Timestamp executeAt, FullRoute<?> route, PartialTxn txn, PartialDeps deps, Writes writes, Result result, Seekables<?, ?> notify)
     {
-        super(txnId, readScope, executeAtEpoch);
```
belliottsmith, line 92: iirc ExclusiveSyncPoints maybe supply a different execution epoch. BUT I need to revisit how ExclusiveSyncPoints are executed anyway, to simplify some other stuff (and also fix some minor correctness issues across range movements). So, if this doesn't clearly break anything don't worry about it, they'll get their own dedicated execution class when I do that.

belliottsmith (approved):

accord-core/src/main/java/accord/primitives/RangeDeps.java
aweisberg: Are you asking to keep the `.*`? Cassandra code style is to use explicit imports if it's < 100!

accord-core/src/main/java/accord/utils/CheckpointIntervalArray.java
aweisberg: How can I follow up on this? Is there a test that should be failing or something I should look at?

accord-core/src/main/java/accord/primitives/RangeDeps.java
dcapwell: nope, not asking to keep it, more calling out its a good find... your patch is good here, no work to be done =)

accord-core/src/main/java/accord/messages/ApplyThenWaitUntilApplied.java
aweisberg: Currently this is only used by inclusive sync points. Exclusive uses `WaitUntilApplied` and provides the `TxnId` as the `executionEpoch`.

accord-core/src/main/java/accord/utils/CheckpointIntervalArray.java
dcapwell: I don't remember but here are a few places you should run tests multiple times to see if anything breaks

`org.apache.cassandra.index.accord.CheckpointIntervalArrayIndexTest`
`org.apache.cassandra.utils.RangeTreeTest`
`accord.utils.SearchableRangeListTest`

dcapwell: I double checked and those tests touch this function.

accord-core/src/main/java/accord/local/SerializerSupport.java
aweisberg: Then what is the code in `SerializerSupport` supposed to be doing with `PROPAGATE_OTHER_MSG`? Because right now it's doesn't process it at all.

I can create a new type for truncated or invalidated, but then PROPAGATE_OTHER_MSG will be a no-op again.

accord-core/src/main/java/accord/utils/CheckpointIntervalArray.java
dcapwell:
```
java.lang.IndexOutOfBoundsException: Start is from (0, 10]; attempted to access 10
  at org.apache.cassandra.index.accord.CheckpointIntervalArrayIndex$SortedListReader.fileOffsetStart(CheckpointIntervalArrayIndex.java:336)
  at org.apache.cassandra.index.accord.CheckpointIntervalArrayIndex$SegmentSearcher.lambda$intersects$1(CheckpointIntervalArrayIndex.java:618)
  at accord.utils.CheckpointIntervalArray.forEach(CheckpointIntervalArray.java:219)
  at accord.utils.CheckpointIntervalArray.forEach(CheckpointIntervalArray.java:127)
  at org.apache.cassandra.index.accord.CheckpointIntervalArrayIndex$SegmentSearcher.intersects(CheckpointIntervalArrayIndex.java:612)
  at org.apache.cassandra.index.accord.CheckpointIntervalArrayIndexTest$1.intersects(CheckpointIntervalArrayIndexTest.java:410)
  at org.apache.cassandra.index.accord.CheckpointIntervalArrayIndexTest.fuzz(CheckpointIntervalArrayIndexTest.java:146)
  at org.apache.cassandra.index.accord.CheckpointIntervalArrayIndexTest.lambda$fuzzSmall$1(CheckpointIntervalArrayIndexTest.java:121)
  at accord.utils.Property$SingleBuilder.checkInternal(Property.java:245)
  at accord.utils.Property$SingleBuilder.lambda$check$0(Property.java:230)
  at accord.utils.async.TimeoutUtils.lambda$runBlocking$0(TimeoutUtils.java:39)
  at java.base/java.lang.Thread.run(Thread.java:829)

```

dcapwell: looks like a simple fix in `org.apache.cassandra.index.accord.CheckpointIntervalArrayIndex.SegmentSearcher#intersects`

```
searcher.forEach(start, end, (i1, i2, i3, i4, index) -> {
                    stats.matches++;
                    callback.accept(reader.copyTo(accessor.get(indexInput, index), buffer));
                }, (i1, i2, i3, i4, startIdx, endIdx) -> {
                    if (startIdx == endIdx)
                        return;
```

accord-core/src/main/java/accord/local/SerializerSupport.java
aweisberg: It's already referenced here https://github.com/apache/cassandra-accord/blob/trunk/accord-core/src/main/java/accord/local/SerializerSupport.java#L415 and seems to assume everything might be missing.

In this case what needed to propagate was invalidation.

belliottsmith:
+1

PR #98, opened 2024-06-14
C19288 accord async reads are unsafe

bdeggleston:


accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -359,11 +359,13 @@ void cancel(@Nullable SafeCommandStore safeStore)
             if (safeCommand != null) safeCommand.removeListener(this);
             waitingOn.clear(safeStore.commandStore().id());
         }
+
         // TODO (expected): efficient unsubscribe mechanism
-        node.commandStores().mapReduceConsume(this, waitingOn.stream(), forEach(in -> {
-            SafeCommand safeCommand = in.ifInitialised(txnId);
-            if (safeCommand != null) safeCommand.removeListener(this);
-        }, node.agent()));
+        if (waitingOn != null)
```
dcapwell, line 364: did you hit a NPE?  since we touch `waitingOn.stream()` we would hit a NPE.

dcapwell (commented):
am I right that async reads are fine, its just async ephemeral reads is where we have issues?  Else I don't see the real change that fixes async reads

accord-core/src/main/java/accord/messages/ReadData.java
bdeggleston: yes, if there are multiple calls to cancel it will throw an NPE

bdeggleston:
I think that async reads are safe by virtue of the execution process, but the assertion was assuming they were synchronous. Was there another issue with ephemeral reads? I'm less familiar with them, but it seems like it would be ok for an ephemeral read to read the from "future", so long as they don't go backwards in time

dcapwell:
We disable Ephemeral Reads in C* as they don't work (let alone a single cache evict will cause chaos), so thats fine by me.

PR #100, opened 2024-06-20
Cassandra 19719

dcapwell:


accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -110,6 +110,11 @@ int size()
             return epochs.size();
         }
 
+        public boolean isEmpty()
+        {
+            return lastReceived == 0;
```
dcapwell, line 115: we have epochs array but until its received we can't fetch it in `epochs.topologyFor`.  Can see this clearly here

```
public synchronized Topology currentTopology()
    {
        return epochs.topologyFor(epochs.lastReceived);
    }
```

accord-core/src/test/java/accord/utils/Gens.java
```diff
@@ -112,8 +113,18 @@ public static <T> Gen<T> pick(Map<T, Integer> values)
     {
         if (values == null || values.isEmpty())
             throw new IllegalArgumentException("values is empty");
+        // if 2 values have the same weight we need some way to tie-break, but that isn't always possible...
```
dcapwell, line 116: if no keys have the same value then we are fine, but there are cases where the values are randomized for each test... so figured its still best to block to specific classes just to be safe and find this issue earlier

accord-core/src/test/java/accord/utils/LoggingRandomSource.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.stream.Stream;
+
+/**
+ * This class is helpful for finding non-deterministic behavior as it tracks every call to {@link RandomSource} so that
+ * different runs can be compared to find where they differ.  The most common way to use this is to replace the tests
+ * {@link RandomSource} with this class and run it against 2 or more environments (that are experiencing a different outcome),
+ * then compare the logs to find where the first difference is (and the previous access is likely where the non-deterministic fault is).
+ */
+@SuppressWarnings("unused")
+public class LoggingRandomSource implements RandomSource
```
dcapwell, line 39: I wrote this to help find where the histories were drifting, and was able to find `oneOf` was at fault.  Rather than implementing w/e we find something non-deterministic can use this to help find it.

bdeggleston (approved):
+1

PR #101, opened 2024-07-02
CASSANDRA-19744 Accord migration and interop correctness

aweisberg:
[CASSANDRA-19744](https://issues.apache.org/jira/browse/CASSANDRA-19744)

The only Accord change here is adding timeouts to `TopologyManager`. We need those timeouts, but this was largely driven by it being basically untestable to have a coordinator be behind when attempting to execute an Accord transaction. It will learn about the new epoch from the other nodes during `PreAccept` and then block forever trying to adopt it so you can't test the timeout handling code paths in Cassandra for things like batches and hints.

### DIFF
--- a/accord-core/src/main/java/accord/config/LocalConfig.java
+++ b/accord-core/src/main/java/accord/config/LocalConfig.java
@@ -22,8 +22,22 @@ import java.time.Duration;
 
 public interface LocalConfig
 {
+    LocalConfig DEFAULT = new LocalConfig() {};
+
     default Duration getProgressLogScheduleDelay()
     {
         return Duration.ofSeconds(1);
+    }
+
+    // How long before we start notifying waiters on an epoch of timeout,
+    default Duration epochFetchInitialTimeout()
+    {
+        return Duration.ofSeconds(10);
+    }
+
+    // How often to check for timeout, and once an epoch has timed out, how often we timeout new waiters
+    default Duration epochFetchWatchdogInterval()
+    {
+        return Duration.ofSeconds(10);
     }
 }
--- a/accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
+++ b/accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
@@ -22,7 +22,6 @@ import java.util.Collection;
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.function.BiConsumer;
-
 import javax.annotation.Nullable;
 
 import accord.coordinate.tracking.QuorumTracker;
@@ -212,7 +211,12 @@ abstract class AbstractCoordinatePreAccept<T, R> extends SettableResult<T> imple
     {
         // TODO (desired, efficiency): check if we have already have a valid quorum for the future epoch
         //  (noting that nodes may have adopted new ranges, in which case they should be discounted, and quorums may have changed shape)
-        node.withEpoch(latestEpoch, () -> {
+        node.withEpoch(latestEpoch, (ignored, withEpochFailure) -> {
+            if (withEpochFailure != null)
+            {
+                tryFailure(CoordinationFailed.wrap(withEpochFailure));
+                return;
+            }
             TopologyMismatch mismatch = TopologyMismatch.checkForMismatch(node.topology().globalForEpoch(latestEpoch), txnId, route.homeKey(), keysOrRanges());
             if (mismatch != null)
             {
--- a/accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
+++ b/accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
@@ -152,7 +152,12 @@ abstract class CoordinatePreAccept<T> extends AbstractCoordinatePreAccept<T, Pre
     void onPreAccepted(Topologies topologies)
     {
         Timestamp executeAt = foldl(oks, (ok, prev) -> mergeMax(ok.witnessedAt, prev), Timestamp.NONE);
-        node.withEpoch(executeAt.epoch(), () -> onPreAccepted(topologies, executeAt, oks));
+        node.withEpoch(executeAt.epoch(), (ignored, withEpochFailure) -> {
+            if (withEpochFailure != null)
+                tryFailure(CoordinationFailed.wrap(withEpochFailure));
+            else
+                onPreAccepted(topologies, executeAt, oks);
+        });
     }
 
     abstract void onPreAccepted(Topologies topologies, Timestamp executeAt, List<PreAcceptOk> oks);
--- a/accord-core/src/main/java/accord/coordinate/CoordinationFailed.java
+++ b/accord-core/src/main/java/accord/coordinate/CoordinationFailed.java
@@ -26,10 +26,11 @@ import accord.primitives.TxnId;
 /**
  * Thrown when a transaction exceeds its specified timeout for obtaining a result for a client
  */
-public class CoordinationFailed extends RuntimeException
+public abstract class CoordinationFailed extends RuntimeException
 {
     private @Nullable TxnId txnId;
     private @Nullable RoutingKey homeKey;
+
     public CoordinationFailed(@Nullable TxnId txnId, @Nullable RoutingKey homeKey)
     {
         this.txnId = txnId;
@@ -39,6 +40,13 @@ public class CoordinationFailed extends RuntimeException
     public CoordinationFailed(@Nullable TxnId txnId, @Nullable RoutingKey homeKey, String message)
     {
         super(message);
+        this.txnId = txnId;
+        this.homeKey = homeKey;
+    }
+
+    protected CoordinationFailed(@Nullable TxnId txnId, @Nullable RoutingKey homeKey, CoordinationFailed cause)
+    {
+        super(cause);
         this.txnId = txnId;
         this.homeKey = homeKey;
     }
@@ -60,5 +68,41 @@ public class CoordinationFailed extends RuntimeException
     public @Nullable RoutingKey homeKey()
     {
         return homeKey;
+    }
+
+    /**
+     * Wrap the exception without changing the type so asynchronous callbacks can add their own stack
+     */
+    public abstract CoordinationFailed wrap();
+
+    public static Throwable wrap(Throwable t)
+    {
+        if (t instanceof CoordinationFailed)
+        {
+            CoordinationFailed wrapped = ((CoordinationFailed)t).wrap();
+            if (wrapped.getClass() != t.getClass())
+            {
+                IllegalStateException ise = new IllegalStateException("Wrapping should not change type");
+                ise.addSuppressed(t);
+                throw ise;
+            }
+            return wrapped;
+        }
+        else if (t instanceof AssertionError)
+        {
+            return new AssertionError(t);
+        }
+        else if (t instanceof OutOfMemoryError)
+        {
+            return t;
+        }
+        else if (t instanceof Error)
+        {
+            return new Error(t);
+        }
+        else
+        {
+            return new RuntimeException(t);
+        }
     }
 }
--- a/accord-core/src/main/java/accord/coordinate/EpochTimeout.java
+++ b/accord-core/src/main/java/accord/coordinate/EpochTimeout.java
@@ -18,33 +18,28 @@
 
 package accord.coordinate;
 
-import javax.annotation.Nullable;
-
-import accord.api.RoutingKey;
-import accord.primitives.TxnId;
-
 import static accord.utils.Invariants.checkState;
 
-/**
- * Thrown when a coordinator is preempted by another recovery
- * coordinator intending to complete the transaction
- */
-public class Preempted extends CoordinationFailed
+public class EpochTimeout extends Timeout
 {
-    public Preempted(TxnId txnId, @Nullable RoutingKey homeKey)
+    public final long epoch;
+
+    public EpochTimeout(long epoch)
     {
-        super(txnId, homeKey);
+        super(null, null);
+        this.epoch = epoch;
     }
 
-    private Preempted(TxnId txnId, @Nullable RoutingKey homeKey, Preempted cause)
+    private EpochTimeout(long epoch, EpochTimeout cause)
     {
-        super(txnId, homeKey, cause);
+        super(null, null, cause);
+        this.epoch = epoch;
     }
 
     @Override
-    public Preempted wrap()
+    public EpochTimeout wrap()
     {
-        checkState(this.getClass() == Preempted.class);
-        return new Preempted(txnId(), homeKey(), this);
+        checkState(this.getClass() == EpochTimeout.class);
+        return new EpochTimeout(epoch, this);
     }
 }
--- a/accord-core/src/main/java/accord/coordinate/Exhausted.java
+++ b/accord-core/src/main/java/accord/coordinate/Exhausted.java
@@ -18,10 +18,12 @@
 
 package accord.coordinate;
 
+import javax.annotation.Nullable;
+
 import accord.api.RoutingKey;
 import accord.primitives.TxnId;
 
-import javax.annotation.Nullable;
+import static accord.utils.Invariants.checkState;
 
 /**
  * Thrown when a transaction exceeds its specified timeout for obtaining a result for a client
@@ -36,5 +38,17 @@ public class Exhausted extends CoordinationFailed
     public Exhausted(TxnId txnId, @Nullable RoutingKey homeKey, String message)
     {
         super(txnId, homeKey, message);
+    }
+
+    protected Exhausted(TxnId txnId, @Nullable RoutingKey homeKey, Exhausted cause)
+    {
+        super(txnId, homeKey, cause);
+    }
+
+    @Override
+    public Exhausted wrap()
+    {
+        checkState(this.getClass() == Exhausted.class);
+        return new Exhausted(txnId(), homeKey(), this);
     }
 }
--- a/accord-core/src/main/java/accord/coordinate/Recover.java
+++ b/accord-core/src/main/java/accord/coordinate/Recover.java
@@ -25,7 +25,6 @@ import java.util.Map;
 import java.util.TreeMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiConsumer;
-import java.util.function.Consumer;
 import java.util.stream.Stream;
 
 import accord.api.Result;
@@ -252,7 +251,7 @@ public class Recover implements Callback<RecoverReply>, BiConsumer<Result, Throw
         RecoverOk acceptOrCommit = maxAccepted(recoverOkList);
         RecoverOk acceptOrCommitNotTruncated = acceptOrCommit == null || acceptOrCommit.status != Status.Truncated
                                                ? acceptOrCommit : maxAcceptedNotTruncated(recoverOkList);
-        
+
         if (acceptOrCommitNotTruncated != null)
         {
             Timestamp executeAt = acceptOrCommitNotTruncated.executeAt;
@@ -269,9 +268,17 @@ public class Recover implements Callback<RecoverReply>, BiConsumer<Result, Throw
                 case Applied:
                 case PreApplied:
                 {
-                    withCommittedDeps(executeAt, stableDeps -> {
+                    withCommittedDeps(executeAt, (stableDeps, withEpochFailure) -> {
+                        if (withEpochFailure != null)
+                        {
+                            node.agent().onUncaughtException(CoordinationFailed.wrap(withEpochFailure));
+                            return;
+                        }
                         // TODO (future development correctness): when writes/result are partially replicated, need to confirm we have quorum of these
-                        persist(adapter, node, tracker.topologies(), route, txnId, txn, executeAt, stableDeps, acceptOrCommitNotTruncated.writes, acceptOrCommitNotTruncated.result, null);
+                        persist(adapter, node, tracker.topologies(), route, txnId, txn, executeAt, stableDeps, acceptOrCommitNotTruncated.writes, acceptOrCommitNotTruncated.result, (ignored, failure) -> {
+                            if (failure != null)
+                                node.agent().onUncaughtException(CoordinationFailed.wrap(withEpochFailure));
+                        });
                     });
                     accept(acceptOrCommitNotTruncated.result, null);
                     return;
@@ -279,7 +286,12 @@ public class Recover implements Callback<RecoverReply>, BiConsumer<Result, Throw
 
                 case Stable:
                 {
-                    withCommittedDeps(executeAt, stableDeps -> {
+                    withCommittedDeps(executeAt, (stableDeps, withEpochFailure) -> {
+                        if (withEpochFailure != null)
+                        {
+                            node.agent().onUncaughtException(CoordinationFailed.wrap(withEpochFailure));
+                            return;
+                        }
                         execute(adapter, node, tracker.topologies(), route, RECOVER, txnId, txn, executeAt, stableDeps, this);
                     });
                     return;
@@ -288,7 +300,12 @@ public class Recover implements Callback<RecoverReply>, BiConsumer<Result, Throw
                 case PreCommitted:
                 case Committed:
                 {
-                    withCommittedDeps(executeAt, committedDeps -> {
+                    withCommittedDeps(executeAt, (committedDeps, withEpochFailure) -> {
+                        if (withEpochFailure != null)
+                        {
+                            node.agent().onUncaughtException(CoordinationFailed.wrap(withEpochFailure));
+                            return;
+                        }
                         stabilise(adapter, node, tracker.topologies(), route, ballot, txnId, txn, executeAt, committedDeps, this);
                     });
                     return;
@@ -362,14 +379,19 @@ public class Recover implements Callback<RecoverReply>, BiConsumer<Result, Throw
         propose(txnId, proposeDeps);
     }
 
-    private void withCommittedDeps(Timestamp executeAt, Consumer<Deps> withDeps)
+    private void withCommittedDeps(Timestamp executeAt, BiConsumer<Deps, Throwable> withDeps)
     {
         LatestDeps.MergedCommitResult merged = LatestDeps.mergeCommit(txnId, executeAt, Arrays.asList(recoverOks), ok -> ok == null ? null : ok.deps);
-        node.withEpoch(executeAt.epoch(), () -> {
+        node.withEpoch(executeAt.epoch(), (ignored, withEpochFailure) -> {
+            if (withEpochFailure != null)
+            {
+                withDeps.accept(null, CoordinationFailed.wrap(withEpochFailure));
+                return;
+            }
             Seekables<?, ?> missing = txn.keys().subtract(merged.sufficientFor);
             if (missing.isEmpty())
             {
-                withDeps.accept(merged.deps);
+                withDeps.accept(merged.deps, null);
             }
             else
             {
@@ -381,7 +403,7 @@ public class Recover implements Callback<RecoverReply>, BiConsumer<Result, Throw
                     }
                     else
                     {
-                        withDeps.accept(merged.deps.with(extraDeps));
+                        withDeps.accept(merged.deps.with(extraDeps), null);
                     }
                 });
             }
@@ -401,14 +423,28 @@ public class Recover implements Callback<RecoverReply>, BiConsumer<Result, Throw
         // If not accepted then the executeAt is not consistent cross the peers and likely different on every node.  There is also an edge case
         // when ranges are removed from the topology, during this case the executeAt won't know the ranges and the invalidate commit will fail.
         Timestamp invalidateUntil = Stream.of(recoverOks).map(ok -> ok == null ? null : ok.status.hasBeen(Status.Accepted) ? ok.executeAt : ok.txnId).reduce(txnId, Timestamp::nonNullOrMax);
-        node.withEpoch(invalidateUntil.epoch(), () -> Commit.Invalidate.commitInvalidate(node, txnId, route, invalidateUntil));
+        node.withEpoch(invalidateUntil.epoch(), (ignored, withEpochFailure) -> {
+            if (withEpochFailure != null)
+            {
+                node.agent().onUncaughtException(CoordinationFailed.wrap(withEpochFailure));
+                return;
+            }
+            Commit.Invalidate.commitInvalidate(node, txnId, route, invalidateUntil);
+        });
         isDone = true;
         locallyInvalidateAndCallback(node, txnId, invalidateUntil, route, ProgressToken.INVALIDATED, callback);
     }
 
     private void propose(Timestamp executeAt, Deps deps)
     {
-        node.withEpoch(executeAt.epoch(), () -> Invoke.propose(adapter, node, route, ballot, txnId, txn, executeAt, deps, this));
+        node.withEpoch(executeAt.epoch(), (ignored, withEpochFailure) -> {
+            if (withEpochFailure != null)
+            {
+                this.accept(null, CoordinationFailed.wrap(withEpochFailure));
+                return;
+            }
+            Invoke.propose(adapter, node, route, ballot, txnId, txn, executeAt, deps, this);
+        });
     }
 
     private void retry()
--- a/accord-core/src/main/java/accord/coordinate/RecoverWithRoute.java
+++ b/accord-core/src/main/java/accord/coordinate/RecoverWithRoute.java
@@ -19,6 +19,7 @@
 package accord.coordinate;
 
 import java.util.function.BiConsumer;
+import javax.annotation.Nullable;
 
 import accord.local.Node;
 import accord.local.Node.Id;
@@ -40,7 +41,6 @@ import accord.primitives.Txn;
 import accord.primitives.TxnId;
 import accord.topology.Topologies;
 import accord.utils.Invariants;
-import javax.annotation.Nullable;
 
 import static accord.coordinate.CoordinationAdapter.Factory.Step.InitiateRecovery;
 import static accord.coordinate.CoordinationAdapter.Invoke.persist;
@@ -211,7 +211,12 @@ public class RecoverWithRoute extends CheckShards<FullRoute<?>>
                 if (known.executeAt.isDecidedAndKnownToExecute() && known.deps.hasDecidedDeps() && known.outcome == Apply)
                 {
                     Deps deps = full.stableDeps.reconstitute(route());
-                    node.withEpoch(full.executeAt.epoch(), () -> {
+                    node.withEpoch(full.executeAt.epoch(), (ignored, withEpochFailure) -> {
+                        if (withEpochFailure != null)
+                        {
+                            node.agent().onUncaughtException(CoordinationFailed.wrap(withEpochFailure));
+                            return;
+                        }
                         persist(node.coordinationAdapter(txnId, InitiateRecovery), node, topologies, route(), txnId, txn, full.executeAt, deps, full.writes, full.result, null);
                     });
                     callback.accept(APPLIED, null);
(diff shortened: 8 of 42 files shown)


accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
```diff
@@ -45,6 +47,8 @@
  */
 abstract class AbstractCoordinatePreAccept<T, R> extends SettableResult<T> implements Callback<R>, BiConsumer<T, Throwable>
 {
+    private static final Logger logger = LoggerFactory.getLogger(AbstractCoordinatePreAccept.class);
```
dcapwell: you don't log anything, so this seems like dead code?

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
```diff
@@ -152,7 +152,12 @@ void onNewEpochTopologyMismatch(TopologyMismatch mismatch)
     void onPreAccepted(Topologies topologies)
     {
         Timestamp executeAt = foldl(oks, (ok, prev) -> mergeMax(ok.witnessedAt, prev), Timestamp.NONE);
-        node.withEpoch(executeAt.epoch(), () -> onPreAccepted(topologies, executeAt, oks));
+        node.withEpoch(executeAt.epoch(), (ignored, withEpochFailure) -> {
```
dcapwell, line 155: makes me wonder if `AsyncChain` is the better solution.  What happens if the success path throws? In this case `onPreAccepted` throws then we don't have any way to catch/trap that exception

accord-core/src/main/java/accord/coordinate/FetchMaxConflict.java
```diff
@@ -45,32 +46,34 @@ public class FetchMaxConflict extends AbstractCoordinatePreAccept<Timestamp, Get
...
         super(node, route, null, topologies);
+        this.store = store;
         this.keysOrRanges = keysOrRanges;
         this.maxConflict = Timestamp.NONE;
         this.executionEpoch = executionEpoch;
         this.tracker = new QuorumTracker(topologies);
     }
 
-    public static AsyncResult<Timestamp> fetchMaxConflict(Node node, Seekables<?, ?> keysOrRanges)
+    public static AsyncResult<Timestamp> fetchMaxConflict(Node node, CommandStore store, Seekables<?, ?> keysOrRanges)
```
dcapwell: why are you passing the `store`?  Shouldn't `CommandStore.current()` be the same thing?

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -262,7 +261,12 @@ private void recover()
                 case Applied:
                 case PreApplied:
                 {
-                    withCommittedDeps(executeAt, stableDeps -> {
+                    withCommittedDeps(executeAt, (stableDeps, failure) -> {
+                        if (failure != null)
+                        {
+                            // TODO (review): This case we don't pass a callback to persist, is that intentional?
```
dcapwell: so its unhandled?  so shouldn't the `agent` be notified then?

accord-core/src/main/java/accord/coordinate/RecoverWithRoute.java
```diff
@@ -211,9 +211,15 @@ else if (!known.definition.isOrWasKnown())
                 if (known.executeAt.isDecidedAndKnownToExecute() && known.deps.hasDecidedDeps() && known.outcome == Apply)
                 {
                     Deps deps = full.stableDeps.reconstitute(route());
-                    node.withEpoch(full.executeAt.epoch(), () -> {
+                    node.withEpoch(full.executeAt.epoch(), (ignored, withEpochFailure) -> {
+                        if (withEpochFailure != null)
+                        {
+                            // TODO (review): We already ignore persist errors here by not providing a callback to persist
```
dcapwell: same as above, shouldn't the `agent` be notified?

accord-core/src/main/java/accord/impl/CoordinateDurabilityScheduling.java
```diff
@@ -214,10 +215,15 @@ private void run()
...
+        node.scheduler().once(() -> node.withEpoch(at.epoch(), (ignored, withEpochFailure) -> {
+                           if (withEpochFailure != null)
+                           {
+                               logger.trace("Exception waiting for epoch before coordinating exclusive sync point for local shard durability, epoch {}: {}", at.epoch(), getStackTraceAsString(withEpochFailure));
+                               return;
+                           }
                            CoordinateSyncPoint.exclusive(node, at, ranges)
                                .addCallback((success, fail) -> {
-                                   if (fail != null) logger.trace("Exception coordinating exclusive sync point for local shard durability of {}", ranges, fail);
+                                   if (fail != null) logger.trace("Exception coordinating exclusive sync point for local shard durability of {}: {}", ranges, getStackTraceAsString(fail));
```
dcapwell: why are you doing `getStackTraceAsString`?  logger does this for you and cheaper (as it can be only when trace is enabled)

accord-core/src/main/java/accord/local/Bootstrap.java
```diff
@@ -124,8 +124,9 @@ void start(SafeCommandStore safeStore0)
 
             if (!node.topology().hasEpoch(globalSyncId.epoch()))
             {
-                node.topology().awaitEpoch(globalSyncId.epoch())
-                    .addCallback(() -> store.execute(empty(), this::start).begin(node.agent())).begin(node.agent());
+                // Ignore timeouts fetching the epoch, always keep trying to bootstrap
+                // TODO (review): begin wasn't being called so this leaked, but what to do with the error?
```
dcapwell: why do you say `begiin` wasn't called?  There are 2 `begin` calls, one for `execute` and one for `addCallback`

accord-core/src/main/java/accord/local/Bootstrap.java
```diff
@@ -232,8 +233,10 @@ private void started(SafeToRead state, @Nullable Timestamp maxApplied)
                         state.startedAt = logicalClock++;
                 }
                 // TODO (expected): associate callbacks with this CommandStore, to remove synchronization
-                FetchMaxConflict.fetchMaxConflict(node, state.ranges)
-                                .begin((executeAt, failure) -> safeToReadCallback(state, executeAt, failure));
+                FetchMaxConflict.fetchMaxConflict(node, store, state.ranges)
+                                .begin((executeAt, failure) -> {
+                                    store.execute(() -> safeToReadCallback(state, executeAt, failure));
```
dcapwell: I think this implies that most of the time we are in the store, but there is an edge case that required you to pass it through?  There is a new method now to run right away when in-store else enqueue; `maybeExecuteImmediately`

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -285,31 +286,24 @@ public void onEpochRedundant(Ranges ranges, long epoch)
...
         if (topology.hasEpoch(epoch))
         {
-            runnable.run();
+            callback.accept(null, null);
         }
         else
         {
             configService.fetchTopologyForEpoch(epoch);
-            topology.awaitEpoch(epoch).addCallback(runnable).begin(agent);
+            topology.awaitEpoch(epoch).addCallback(callback).begin((ignored1, ignored2) -> {});
```
dcapwell: why are you ignoring the error?  If the callback throws we now no longer notify the `agent`?

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -738,7 +725,10 @@ public void receive(Request request, Id from, ReplyContext replyContext)
         if (waitForEpoch > topology.epoch())
         {
             configService.fetchTopologyForEpoch(waitForEpoch);
-            topology().awaitEpoch(waitForEpoch).addCallback(() -> receive(request, from, replyContext));
+            topology().awaitEpoch(waitForEpoch).addCallback((ignored, failure) -> {
```
dcapwell, line 711: TODO handle failure?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -70,7 +76,18 @@
...
+    static
+    {
+        SUCCESS = new FutureEpoch(Long.MAX_VALUE);
+        SUCCESS.future.trySuccess(null);
+    }
+
+    // How long before we start notifying waiters on an epoch of timeout,
+    private static final long EPOCH_INITIAL_TIMEOUT_MILLIS = 10_000;
+    // How often we check for timeout, and once an epoch has timed out, how often we timeout new waiters
+    private static final long WATCHDOG_INTERVAL_MILLIS = 2_000;
```
dcapwell: we have a config class now, we should be pulling from there

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -371,17 +395,86 @@ private int indexOf(long epoch)
...
+         */
+        @GuardedBy("TopologyManager.this")
+        public void timeOutCurrentListeners(long newDeadline, Agent agent)
+        {
+            deadlineMillis = newDeadline;
+            AsyncResult.Settable<Void> oldFuture = future;
+            if (oldFuture.isDone())
+                return;
+            future = AsyncResults.settable();
+            future.addCallback(agent);
```
dcapwell, line 424: this is unsafe (check-then-set pattern), if the `tryFailure` returns `false` it means the future got resolved *after* the `isDone` check, and in this case we would leak and case this new future to produce a timeout even though the epoch is ready.

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -371,17 +395,86 @@ private int indexOf(long epoch)
...
 
+    private static class FutureEpoch
+    {
+        private volatile AsyncResult.Settable<Void> future;
+        private long deadlineMillis;
+
+        public FutureEpoch(long deadlineMillis)
+        {
+            this.future = AsyncResults.settable();
+            this.deadlineMillis = deadlineMillis;
```
dcapwell, line 405: shouldn't we use `nano` rather than `millis`?  `millis` can go backwards but `nano` can't.

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -371,17 +395,86 @@ private int indexOf(long epoch)
...
+        /*
+         * Notify any listeners that are waiting for the epoch that is has been a long time since
+         * we started waiting for the epoch. We may still eventually get the epoch so also create
+         * a new future so subsequent operations may have a chance at seeing the epoch if it ever appears.
+         *
+         * Subsequent waiters may get a timeout notification far sooner (WATCHDOG_INTERVAL_MILLISS)
+         * instead of EPOCH_INITIAL_TIMEOUT_MILLIS
+         */
+        @GuardedBy("TopologyManager.this")
+        public void timeOutCurrentListeners(long newDeadline, Agent agent)
```
dcapwell: feel this should be private as its unsafe to call directly.  If we need to be visible for testing we can make it package-private and denote that with `VisibleForTesting`?

dcapwell (changes requested):

accord-core/src/main/java/accord/impl/CoordinateDurabilityScheduling.java
aweisberg: You can't do string formatting and provide exceptions at the same time. It's the failure case so I could do the string formatting every time. The way this worked before the exception doesn't appear at all.

accord-core/src/main/java/accord/local/Bootstrap.java
aweisberg: Comment is out of date, I fixed this earlier in https://github.com/apache/cassandra-accord/commit/37c957c719491634f081b39900ebf708079ef3ee

accord-core/src/main/java/accord/topology/TopologyManager.java
aweisberg: Is it actually a race? They are guarded by the same lock so they aren't racing.

accord-core/src/main/java/accord/topology/TopologyManager.java
aweisberg: It doesn't need to be public. I'll fix it.

accord-core/src/main/java/accord/local/Node.java
aweisberg: It's redundant because we supply the agent in `TopologyManager` as a listener already. These timeouts produce a lot of not very informative duplicate errors because the exception doesn't actually tell you anything about what witnessed the failure only what failed.

I think I need to do a pass where I make sure the error handlers wrap the exception so they include their stack as well.

accord-core/src/main/java/accord/topology/TopologyManager.java
aweisberg: That makes sense. We don't expose that to Accord right now, but I could add it as something like `elapsedTime` which is elapsed time since some arbitrary point and then use that for things that rely on time. For test code it would be the same time progression as `currentTimeMillis`

Durability scheduling also uses `currentTimeMillis` indirectly so it would also benefit.

accord-core/src/main/java/accord/local/Node.java
aweisberg: How though? The agent already got notified of the timeout. We are going to drop the message, and when we drop messages we don't normally send a response.

The error handling for this dropped message will be somewhere else that relies on the message to be received to make progress and will already have its own timeout.

accord-core/src/main/java/accord/coordinate/RecoverWithRoute.java
```diff
@@ -211,9 +211,15 @@ else if (!known.definition.isOrWasKnown())
...
+                            return;
+                        }
                         persist(node.coordinationAdapter(txnId, InitiateRecovery), node, topologies, route(), txnId, txn, full.executeAt, deps, full.writes, full.result, null);
                     });
+                    // TODO (review): We say it was applied regardless of what happens elsewhere? Could/should this move into the withEpoch?
                     callback.accept(APPLIED, null);
```
aweisberg, line 222: @belliottsmith I want to call this out. It seems like we assume it's applied without checking if it actually worked?

accord-core/src/main/java/accord/impl/CoordinateDurabilityScheduling.java
dcapwell: > You can't do string formatting and provide exceptions at the same Tim

yes you can...

dcapwell:
```
logger.info("What {} {}", "is", "happening", new Throwable());
```

produces 

```
INFO  [main] 2024-07-02 13:52:56,336 EpochGossipTest.java:95 - What is happening
java.lang.Throwable: null
	at org.apache.cassandra.service.accord.EpochGossipTest.main(EpochGossipTest.java:95)
```

accord-core/src/main/java/accord/local/Node.java
dcapwell: my point was that if the callback fails the failure is hidden... by keeping the agent we will detect such cases.  Now, because `callback` doesn't finish the chain you would need to do something else... 

thinking about it, why have `callback` and `begin`... why not just use `begin` directly?

accord-core/src/main/java/accord/topology/TopologyManager.java
dcapwell: looks like its set in `accord.topology.TopologyManager#onTopologyUpdate`, so both share the same lock; so this code is fine

accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
aweisberg: I'll remove it

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
aweisberg: Not sure what you mean here?

accord-core/src/main/java/accord/coordinate/FetchMaxConflict.java
aweisberg: Most of the existing code passes around a `store` instead of using a thread local.

That said... the store is unused? I should be able revert this change entirely.

accord-core/src/main/java/accord/coordinate/Recover.java
aweisberg: I reworked this to have them all pass a wrapped exception to the agent.

accord-core/src/main/java/accord/coordinate/RecoverWithRoute.java
aweisberg: I reworked this to have them all pass a wrapped exception to the agent.

accord-core/src/main/java/accord/impl/CoordinateDurabilityScheduling.java
aweisberg: I'll put it back then!

accord-core/src/main/java/accord/local/Bootstrap.java
aweisberg: Updated to use `maybeExecuteImmediately`

accord-core/src/main/java/accord/local/Node.java
aweisberg: `begin` is documented as not being safe to call multiple times and the result of `awaitEpoch` might not be a new `AsyncChain`, it might be one that is already started. It happens to be that right now `AsyncResult` doesn't actually error out when `begin` is called.

So by contract this is all pretty problematic and in fact `begin` should not even need to be called. I think the problem is partially that `awaitEpoch` should probably return `AsyncResult` not `AsyncChain` so you don't need to call begin you can just add the callback.

The reason it returns `AsyncChain` is because the `withExecutor` in `TopologyManager.awaitEpoch` converts it to an `AsyncChain`.

It might be better to add an `AsyncResults.withExecutor` and use that?

accord-core/src/main/java/accord/topology/TopologyManager.java
aweisberg: Will move it to `LocalConfig`

accord-core/src/main/java/accord/local/Node.java
aweisberg: I updated this to follow the pattern I used elsewhere to wrap the exception and pass it to the agent.

accord-core/src/main/java/accord/config/LocalConfig.java
```diff
@@ -22,8 +22,22 @@
...
 {
+    LocalConfig DEFAULT = new LocalConfig() {};
+
     default Duration getProgressLogScheduleDelay()
     {
         return Duration.ofSeconds(1);
     }
+
+    // How long before we start notifying waiters on an epoch of timeout,
+    default long epochFetchInitialTimeout()
```
dcapwell: what is the unit?  `Duration` is best as thats value + unit

accord-core/src/main/java/accord/config/LocalConfig.java
```diff
@@ -22,8 +22,22 @@
...
+    {
+        return 10_000;
+    }
+
+    // How often to check for timeout, and once an epoch has timed out, how often we timeout new waiters
+    default long epochFetchWatchdogIntervalMillis()
```
dcapwell: can you use `Duration`?  Forcing `mills` is something we try to avoid in C* config now

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
dcapwell: The issue with callbacks is that failures bubble up in random places, where as `AsyncChain` kinda forces `begin` to handle everything.  Right now if the epoch is known this call trace must handle the exceptions, and if not then we send to the `agent` (as it becomes a `AsyncChain` under the hood).

accord-core/src/main/java/accord/coordinate/CoordinationFailed.java
```diff
@@ -61,4 +70,21 @@ void set(TxnId txnId, RoutingKey homeKey)
...
+     */
+    public CoordinationFailed wrap()
+    {
+        checkState(this.getClass() == CoordinationFailed.class);
+        return new CoordinationFailed(txnId, homeKey, this);
+    }
+
+    public static Throwable wrap(Throwable t)
+    {
+        if (t instanceof CoordinationFailed)
```
dcapwell, line 85: not safe... we almost never use `CoordinationFailed` directly, we use an extending type... so the type check will match many things, then `wrap()` will fail as we *require* the top level class to be `CoordinationFailed` (which is almost never true)

accord-core/src/main/java/accord/coordinate/CoordinationFailed.java
```diff
@@ -61,4 +70,21 @@ void set(TxnId txnId, RoutingKey homeKey)
...
+            return ((CoordinationFailed)t).wrap();
+        else
+            throw new RuntimeException(t);
```
dcapwell: not a fan of this case... `CoordinationFailed` has nothing to do in this case... Why not do something like `EpochException` where the `cause` is the reason?  Most of the callers are just wrapping the epoch fail cause... so why not make that type more clear?

accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
dcapwell: another advantage is that you can focus on the happy-path when building the chain and only focus on the failure in one place.  With callbacks we keep having to focus on both cases in every single block...

dcapwell: Gave this example on slack...

`accord-core/src/main/java/accord/coordinate/Recover.java`

```
node.withEpoch(invalidateUntil.epoch(), (ignored, withEpochFailure) -> {
            // TODO (review): There was already nothing here to handle failures in commitInvalidate
            if (withEpochFailure != null)
            {
                node.agent().onUncaughtException(CoordinationFailed.wrap(withEpochFailure));
                return;
            }
            Commit.Invalidate.commitInvalidate(node, txnId, route, invalidateUntil);
        });
```

vs

```
node.withEpoch(invalidateUntil.epoch())
    .accept(ignore -> Commit.Invalidate.commitInvalidate(node, txnId, route, invalidateUntil))
    .begin(agent);
```

For this project, when given the success and failure we keep ignoring the errors, and this patch is kinda trying to address that by handling the errors... but we keep having this problem over and over again; callbacks have been very error prone for us...

By using `AsyncChain` we can focus on the success cases only and can send the unexpected / unhandled errors to the `agent`

accord-core/src/main/java/accord/coordinate/Exhausted.java
```diff
@@ -37,4 +39,16 @@ public Exhausted(TxnId txnId, @Nullable RoutingKey homeKey, String message)
...
         super(txnId, homeKey, message);
     }
+
+    protected Exhausted(TxnId txnId, @Nullable RoutingKey homeKey, Exhausted cause)
+    {
+        super(txnId, homeKey, cause);
+    }
+
+    @Override
+    public Exhausted wrap()
```
dcapwell, line 49: not a fan of this pattern as its hard to maintain... you *must* extend, but the base type also defines, so adding a new type will fail eventually when `wrap` is called... I don't personally see the benefit of keeping the type as it just makes things more confusing to me... `ExecutionException` is a good type as its clear the `cause` is what you want, but also keeps the stack trace of the call... but when we have the main error as the top level type, it can be confusing what happened

accord-core/src/main/java/accord/coordinate/CoordinationFailed.java
aweisberg: This isn't specifically for `withEpoch` it is for all the cases where the callback has nothing to add other then the failure that happened impacted this particular callback here.

The thing that sent me down his road was the pain point of burying the actual exception type in the causal chain. I am trying to preserve actual root cause type, so as not to break a lot of the existing code that just wants to catch the root cause type.

If it's not already an Accord exception then there isn't much to add beyond, yes this occurred at runtime. I think the biggest improvement here would be to preserve a few more things like `Error`, `OutOfMemoryError`, and `AssertionError`

accord-core/src/main/java/accord/local/NodeTimeService.java
```diff
@@ -36,18 +37,53 @@ public interface NodeTimeService
     long now();
 
     /**
-     * Return the current time since the Unix epoch in the specified time unit. May still be simulated time and not
-     * real time.
+     * Return the current time since some arbitrary epoch in the specified time unit. May still be simulated time and not
+     * real time. The time returned by this will be monotonic.
      */
-    long unix(TimeUnit unit);
+    long elapsed(TimeUnit unit);
```
dcapwell, line 43: I don't think this change is good for the project.  We do require `unix` and not `elapsed` in most places, and `elapsed` is relative to the local node only and not cross the cluster.  This logic is used as part of our `TxnId` so switching to `elapsed` *should* break us in subtle ways.

accord-core/src/main/java/accord/local/NodeTimeService.java
```diff
@@ -36,18 +37,53 @@ public interface NodeTimeService
...
+    {
+        return resultUnit ->  resultUnit.convert(monotonicNowSupplier.getAsLong(), sourceUnit);
+    }
+
+    /**
+     * Allow time progression to be controlled using a potentially non-monotonic time source for testing with simulated
+     * time sources. This is not designed to be fast and real usage should be with a monotonic time source.
+     */
+    @VisibleForTesting
+    static ToLongFunction<TimeUnit> elapsedWrapperFromNonMonotonicSource(TimeUnit sourceUnit, LongSupplier nonMonotonicNowSupplier)
```
dcapwell, line 57: is this really needed?  In the callers found they are actually monotonic... I assume C* is going to use `nanoTime`, so we don't have a `non-monotonic` case; and if we do thats actually a bug in the test code...

accord-core/src/main/java/accord/local/NodeTimeService.java
```diff
@@ -36,18 +37,53 @@ public interface NodeTimeService
...
+       // Use an arbitrary epoch
+       private long epoch = Long.MAX_VALUE / 4;
+
+       private MonotonicWrapper(LongSupplier nowSupplier)
+       {
+           this.nowSupplier = nowSupplier;
+       }
+
+       @Override
+       public synchronized long getAsLong()
```
dcapwell, line 77: this is not monotinc

```
call 1:
lastNow = MIN_VALUE
now = 100
delta = 0
return (100 + 0) = 100

call 2:
lastNow = 100
now = 99
delta = 1

lastNow = 99
return (99 + 1) = 100
```

If you want something like this its far simpler just to fallback to a counter when `now` rolls backwards

```
private long lastNow = Long.MIN_VALUE;

long getAsLong() {
  long now = nowSupplier.getAsLong();
  if (now < lastNow)
    now = lastNow + 1;
  lastNow = now;
  return now;
}
```

accord-core/src/main/java/accord/topology/TopologyManager.java
dcapwell: > Durability scheduling 

as far as I can tell that is the only caller of `unixTime(TimeUnit)` and it doesn't need `currentTimeMillis` and is better as `nanoTime` (as you are pointing out)... so moving `unixTime(TimeUnit)` to `elapsedTime` makes sense to me...

accord-core/src/main/java/accord/coordinate/Exhausted.java
aweisberg: I can have this generate an illegal state exception if you try to wrap and it changes the type. Alternatively I can restructure hierarchy so there is no default implementation and then it's statically enforced.

The problem with changing the type is that then you have catch blocks that don't work because they don't go through the entire causal chain. That is what originally sent me down this rabbit hole.

dcapwell: if you require each type to override `wrap` then `CoordinationFailed` should be `abstract` and this method *must* be `abstract`.  All these checks make maintenance harder

dcapwell: I just checked, no one returns `CoordinationFailed`, they all use the higher up types... so making this class abstract should be fine

accord-core/src/main/java/accord/local/NodeTimeService.java
aweisberg: If we are concerned about non-monotonic time won't that be something that test code such as `BurnTest` exercises? I wanted the passage of monotonic time to match non-monotonic time, but just be monotonic and based off of a different epoch similar to how `nanoTime` works.

We have code in `Node.uniqueNow` to ensure that it remains monotonic even when the time that is provided so we should exercise it if we aren't already.

accord-core/src/main/java/accord/local/NodeTimeService.java
aweisberg: It's normal for something that reports time to be able to report the same value multiple times and it also fits the definition for monotonic which is either non-increasing or non-decreasing. In this case the function is non-decreasing.

This isn't a transaction id generation function trying to create new unique values where counting makes sense to keep generating unique values when time doesn't move forward.

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -371,15 +391,87 @@ private int indexOf(long epoch)
...
         this.epochs = Epochs.EMPTY;
+        this.localConfig = localConfig;
+    }
+
+    public void shutdown()
+    {
+        topologyUpdateWatchdog.cancel();
+    }
+
+    public void scheduleTopologyUpdateWatchdog()
```
dcapwell, line 455: thinking about this more, here is an annoying edge case

```
awaitEpoch(1) -> timeout is 10s
...

// 10 seconds later, but before `scheduleTopologyUpdateWatchdog` is called
awaitEpoch(1) -> this is going to see a timeout right away rather than 10 seconds later...
```

since timeouts are for the start of the batch, anything that joined *after* will timeout much sooner than the timeout window...

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -371,15 +391,87 @@ private int indexOf(long epoch)
...
+        @GuardedBy("TopologyManager.this")
+        private void timeOutCurrentListeners(long newDeadline, Agent agent)
+        {
+            deadlineMillis = newDeadline;
+            AsyncResult.Settable<Void> oldFuture = future;
+            if (oldFuture.isDone())
+                return;
+            future = AsyncResults.settable();
+            future.addCallback(agent);
+            oldFuture.tryFailure(new Timeout(null, null));
```
dcapwell: Can we use a new type?  `Timeout` implies we waited the allowed amount of time for a txn, but this timeout is different... There is an edge case (documented above) where a txn sees this `Timeout` after < 1s of starting... which would be super confusing to figure out

accord-core/src/main/java/accord/local/Bootstrap.java
```diff
@@ -124,8 +125,11 @@ void start(SafeCommandStore safeStore0)
 
             if (!node.topology().hasEpoch(globalSyncId.epoch()))
             {
-                node.topology().awaitEpoch(globalSyncId.epoch())
-                    .addCallback(() -> store.execute(empty(), this::start).begin(node.agent())).begin(node.agent());
+                // Ignore timeouts fetching the epoch, always keep trying to bootstrap
+                node.withEpoch(globalSyncId.epoch(), (ignored, failure) -> store.execute(empty(), Attempt.this::start).begin((ignored1, failure2) -> {
```
dcapwell, line 129: if the epoch timed out why are we trying to start the attempt?

accord-core/src/main/java/accord/config/LocalConfig.java
```diff
@@ -19,11 +19,26 @@
...
+
     default Duration getProgressLogScheduleDelay()
     {
         return Duration.ofSeconds(1);
     }
+
+    // How long before we start notifying waiters on an epoch of timeout,
+    default Duration epochFetchInitialTimeout()
+    {
+        return Duration.of(10, ChronoUnit.SECONDS);
```
dcapwell: nit: 

```suggestion
        return Duration.ofSeconds(10);
```

accord-core/src/main/java/accord/config/LocalConfig.java
```diff
@@ -19,11 +19,26 @@
...
+    }
+
+    // How often to check for timeout, and once an epoch has timed out, how often we timeout new waiters
+    default Duration epochFetchWatchdogInterval()
+    {
+        return Duration.of(10, ChronoUnit.SECONDS);
```
dcapwell: nit:

```suggestion
        return Duration.ofSeconds(10);
```

accord-core/src/main/java/accord/coordinate/Exhausted.java
```diff
@@ -37,4 +39,16 @@ public Exhausted(TxnId txnId, @Nullable RoutingKey homeKey, String message)
...
+    {
+        checkState(this.getClass() == Exhausted.class);
```
dcapwell, line 51: you could replace this with `final` in `wrap`

dcapwell (commented):
forgot to post, sorry

accord-core/src/main/java/accord/coordinate/Exhausted.java
aweisberg: How? There is a derived class `RangeUnavailable` that appears to be unused that could fail to implement `wrap()`?

Making the method `final` wouldn't stop the creation of derived classes although I could make the classes `final`. I don't really have a preference I just didn't want to dictate whether the classes could be extended since we already had a case of extending them.

dcapwell: spoke in slack, `RangeUnavailable` is dead code so could remove, but I am cool with leaving as is.  This isn't a blocking comment

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -262,17 +261,30 @@ private void recover()
                 case Applied:
                 case PreApplied:
                 {
-                    withCommittedDeps(executeAt, stableDeps -> {
+                    withCommittedDeps(executeAt, (stableDeps, withEpochFailure) -> {
+                        if (withEpochFailure != null)
+                        {
+                            node.agent().onUncaughtException(new RuntimeException(withEpochFailure));
```
dcapwell: can't we try to wrap here for the cases that are coordination errors?

accord-core/src/main/java/accord/topology/TopologyManager.java
dcapwell: spoke with Ariel and this likely isn't a big deal.... when a local node sees an epoch from its peer it won't have "received" it until the following happens.

1) TCM notifies the node
2) the previous epoch was `epochs.receive(topology);`
3) the previous epoch has constructed the `Bootstrap` class (or `EpochReady.DONE`).

Once all 3 are done, then the local node will receive the topology.  This most likely causes us to mostly wait on TCM and not the other steps as they are "fast" (assuming nothing hangs forever)


given this, a batch timeout should be fine... we can always change later on if its an issue

accord-core/src/main/java/accord/coordinate/EpochTimeout.java
```diff
@@ -0,0 +1,39 @@
...
+    {
+        super(null, null);
+        this.epoch = epoch;
+    }
+
+    @Override
+    public EpochTimeout wrap()
+    {
+        checkState(this.getClass() == EpochTimeout.class);
+        return new EpochTimeout(epoch);
```
dcapwell: don't you need to preserve the old stack trace too?

dcapwell (approved):
`EpochTimeout.wrap` is not correct, but approving as that can be fixed during the merge

accord-core/src/main/java/accord/coordinate/EpochTimeout.java
aweisberg: Yes, missed that.

PR #102, opened 2024-07-08
Journal with command store

ifesdjeen:
Accord counterpart of https://github.com/apache/cassandra/pull/3409

### DIFF
--- a/accord-core/src/main/java/accord/local/Command.java
+++ b/accord-core/src/main/java/accord/local/Command.java
@@ -645,9 +645,15 @@ public abstract class Command implements CommonAttributes
         SaveStatus saveStatus = saveStatus();
         return saveStatus.hasBeen(Status.Committed) && !saveStatus.hasBeen(Invalidated);
     }
+
     public final boolean isStable()
     {
         SaveStatus saveStatus = saveStatus();
+        return isStable(saveStatus);
+    }
+
+    public static boolean isStable(SaveStatus saveStatus)
+    {
         return saveStatus.hasBeen(Status.Stable) && !saveStatus.hasBeen(Invalidated);
     }
 
@@ -664,6 +670,11 @@ public abstract class Command implements CommonAttributes
     public final boolean isTruncated()
     {
         return status().hasBeen(Status.Truncated);
+    }
+
+    public static boolean isTruncated(Status status)
+    {
+        return status.hasBeen(Status.Truncated);
     }
 
     public abstract Command updateAttributes(CommonAttributes attrs, Ballot promised);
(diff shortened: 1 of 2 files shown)


accord-core/src/main/java/accord/local/Command.java
```diff
@@ -666,6 +672,11 @@ public final boolean isTruncated()
         return status().hasBeen(Status.Truncated);
     }
 
+    public static boolean isTruncated(Status status)
+    {
+        return status.hasBeen(Status.Truncated);
```
dcapwell, line 677:
```
Truncated         (Cleanup, Maybe, DefinitionErased,  ExecuteAtErased,  DepsErased,   Outcome.Erased),
Invalidated       (Persist, Maybe, NoOp,              NoExecuteAt,      NoDeps,       Outcome.Invalidated),
```

wouldn't that include `Invalidated`?

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -1240,7 +1251,7 @@ public boolean equals(Object o)
             if (!super.equals(o)) return false;
             Executed executed = (Executed) o;
             return Objects.equals(writes, executed.writes)
-                && Objects.equals(result, executed.result);
+                   && Objects.equals(result, executed.result); // TODO: find a different way to check/assert?
```
dcapwell: why this todo?  what is the issue with the check?

dcapwell: `TODO` should be `//TODO (classification 1, classification 2, etc.): msg`

This is so we can find them easier and find the *why* and *when* to fix issues.

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -558,6 +558,9 @@ protected void start(BiConsumer<? super R, Throwable> callback)
 
     public void reply(Id replyingToNode, ReplyContext replyContext, Reply send, Throwable failure)
     {
+        if (replyingToNode == Id.NONE)
```
dcapwell: when is this case in the protocol?

accord-core/src/main/java/accord/local/Command.java
belliottsmith: the main classifications are priority ones, and should go _first_:

desired
expected
required

accord-core/src/main/java/accord/local/Node.java
ifesdjeen: This _used_ to be the case during replay. But you are right, maybe with my latest changes this will not be the case anymore.

accord-core/src/main/java/accord/local/Command.java
ifesdjeen: So I have simply copied the code from above; my intention was to make non-static version use the static one as I did with `isStable`. You are right `isTruncated` is included here.

accord-core/src/main/java/accord/local/Command.java
ifesdjeen: Reverted this; this was a tmp TODO.

accord-core/src/main/java/accord/local/Node.java
ifesdjeen: Removed now.

PR #103, opened 2024-07-12
CASSANDRA-19769: CEP-15: (Accord) sequence EpochReady.coordinating to allow syncComplete to be learned from newer epochs

dcapwell:


accord-core/src/main/java/accord/coordinate/tracking/QuorumTracker.java
```diff
@@ -66,6 +66,15 @@ boolean hasFailed()
         {
             return failures > shard.maxFailures;
         }
+
+        @Override
+        public String toString()
```
dcapwell, line 71: not due to logging but more to debugging... we will have 30 to 60 shards and finding which ones to look at in a debugger is annoying without logging

accord-core/src/main/java/accord/local/Bootstrap.java
```diff
@@ -140,6 +140,8 @@ void start(SafeCommandStore safeStore0)
                .flatMap(syncPoint -> node.withEpoch(epoch, () -> store.submit(contextFor(localSyncId, syncPoint.waitFor.keyDeps.keys(), KeyHistory.COMMANDS), safeStore1 -> {
                    if (valid.isEmpty()) // we've lost ownership of the range
                        return AsyncResults.success(Ranges.EMPTY);
+                   if (!syncPoint.waitFor.isEmpty() && !node.topology().hasEpoch(syncPoint.waitFor.minTxnId().epoch()))
```
dcapwell: this would happen when new nodes join the ring in C* as they would load the epochs from w/e existed when the JVM started, but not be based off what's in-flight.  Since C* now discovers what epochs to care about, this assertion makes sure we never try to add a txn with an unknown epoch

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -128,6 +128,7 @@ public boolean equals(Object that)
 
         public boolean equals(Id that)
         {
+            if (that == null) return false;
```
dcapwell, line 131: I forgot what hit this but likely was a test and not real code.  Equals is allowed to see `null` and is common to have `foo.equals(null)`

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -241,7 +242,21 @@ public long epoch()
     private synchronized EpochReady onTopologyUpdateInternal(Topology topology, boolean startSync)
     {
         Supplier<EpochReady> bootstrap = commandStores.updateTopology(this, topology, startSync);
-        return this.topology.onTopologyUpdate(topology, bootstrap);
+        Supplier<EpochReady> ordering = () -> {
```
dcapwell, line 249: this is one of the two core changes in this PR.  By ordering `EpochReady.coordination`, we have the property that epoch=5 being ready implies all epochs before are also, this property is required for C* to recover the known epochs learned *before* joining the ring.

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -241,7 +242,21 @@ public long epoch()
...
+            if (this.topology.isEmpty()) return bootstrap.get();
+            return order(this.topology.epochReady(topology.epoch() - 1), bootstrap.get());
+        };
+        return this.topology.onTopologyUpdate(topology, ordering);
+    }
+
+    private static EpochReady order(EpochReady previous, EpochReady next)
+    {
+        if (previous.coordination.isDone()) return next;
```
dcapwell, line 260: if we did `flatMap` 100% of the time then everything is linked together, so this allows the more common case (epochs don't all happen at once) to avoid linking

accord-core/src/main/java/accord/primitives/TxnId.java
```diff
@@ -166,7 +166,7 @@ public static TxnId minForEpoch(long epoch)
         return new TxnId(epochMsb(epoch), 0, Id.NONE);
     }
 
-    private static final Pattern PARSE = Pattern.compile("\\[(?<epoch>[0-9]+),(?<hlc>[0-9]+),(?<flags>[0-9]+)\\([KREWNSXL]\\),(?<node>[0-9]+)]");
+    private static final Pattern PARSE = Pattern.compile("\\[(?<epoch>[0-9]+),(?<hlc>[0-9]+),(?<flags>[0-9]+)\\([KREWNSXL]+\\),(?<node>-?[0-9]+)]");
```
dcapwell: bug found in testing.  The property `this == parse(this.toString())` was not valid, and we actually could not parse anything at all!  There is a new property test to make sure this is working and doesn't break again

accord-core/src/main/java/accord/topology/Shard.java
```diff
@@ -41,27 +41,34 @@ public class Shard
     public final List<Id> nodes;
     public final ExtendedSortedArrayList<Id> sortedNodes;
     public final Set<Id> fastPathElectorate;
-    public final Set<Id> joining;
+    public final Set<Id> joining, leaving;
```
dcapwell: in accord we don't leverage this, though it would be nice... we have cases where we ask nodes that are leaving for information, and then rely on timeouts to select from another peer.  I didn't want to change this behavior in this patch, but `leaving` is important during C* boot up so it knows which nodes to mark removed

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -98,36 +97,33 @@ static class EpochState
...
-                return false;
-            Invariants.checkState(newPrevSyncComplete.containsAll(prevSyncComplete), "Expected %s to contain all ranges in %s; but did not", newPrevSyncComplete, prevSyncComplete);
-            prevSyncComplete = newPrevSyncComplete;
-            syncComplete = curSyncComplete.slice(newPrevSyncComplete, Minimal).union(MERGE_ADJACENT, addedRanges);
-            return true;
+            return syncTracker == null ? true : syncTracker.hasReachedQuorum();
         }
 
-        public boolean hasReachedQuorum()
+        private boolean recordSyncCompleteFromFuture()
```
dcapwell, line 122: I actually want to rework this so its node based and not complete based... for testing it was easier to do complete based, but in practice node base will be best as the property is relative to the node, so node sending for epoch=N basically just sent for <= N as well

accord-core/src/main/java/accord/utils/RandomSource.java
```diff
@@ -350,7 +350,7 @@ default long reset()
 
     default Random asJdkRandom()
     {
-        return new Random()
+        return new Random(nextLong())
```
dcapwell: this was a bug that impacted deterministic execution.  When we fork to a java `Random` we didn't control the seed so could have different histories cross runs with the same seed!

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -551,40 +549,22 @@ public void testOne()
...
-                promise.setFailure(e);
-            }
-        });
-        t.setName("BurnTest with timeout");
-        t.setDaemon(true);
         try
         {
-            t.start();
-            AsyncChains.getBlocking(promise, timeout.toNanos(), TimeUnit.NANOSECONDS);
+            TimeoutUtils.runBlocking(timeout, "BurnTest with timeout", () -> run(seed, 1000));
```
dcapwell, line 584: removing duplicate logic...  This logic was forked from property testing, and now that its a utility I just leverage that rather than fork the logic still.

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -551,40 +549,22 @@ public void testOne()
...
-            if (cause instanceof InterruptedException || cause instanceof TimeoutException)
-                t.interrupt();
             if (cause instanceof TimeoutException)
             {
                 TimeoutException override = new TimeoutException("test did not complete within " + timeout);
                 override.setStackTrace(new StackTraceElement[0]);
                 cause = override;
             }
-            logger.error("Exception running burn test for seed {}:", seed, t);
+            logger.error("Exception running burn test for seed {}:", seed, cause);
```
dcapwell, line 597: `t` was a bug, ment to do `cause` or `thrown` to see what actually caused the failure

accord-core/src/test/java/accord/utils/Gens.java
```diff
@@ -117,6 +119,8 @@ public static <T> Gen<T> pick(Map<T, Integer> values)
         // this method relies on the map having some order and will reject any map that doesn't define a deterministic order
         if (!(values instanceof EnumMap || values instanceof LinkedHashMap))
             throw new IllegalArgumentException("pick(Map) requires a map with deterministic iteration; given " + values.getClass());
+        if (values.size() == 1)
```
dcapwell: not required but saw it was a common case, so good to optimize

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -29,17 +27,32 @@
...
+            long current = seedUniquifier.get();
+            long next = current * 1181783497276652981L;
+            if (seedUniquifier.compareAndSet(current, next))
+                return next;
+        }
+    }
+
     public static abstract class Common<T extends Common<T>>
     {
-        protected long seed = ThreadLocalRandom.current().nextLong();
```
dcapwell: this is a bug... This tends to produce the same seed over and over when you restart the JVM... this would mean we would run with the same seed every time!  Rather than do that, the new logic uses `java.util.Random`'s seed logic

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -415,6 +413,8 @@ public static StatefulBuilder stateful()
     public static class StatefulBuilder extends Common<StatefulBuilder>
     {
         protected int steps = 1000;
+        @Nullable
+        protected Duration stepTimeout = null;
```
dcapwell: `TopologyMixupTest` is what. uses this.  The test runs long running tasks that could block forever, so having a timeout per step will cause us to fail instead and recover the seed.

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -454,16 +460,35 @@ public <State, SystemUnderTest> void check(Commands<State, SystemUnderTest> comm
                                     throw new IllegalArgumentException("Unable to find next command");
                                 cmd = cmdGen.next(rs);
                             }
-                            history.add(cmd.detailed(state));
-                            Object stateResult = cmd.apply(state);
-                            cmd.checkPostconditions(state, stateResult,
-                                                    sut, cmd.run(sut));
+                            if (cmd instanceof MultistepCommand)
```
dcapwell: Added for `TopologyMixupTest`.  Some times it's nice to produce a command that actually is multiple commands that run back to back.  This is mostly done to have a more detailed history more than anything else.

I am not a fan of the interface and finds that the sub command `details` isn't as nice as we compute them *before* running the first command, it would be nice to defer to right before executing the command...

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -510,12 +604,42 @@ default Void run(SystemUnderTest sut) throws Throwable
         }
     }
 
+    public static class SimpleCommand<State> implements UnitCommand<State, Void>
```
dcapwell: a few tests don't have an explicit `SystemUnderTest`, so this tends to shorten things

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -510,12 +604,42 @@ default Void run(SystemUnderTest sut) throws Throwable
...
+        }
+    }
+
     public interface Commands<State, SystemUnderTest>
     {
         Gen<State> genInitialState() throws Throwable;
         SystemUnderTest createSut(State state) throws Throwable;
-        default void destroyState(State state) throws Throwable {}
-        default void destroySut(SystemUnderTest sut) throws Throwable {}
+        default void destroyState(State state, @Nullable Throwable cause) throws Throwable {}
```
dcapwell: There are cases that need to know if the test failed or not to figure out how to destroy; one such test is `EpochSyncTest` which finishes running all scheduled tasks in `destroyState` but has no reason to if the test already failed.

accord-core/src/main/java/accord/local/Bootstrap.java
```diff
@@ -140,6 +140,8 @@ void start(SafeCommandStore safeStore0)
...
+                       throw Invariants.illegalState("Deps has a txn %s that contains an unknown epoch; min=%d, max=%d", syncPoint.waitFor.minTxnId(), node.topology().minEpoch(), node.topology().current().epoch());
```
ifesdjeen: Should we throw here or should we return `AyncResults.failure`?

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -241,7 +242,21 @@ public long epoch()
...
         Supplier<EpochReady> bootstrap = commandStores.updateTopology(this, topology, startSync);
-        return this.topology.onTopologyUpdate(topology, bootstrap);
+        Supplier<EpochReady> ordering = () -> {
+            if (this.topology.isEmpty()) return bootstrap.get();
+            return order(this.topology.epochReady(topology.epoch() - 1), bootstrap.get());
+        };
+        return this.topology.onTopologyUpdate(topology, ordering);
+    }
+
+    private static EpochReady order(EpochReady previous, EpochReady next)
```
ifesdjeen, line 256: Should we add an assert that `prev` and `next` are immediately subsequent? ie epochs diff is 1

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -98,36 +97,33 @@ static class EpochState
 
             this.addedRanges = global.ranges.subtract(prevRanges).mergeTouching();
             this.removedRanges = prevRanges.mergeTouching().subtract(global.ranges);
-            this.prevSyncComplete = addedRanges.union(MERGE_ADJACENT, prevSyncComplete.subtract(removedRanges));
-            this.curSyncComplete = this.syncComplete = addedRanges;
+            this.syncComplete = addedRanges;
```
ifesdjeen, line 114: Could you explain why we do not need to use reemoved ranges anymore?..

accord-core/src/main/java/accord/local/Node.java
dcapwell: done

accord-core/src/main/java/accord/topology/TopologyManager.java
dcapwell: because `prevSyncComplete` concept no longer exists.  

The logic is used when we launch transactions to see what older epochs we don't know the state of that intersect the transaction, that way we include them.  The target reason is token movements... we need to include the node that had the range removed until we hear from it that it knows the new epoch.

So, when a range is added, we know we don't need to track it as no older epochs have it as well
When a range is removed, it is no longer present... so we won't add to syncComplete...

does this make sense?

accord-core/src/main/java/accord/local/Bootstrap.java
dcapwell: we could... the end result will be the same.  I prefer throws but its not a strong preference... in this lambda `Commands.createBootstrapCompleteMarkerTransaction` and `safeStore1.registerHistoricalTransactions` both have similar checks as well

accord-core/src/main/java/accord/coordinate/Exhausted.java
```diff
@@ -32,27 +32,27 @@
...
-
     Exhausted(TxnId txnId, @Nullable RoutingKey homeKey, Ranges unavailable, Exhausted cause)
     {
         super(txnId, homeKey, cause);
         this.unavailable = unavailable;
     }
 
+    private static String getMessage(TxnId txnId, @Nullable Ranges unavailable)
+    {
+        String msg = "No more nodes to try for " + txnId;
```
dcapwell, line 50: this isn't a core change, but when debugging you need to know what txn so you can inspect the state.  "something failed" is hard to work with and was hard to debug while creating this patch

accord-core/src/main/java/accord/primitives/Range.java
```diff
@@ -216,12 +216,12 @@ private Range(RoutingKey start, RoutingKey end)
         this.end = end;
     }
 
-    public final RoutingKey start()
+    public RoutingKey start()
```
dcapwell, line 219: in `TokenRange` we kept doing casts making the code ugly... this change was to try to make that logic cleaner... its 100% style change

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -365,6 +365,11 @@ static <T extends Timestamp> T merge(Timestamp a, Timestamp b, Constructor<T> co
 
     @Override
     public String toString()
+    {
+        return toStandardString();
+    }
+
+    public String toStandardString()
```
dcapwell, line 372: so, when we serializer in some cases we only know `Timestamp` but we have many types... and almost all override `toString` so the parser of string to `Timestamp` breaks...  By making this method it makes sure we can always do the following property

```
new Timestamp(this).equals(Timestamp.parse(this.toStandardString()))
```

dcapwell: this is not a core change, it was just found in testing....

accord-core/src/main/java/accord/primitives/TxnId.java
```diff
@@ -181,7 +181,7 @@ public static TxnId minForEpoch(long epoch)
         return new TxnId(epochMsb(epoch), 0, Id.NONE);
     }
 
-    private static final Pattern PARSE = Pattern.compile("\\[(?<epoch>[0-9]+),(?<hlc>[0-9]+),(?<flags>[0-9]+)\\([KR][REWSXL]\\),(?<node>[0-9]+)]");
+    private static final Pattern PARSE = Pattern.compile("\\[(?<epoch>[0-9]+),(?<hlc>[0-9]+),(?<flags>[0-9]+)\\([KR][REWSXL]\\),(?<node>-?[0-9]+)]");
```
dcapwell: TCM allows negative nodes (and epoch)... so extended the pattern to better support TCM (though this doesn't support negative epochs... those are for transitioning from Gossip)

accord-core/src/main/java/accord/topology/Shard.java
```diff
@@ -116,6 +116,8 @@ public String toString(boolean extendedInfo)
                     sb.append('f');
             }
             sb.append(')');
+            if (!joining.isEmpty())
```
dcapwell, line 119: unrelated but its been useful to see how we handle TCM transitions... during host replacements we go into "joining" but logically we are "leaving" =D

accord-core/src/main/java/accord/utils/Invariants.java
```diff
@@ -38,9 +38,12 @@ public enum ParanoiaCostFactor
         LOW, HIGH
     }
 
-    private static final int PARANOIA_COMPUTE = Paranoia.valueOf(System.getProperty("accord.paranoia.cpu", "NONE").toUpperCase()).ordinal();
-    private static final int PARANOIA_MEMORY = Paranoia.valueOf(System.getProperty("accord.paranoia.memory", "NONE").toUpperCase()).ordinal();
-    private static final int PARANOIA_FACTOR = ParanoiaCostFactor.valueOf(System.getProperty("accord.paranoia.costfactor", "LOW").toUpperCase()).ordinal();
+    public static final String KEY_PARANOIA_CPU = "accord.paranoia.cpu";
```
dcapwell, line 41: in C* I put these into `CassandraRelativeProperties` so check styles doesn't yell at us

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -581,40 +579,22 @@ public void testOne()
     private static void run(long seed)
     {
         Duration timeout = Duration.ofMinutes(3);
-        Runnable fn = () -> run(seed, 1000);
-        AsyncResult.Settable<?> promise = AsyncResults.settable();
-        Thread t = new Thread(() -> {
```
dcapwell, line 586: this timeout logic is now in `TimeoutUtils` as its copy/pasted in 2 places and a 3rd needed it...

accord-core/src/test/java/accord/primitives/TxnIdTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+class TxnIdTest
+{
+    @Test
+    void stringSerde()
+    {
+        qt().forAll(AccordGens.txnIds()).check(id -> {
+            Assertions.assertThat(TxnId.parse(id.toString())).isEqualTo(id);
+
+            Assertions.assertThat(Timestamp.fromString(id.toStandardString())).isEqualTo(new Timestamp(id));
```
dcapwell, line 36: this is the test referenced above... `Timestamp.fromString()` for a txn wouldn't work with `toString` so need a `toStandardString`

accord-core/src/test/java/accord/topology/TopologyManagerTest.java
```diff
@@ -146,48 +146,33 @@ void syncCompleteFor()
...
         Assertions.assertFalse(service.getEpochStateUnsafe(2).syncComplete());
         Assertions.assertTrue(service.getEpochStateUnsafe(2).syncCompleteFor(keys(150).toParticipants()));
         Assertions.assertFalse(service.getEpochStateUnsafe(2).syncCompleteFor(keys(250).toParticipants()));
     }
 
-    /**
-     * Epochs should only report being synced if every preceding epoch is also reporting synced
-     */
     @Test
-    void existingEpochPendingSync()
```
dcapwell, line 160: deleted the test as its no longer relevant... making sure that a new epoch doesn't resolve the past is no longer true... new epochs resolve the previous ones!

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -816,13 +816,20 @@ public IfBuilder<State, SystemUnderTest> add(Setup<State, SystemUnderTest> cmd)
                     CommandsBuilder.this.addIf(predicate, cmd);
                     return this;
                 }
+
+                @Override
+                public IfBuilder<State, SystemUnderTest> addIf(Predicate<State> nextPredicate, Setup<State, SystemUnderTest> cmd) {
```
dcapwell, line 821: used by Cassandra tests...

accord-core/src/main/java/accord/primitives/TxnId.java
belliottsmith: Accord doesn’t, so we shouldn’t support that here without a lot of wider work to make sure it is safe.

in fact epochs alone should immediately break as we truncate to only 48 bits.

belliottsmith: I think in principle negative node ids may be fine, but I would prefer not to find out if possible, and would prefer anyway to encode them as positive values (treat as unsigned ints)

accord-core/src/main/java/accord/topology/Shard.java
belliottsmith: This is probably a bug and we can just change it?

accord-core/src/main/java/accord/topology/TopologyManager.java
belliottsmith: Could you elaborate a bit on this? This is critical for protocol correctness so would be good to understand what is being removed. The comment appears to be against a dead link.

dcapwell: `syncComplete` is based off `accord.local.Node#onTopologyUpdateInternal`  and `accord.api.ConfigurationService.EpochReady#coordination`

```
ready.coordination.addCallback(() ->  {
            synchronized (AbstractConfigurationService.this)
            {
                localSyncComplete(epochs.getOrCreate(ready.epoch).topology, startSync);
            }
        });
```

This logic was changed (we talked about this a long time ago) so that each epoch is now sequential and we don't allow epoch=N to happen-before epoch=N-1 (currently N can be syncComplete before N-1, simple case is that N-1 is a node joining and must stream, and N is a table comment update).

If you look at `accord.local.Node#onTopologyUpdateInternal` coordination is now sequenced so `N` happens-after `N-1`.

In Apache Cassandra this is notifies `localSyncComplete` which marks the epoch as `epochState.setSyncStatus(SyncStatus.NOTIFYING);` and notifies the *nodes in the topology* about the epoch.  Those nodes will do the same which cause `accord.topology.TopologyManager#onEpochSyncComplete` to get called, which updates `syncComplete` (and `prevSyncComplete` and `curSyncComplete`)...

Now we have 2 cases that exist in Apache Cassandra: Expand Cluster, and Host Replacement.  In both of these cases a *new node* starts to join and will know about past epochs (as it's needed for processing), but won't be present in those past epochs.  This leads us to a situation where those past epochs have `syncComplete = []` and we include every single epoch for coordination.

By sequencing `EpochReady.coordination`, once this new node *sees* epoch=N be sync complete by its peers, it *knows* epoch=N-1 is also syncComplete.

Without this change, we would require every node in the cluster to notify the joining node about every epoch they know and their sync status... if the sync is also not complete then the node would be stuck with a partial update as it wouldn't learn about new ones!

accord-core/src/main/java/accord/primitives/TxnId.java
dcapwell: we spoke in slack but confirmed negative nodes isn't a thing in TCM.  I was getting confused about the migration case as `host_id: UUID` is based off `UUID.random` and the conversion logic to `int` would have allowed negatives... but the `host_id != nodeId.asUUID`, so looks like I was just confused

accord-core/src/main/java/accord/topology/Shard.java
dcapwell: spoke with @bdeggleston and this is really `non-readable` in Cassandra... so just a case of bad naming.  Right now this is dead code so it doesn't impact anything, but it is something we should flesh out... we spoke before that we have an issue with this case as we will try to read from non-readable nodes and have to fall back... this adds latency to user requests that could be avoided by leveraging this.

PR #105, opened 2024-07-19
CASSANDRA-19757

ifesdjeen:


accord-core/src/test/java/accord/utils/ReducingRangeMapTest.java
```diff
@@ -214,7 +215,8 @@ private void testRandomAdds(long seed, int numberOfMerges, int numberOfAdditions
         try
         {
             Random random = new Random(seed);
-            logger.info(id);
+            if (ENABLE_LOGGING)
```
ifesdjeen: The log was extremely spammy on a CI, so I had to disable this. Left the boolean here in case someone wants to debug this.

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -666,6 +673,11 @@ public final boolean isTruncated()
         return status().hasBeen(Status.Truncated);
     }
 
+    public static boolean isTruncated(Status status)
+    {
+        return status.hasBeen(Status.Truncated);
```
belliottsmith: I agree with David that this could be misleading, as it can include `Invalidated` and `is` should mean `=` not `>=`.

accord-core/src/test/java/accord/utils/ReducingRangeMapTest.java
belliottsmith: maybe we should default to fewer tests? but yeah, no big

belliottsmith (approved):

accord-core/src/main/java/accord/local/Command.java
ifesdjeen: Removed this part. Turns out it was not necessary after refactorings.

PR #106, opened 2024-07-19
CASSANDRA-19758: CommandsForKeyPruning

belliottsmith:
CommandsForKey should periodically self-prune, so as to continue functioning well in-between garbage collections. This is a bit complicated, as once we prune we are left with potentially incomplete information, and have to sometimes load per-command information from disk. But the payoff is ensuring CommandsForKey objects - which drive the majority of the state machine - are kept to a reasonable size.


accord-core/src/main/java/accord/api/Agent.java
```diff
@@ -74,6 +74,10 @@ default void onLocalBarrier(@Nonnull Seekables<?, ?> keysOrRanges, @Nonnull TxnI
      */
     long preAcceptTimeout();
 
+    long cfkHlcPruneDelta();
```
bdeggleston, line 83: could we get a little docstring about these 2?

accord-core/src/main/java/accord/local/CommandsForKey.java
```diff
@@ -1046,29 +1542,20 @@ static class InfoWithAdditions
...
+        if (newStatus.hasBallot)
+            ballot = command.acceptedOrCommitted();
+
         Timestamp depsKnownBefore = newStatus.depsKnownBefore(txnId, executeAt);
-        return computeInfoAndAdditions(insertPos, updatePos, txnId, newStatus, executeAt, depsKnownBefore, command.partialDeps().keyDeps.txnIds(key));
+        return computeInfoAndAdditions(insertPos, updatePos, txnId, newStatus, ballot, executeAt, depsKnownBefore, command.partialDeps().keyDeps.txnIds(key));
     }
 
-    private Object computeInfoAndAdditions(int insertPos, int updatePos, TxnId plainTxnId, InternalStatus newStatus, Timestamp executeAt, Timestamp depsKnownBefore, SortedList<TxnId> deps)
+    private Object computeInfoAndAdditions(int insertPos, int updatePos, TxnId plainTxnId, InternalStatus newStatus, Ballot ballot, Timestamp executeAt, Timestamp depsKnownBefore, SortedList<TxnId> deps)
```
bdeggleston: Could we rework the class hierarchy of TxnInfo/InfoAndAdditions so we're not returning Object here (and the calling method)?

accord-core/src/main/java/accord/local/CommandsForKey.java
```diff
@@ -372,112 +483,208 @@ public String toString()
...
+        Timestamp executeAtIfKnownElseTxnId()
+        {
+            return status == INVALID_OR_TRUNCATED ? this : executeAt;
+        }
     }
 
-    public static class TxnInfoWithMissing extends TxnInfo
+    public static class TxnInfoExtra extends TxnInfo
     {
         public final TxnId[] missing;
```
bdeggleston: could you add a comment explaining what's missing and/or make the name more descriptive?

accord-core/src/main/java/accord/local/CommandsForKey.java
```diff
@@ -372,112 +483,208 @@ public String toString()
...
                    ", missing=" + Arrays.toString(missing) +
                    '}';
         }
     }
 
     private final Key key;
     private final RedundantBefore.Entry redundantBefore;
+    private final TxnInfo prunedBefore;
+    // TODO (desired): avoid loading if we know a transaction that should have witnessed it has already been registered/applied
+    private final Object[] loadingPruned;
```
bdeggleston: Could this be renamed `loadingPrunedBtree` so the reason for its `Object[]` type is apparent?

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -546,6 +546,7 @@ public static void main(String[] args)
     public void testOne()
     {
         run(System.nanoTime());
+//        run(99448562360375L);
```
bdeggleston: missed seed

accord-core/src/main/java/accord/local/CommonAttributes.java
```diff
@@ -34,6 +34,7 @@ public interface CommonAttributes
     Route<?> route();
     PartialTxn partialTxn();
     @Nullable Seekables<?, ?> additionalKeysOrRanges();
+    // TODO (expected): we don't need PartialDeps, only
```
bdeggleston: I think you missed the rest of the sentence for the TODO here

accord-core/src/main/java/accord/primitives/PartialDeps.java
```diff
@@ -101,7 +98,7 @@ public PartialDeps reconstitutePartial(Ranges covering)
         if (covers(covering))
             return this;
 
-        return new PartialDeps(covering, keyDeps, rangeDeps);
+        return new PartialDeps(covering, keyDeps, rangeDeps, directKeyDeps);
```
bdeggleston: is this even reachable?

accord-core/src/main/java/accord/primitives/Deps.java
```diff
@@ -91,37 +103,55 @@ public Builder()
...
+            return new Deps(keyBuilder.build(),
+                            rangeBuilder == null ? RangeDeps.NONE : rangeBuilder.build(),
+                            directKeyBuilder == null ? KeyDeps.NONE : directKeyBuilder.build());
         }
     }
 
-    public final KeyDeps keyDeps;
+    public final KeyDeps keyDeps; // key dependencies we will track via CommandsForKey
     public final RangeDeps rangeDeps;
+    public final KeyDeps directKeyDeps; // key dependencies we will track via direct dependency relationships
```
bdeggleston: The key dependencies have been split into 2 tiers here, and it's not clear what the difference is and why they need to be separate. Can you add some documentation explaining the types of deps and how they are different and fit together?

accord-core/src/main/java/accord/local/SafeCommandStore.java
```diff
@@ -214,58 +214,64 @@ private void updateMaxConflicts(Command prev, Command updated)
...
-                    self.get(key).registerUnmanaged(self, self.get(t));
-                    return null;
-                }, this, txnId, null, i->false);
-            }
-            else
-            {
-                commandStore().execute(context, safeStore -> safeStore.updateCommandsForKey(prev, updated))
-                              .begin(commandStore().agent);
-            }
+            safeStore = safeStore;
```
bdeggleston: redundant assignment

accord-core/src/main/java/accord/local/SafeCommandStore.java
```diff
@@ -214,58 +214,64 @@ private void updateMaxConflicts(Command prev, Command updated)
...
+        TxnId txnId = next.txnId();
+        Keys keys = (Keys)next.keysOrRanges();
+        if (keys == null || next.hasBeen(Status.Truncated)) keys = (Keys)prev.keysOrRanges();
+        if (keys == null)
+            return;
+
+        // TODO (required): additionalKeysOrRanges may not be being handled entirely correctly here, though it may not matter.
+        //    Once committed without a given key, we should be effectively erasing the command from that CFK
+        PreLoadContext context = PreLoadContext.contextFor(txnId, keys, COMMANDS);
+        // TODO (expected): execute immediately for any keys we already have loaded, and save only those we haven't for later
```
bdeggleston, line 246: is that necessary? If we're going to start a second task to update some, why don't we just use it to update all of them and skip going through the trouble of splitting them up?

accord-core/src/main/java/accord/local/CommandsForKey.java
```diff
@@ -111,32 +111,42 @@
...
+    public static final TxnInfo NO_INFO = new TxnInfo(TxnId.NONE, HISTORICAL, TxnId.NONE);
     public static final TxnInfo[] NO_INFOS = new TxnInfo[0];
     public static final Unmanaged[] NO_PENDING_UNMANAGED = new Unmanaged[0];
 
+    public static boolean manages(TxnId txnId)
+    {
+        return txnId.domain().isKey() && txnId.kind().isGloballyVisible();
+    }
+
+    public static boolean managesExecution(TxnId txnId)
```
bdeggleston: seems like the names for this method and `managesKeyExecution` should be swapped?

accord-core/src/main/java/accord/local/CommandsForKeyUpdate.java
```diff
@@ -0,0 +1,170 @@
...
+import static accord.local.SaveStatus.LocalExecution.WaitingToExecute;
+
+public abstract class CommandsForKeyUpdate
+{
+    @VisibleForTesting
+    public abstract CommandsForKey cfk();
+    abstract ExtraNotify notifier();
+    abstract void notify(SafeCommandStore safeStore, @Nullable CommandsForKey prevCfk, @Nullable Command command, NotifySink notifySink);
+
+    void notify(SafeCommandStore safeStore, @Nullable CommandsForKey prevCfk, @Nullable Command command)
```
bdeggleston: unused

accord-core/src/main/java/accord/local/CommandsForKeyUpdate.java
```diff
@@ -0,0 +1,170 @@
...
+import accord.local.CommandsForKey.TxnInfo;
+import accord.primitives.Keys;
+import accord.primitives.Seekables;
+import accord.primitives.TxnId;
+
+import static accord.local.CommandsForKey.InternalStatus.PREACCEPTED_OR_ACCEPTED_INVALIDATE;
+import static accord.local.KeyHistory.COMMANDS;
+import static accord.local.SaveStatus.LocalExecution.WaitingToExecute;
+
+public abstract class CommandsForKeyUpdate
```
bdeggleston: Does this need to be its own class? My reading of it is that it's kind of a hardcoded command update listener, so why not just keep it as part of the command/cfk update flow instead of adding a layer of indirection here? I see that we plug in another notify sink in the CFK test. Is the test instrumentation something that could be integrated in the the InMemoryCommandStore and it's various Safe* classes and kept separate from the production code?

accord-core/src/main/java/accord/local/CommandsForKey.java
```diff
@@ -290,6 +302,75 @@ public static InternalStatus get(int ordinal)
         }
     }
 
+    static class LoadingPruned extends TxnId
```
bdeggleston: This could use a comment explaining what it does, and maybe a new name. I was able to work out it's a reverse deps index, mapping a txnid to txnids that have it as a dep. However I don't see the list updated in `pruneBefore` and I'm not sure what the loading part is referring to.

bdeggleston (changes requested):
A lot of the feedback I have is regarding readability. There have been a few rounds of changes since I was last in here and even though I'm pretty familiar with it functionally, I found some parts of the patch to be pretty difficult to follow.

The 2 main issues are:
* undocumented classes with somewhat vague or un-descriptive names names (what is a LoadingPruned?!).
* inconsistent naming between connected concepts(ie the connection between Write witnessed txns, managed txnIds, and direct and key deps could be better illustrated through a term shared between them as well as an explanation)

I didn't have time to give this an in-depth review before going out of town, but if you could give the patch a quick skim with an eye towards identifying things that could be clearer, that would be great, and I could wrap this up when I get back. Bonus points for taking a look at the perf-improvements patch with the same goal, I believe that was merged with the intention to review afterwards and some of the confusion originated there.

accord-core/src/main/java/accord/local/CommandsForKey.java
belliottsmith: There is some doc about this, but minimal and not spread widely enough, and probably not at the class level. I will update that to describe what's going on here. This is the "once we prune we are left with potentially incomplete information, and have to sometimes load per-command information from disk" part of the patch. 

Pruned TxnId are those we have decided we probably don't need as we expect them to be transitive dependencies of transactions we've already applied, but that could still appear as dependencies again, but also might just be unwitnessed transactions with an old id - and to know which it is, we have to load the TxnId to find out what the command-level data knows of it.

That is to say, pruning doesn't trigger this, but encountering a `TxnId` dependency that occurs before `prunedBefore` does.

belliottsmith: That is to say, pruning doesn't trigger this, but encountering a `TxnId` dependency that occurs before `prunedBefore` does.

accord-core/src/main/java/accord/local/CommandsForKeyUpdate.java
belliottsmith: We need the indirection because these can trigger updates to `CommandsForKey`- which we may not have actually yet updated in the main update flow. So we first complete updating CFK, then we trigger any "listeners" / notifications.

accord-core/src/main/java/accord/local/CommandsForKey.java
belliottsmith: No, but perhaps needs renaming. This just a method that knows it's testing a key, so can avoid performing the key test.

accord-core/src/main/java/accord/local/SafeCommandStore.java
belliottsmith: Because it's better to perform the update immediately, where possible? Especially as any continuation goes to the back of the processing queue, which might cause everything to take a lot longer (e.g. there may be some command that runs next that cannot now execute because it's waiting for the last command to update CFK, so we now have at least two additional continuations and queue delays)

accord-core/src/main/java/accord/local/CommandsForKey.java
belliottsmith: I don't like adding types to parameter names, but I will certainly document it with comments. Though, I think in the Cassandra codebase `Object[]` is very likely to be a BTree so it's mentally a BTree marker for me without any additional information, but perhaps that is arcane knowledge.

accord-core/src/main/java/accord/local/CommandsForKey.java
belliottsmith: I would prefer not to make the name any longer. There is a brief comment on the accessor method that explains what this is, that I will expand and reference here.

This is also explained in detail in the class level documentation.

accord-core/src/main/java/accord/local/CommandsForKey.java
belliottsmith: I don't think so, at least not without making the allocations larger, and performance is a key concern here. I don't think in this particular case it is onerous, we know it is one of only two types. I can add documentation to make this clear.

accord-core/src/main/java/accord/api/Agent.java
belliottsmith: Yep

belliottsmith:
> inconsistent naming between connected concepts(ie the connection between Write witnessed txns, managed txnIds, and direct and key deps could be better illustrated through a term shared between them as well as an explanation)

I agree this is an area that is under-documented. I will expand the class level documentation to explain the semantics here. I'm not sure there's a good shared name, but to briefly explain:

CFK does two jobs today: 1) compute dependencies, for which it needs to track everything that might be witnessed (i.e. `manages`); 2) trigger local execution once dependencies have executed (i.e. `managesExecution`).

The CFK only directly manages the execution of read and write transactions, essentially. The others listen to the CFK only to know when any read/write dependencies have executed. All other key transactions are managed as "direct" dependencies, via the old/range txn scheme of `NotifyWaitingOn` and tracking the `TxnId` directly in `WaitingOn`.

This isn't actually different to before, except that we now separate out the `directKeyDeps` (those we must take direct dependencies on) and plain `keyDeps` (those we take a dependency on only the "Key" (i.e. we rely on CFK managed execution)). The reason we separate these out early on is because we do actually need to be able to filter these en masse using the Keydets class if and when we have shard-level redundancy information. This was to address an existing bug.

The reason we can't easily let CFK manage the execution of these transactions is that EphemeralReads and ExclusiveSyncPoints don't actually have an execution time, so we would just build a similar kind of logic within CFK to be able to support them in CFK. SyncPoints on the other hand just complicate the execution logic because they aren't visible to Writes, and the efficiency of the approach taken today depends on knowing we won't execute anything behind the last applied write. This element could certainly be reworked in future, but it seemed rational to defer any further improvements there, and to lean on our existing logic.

accord-core/src/main/java/accord/local/RedundantBefore.java
```diff
@@ -109,6 +109,8 @@ public Entry(Range range, long startEpoch, long endEpoch, @Nonnull TxnId locally
             this.shardAppliedOrInvalidatedBefore = shardAppliedOrInvalidatedBefore;
             this.bootstrappedAt = bootstrappedAt;
             this.staleUntilAtLeast = staleUntilAtLeast;
+            Invariants.checkArgument(locallyAppliedOrInvalidatedBefore.equals(TxnId.NONE) || locallyAppliedOrInvalidatedBefore.domain().isRange());
```
dcapwell, line 112: so this just blocks key txn from showing up?  In BurnTest we can do `Range` read/write and I think we wanted to add range reads in C* eventually, so is `range` the best here or is it really `ExclusiveSyncPoint` that we want?

accord-core/src/main/java/accord/messages/Commit.java
```diff
@@ -161,6 +161,7 @@ protected Commit(Kind kind, TxnId txnId, PartialRoute<?> scope, long waitForEpoc
 
     public static void commitMinimal(Node node, Topologies coordinateEpochOnly, Ballot ballot, TxnId txnId, Txn txn, FullRoute<?> route, Timestamp executeAt, Deps unstableDeps, Callback<ReadReply> callback)
     {
+        Invariants.checkArgument(coordinateEpochOnly.size() == 1);
```
dcapwell: for debugging can you show what epochs are included?

accord-core/src/main/java/accord/utils/Invariants.java
```diff
@@ -27,13 +27,16 @@
 
 public class Invariants
 {
-    // TODO (now): configure by system parameter and turn off by default
-    private static final boolean PARANOID = true;
-    private static final boolean DEBUG = true;
+    private static final int PARANOIA = Integer.parseInt(System.getProperty("accord.paranoia", "0"));
+    private static final boolean DEBUG = System.getProperty("accord.debug", "false").equals("true");
```
dcapwell: does this compile in C*?  We have a check style (that does check Accord code) that we don't touch these APIs

accord-core/src/test/java/accord/impl/PrefixedIntHashKey.java
```diff
@@ -310,6 +310,10 @@ public int compareTo(@Nonnull RoutableKey that)
                     int rightInt = that instanceof Sentinel ? (((Sentinel) that).isMin ? -1 : 1) : 0;
                     rc = Integer.compare(leftInt, rightInt);
                 }
+                else if (!(this instanceof PrefixedIntRoutingKey) && (!(that instanceof PrefixedIntRoutingKey)))
```
dcapwell: `PrefixedIntRoutingKey -> Hash | Sentinel`
`PrefixedIntHashKey -> Hash | Sentinel | Key`

so this is just saying `else if (this.isKey() && that.isKey())`?  Since you are checking the key it might be best to do that?

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
```diff
@@ -232,7 +235,7 @@ public <T> AsyncChain<T> submit(PreLoadContext context, Function<? super SafeCom
         public <T> AsyncChain<T> submit(Callable<T> fn)
         {
             Task<T> task = new DelayedTask<>(fn);
-            if (Invariants.isParanoid())
+            if (Invariants.paranoia() >= 3)
```
dcapwell: an issue I have with this is there isn't really any documentation on what levels are allowed or what they mean... this might be a better case for an enum then int level?

accord-core/src/test/java/accord/local/CommandsForKeyTest.java
```diff
@@ -0,0 +1,1067 @@
...
+            test(seed++, 1000);
+        }
+    }
+
+    private static void test(long seed, int minCount)
+    {
+        logger.info("Seed {}", seed);
+        try
+        {
+            final RandomSource rnd = RandomSource.wrap(new Random(seed));
```
dcapwell:
```suggestion
            final RandomSource rnd = new DefaultRandom(seed);
```

accord-core/src/test/java/accord/utils/AccordGens.java
```diff
@@ -227,7 +236,14 @@ public static Gen<Key> keysInsideRanges(Ranges ranges)
 
     public static Gen<KeyDeps> keyDeps(Gen<? extends Key> keyGen)
```
dcapwell, line 258: this is used in maelstrom and `org.apache.cassandra.utils.AccordGenerators#keyDepsGen()`, need to double check the C* to make sure this change is fine.

dcapwell (commented):
first pass at the small stuff, ill do a pass at C* as well then will try to focus on the larger stuff

accord-core/src/main/java/accord/local/RedundantBefore.java
belliottsmith: We shouldn’t be inserting them into this map is all - this is a map for GC and we don’t want that to be at the point/key level

accord-core/src/main/java/accord/utils/Invariants.java
belliottsmith: It compiles. We don’t have the same API available to accord. We could introduce a parallel one for Accord

accord-core/src/test/java/accord/impl/PrefixedIntHashKey.java
belliottsmith: I think the key class might be a common super class so we don’t know if it is a key, only if it isn’t? Will double check

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
belliottsmith: That’s totally fair - just thought this was a flexible and easy thing but fine with defining some descriptive enums like log levels

accord-core/src/test/java/accord/impl/PrefixedIntHashKey.java
dcapwell: `Key` isn't a common super class, it's the final top level class.  We have 3 top level types: `Key | Hash | Sentinel`

dcapwell:
ran the tests locally while reviewing and `accord.local.CommandsForKeyTest` is failing.

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -1138,7 +1081,6 @@ private Committed(CommonAttributes common, SaveStatus status, Timestamp executeA
             super(common, status, promised, executeAt, accepted);
             this.waitingOn = waitingOn;
             Invariants.checkState(common.route().kind().isFullRoute(), "Expected a full route but given %s", common.route().kind());
-            if (status.hasBeen(Stable)) Invariants.checkState(waitingOn == WaitingOn.EMPTY || (waitingOn.txnIds.equals(common.partialDeps().rangeDeps.txnIds()) && waitingOn.keys.equals(common.partialDeps().keyDeps.keys())), "Deps do not match; expected (%s, %s) == (%s, %s)", waitingOn.keys, waitingOn.txnIds, common.partialDeps().keyDeps.keys(), common.partialDeps().rangeDeps.txnIds());
```
dcapwell, line 1143: why was this check removed?

belliottsmith:
>ran the tests locally while reviewing and accord.local.CommandsForKeyTest is failing.

oh neat, I added an additional invariant late in the game and hadn't run the tests since. Will resolve it tomorrow, along with merging your fix. thanks

belliottsmith:
Turns out the problem with the failing test was different - I seemed to have lost a small recent commit somehow. But, it is resolved now.

accord-core/src/main/java/accord/primitives/PartialDeps.java
belliottsmith: hmm, interesting. it looks like it cannot, good catch.

accord-core/src/main/java/accord/local/SafeCommandStore.java
belliottsmith: This is to ensure no accidental usage inside the lambda; will comment to this effect, but it's a norm I would like to encourage, esp. for dangerous references such as this.

belliottsmith:
quick question: what do you think of splitting `CommandsForKey` into its own package? Don't have a strong feeling, but feel like it might have enough ancillary internal classes and complexity that it would aid in comprehension. I wouldn't want to split them all into a higher class in a shared package, however, as it would badly pollute that. So, could introduce accord.local.cfk to contain all of its inner classes and use package-private visibility rather than private where appropriate.

belliottsmith:
Ok, I think I have addressed everything brought up so far

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
```diff
@@ -278,6 +284,7 @@ public DelayedSafeStore(DelayedCommandStore commandStore, RangesForEpoch ranges,
         @Override
         public void postExecute()
         {
+            super.postExecute();
```
ifesdjeen, line 287: This call would invalidate safe InMemorySafeCommand, which will make `safe.isModified` to return `false`, and we will skip persisting anything in the journal.

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
```diff
@@ -163,6 +166,9 @@ public DelayedCommandStore(int id, NodeTimeService time, Agent agent, DataStore
         @Override
         public void validateRead(Command current)
         {
+            if (!Invariants.testParanoia(LINEAR, LINEAR, HIGH))
```
ifesdjeen, line 169: I would argue that for any burn test run `validateRead` should be enabled. This is a crucial part of recovery correctness, would be great if we exercise it a lot. If there are bugs here, I will be happy to assist.

accord-core/src/main/java/accord/utils/SimpleBitSet.java
```diff
@@ -419,6 +419,19 @@ private <P1, P2, P3, P4> void reverseForEach(int index, long mask, P1 p1, P2 p2,
         }
     }
 
+    public void intersect(SimpleBitSet with)
```
dcapwell: this method doesn't do anything... as long as `this.size == with.size` we spend time computing things, but we don't actually do anything with what we compute.

IntelliJ also says this is dead code, can we remove?

accord-core/src/main/java/accord/utils/btree/BTree.java
```diff
@@ -501,6 +501,23 @@ else if (c < 0)
         }
     }
 
+    public static <Compare, Existing extends Compare, Insert extends Compare> Object[] updateExisting(Object[] toUpdate,
```
dcapwell, line 504: style

accord-core/src/main/java/accord/utils/SimpleBitSet.java
belliottsmith: It does do something (it updates one of the two arrays) but yes it’s dead code so will delete - might even be incomplete anyway as from an abandoned change I otherwise deleted.

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -1484,47 +1485,70 @@ public boolean removeWaitingOn(Key key)
...
                 this.executeAtLeast = Timestamp.nonNullOrMax(executeAtLeast, this.executeAtLeast);
             }
 
-            boolean removeWaitingOnTxnId(int i)
+            boolean removeWaitingOnDirectRangeTxnId(int i)
             {
                 return removeWaitingOn(i);
             }
 
+            boolean removeWaitingOnDirectKeyTxnId(int i)
```
dcapwell, line 1546: the remove index methods don't actually check that their input is correct, we should add a check... I added that to my feedback branch

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -856,8 +832,11 @@ static void updateDependencyAndMaybeExecute(SafeCommandStore safeStore, SafeComm
         }
     }
 
-    static void removeWaitingOnKeyAndMaybeExecute(SafeCommandStore safeStore, SafeCommand safeCommand, Key key, boolean executeAsync)
+    static void removeWaitingOnKeyAndMaybeExecute(SafeCommandStore safeStore, SafeCommand safeCommand, Key key)
     {
+        if (safeStore.time().id().id == 3 && safeCommand.txnId().toString().equals("[3,25616,2(KW),3]") && key.toString().startsWith("0:1#"))
```
dcapwell: I assume this is for debugging... you do the same as me it seems ^_^

accord-core/src/main/java/accord/utils/btree/BTree.java
dcapwell: fixed in my feedback branch

accord-core/src/test/java/accord/burn/BurnTest.java
dcapwell: fixed in my feedback branch

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -594,8 +593,68 @@ public final boolean isRejectedIfNotPreAccepted(TxnId txnId, Unseekables<?> part
...
+            /**
+             * Are the participating ranges for the txn fully covered by bootstrapping ranges for this command store
+             */
+            boolean isFullyBootstrapping(WaitingOn.Update builder, Range range, int txnIdx)
+            {
+                if (builder.directKeyDeps.foldEachKey(txnIdx, range, true, (r0, k, p) -> p && r0.contains(k)))
+                    return true;
+
+                if (partiallyBootstrapping == null)
+                    partiallyBootstrapping = new HashMap<>();
```
dcapwell: `org.agrona.collections.Int2ObjectHashMap`?

accord-core/src/main/java/accord/local/Command.java
belliottsmith: It didn't seem as useful as it was given how we load, and also that we don't set it to EMPTY anymore, but in retrospect I will reinstate an updated version with suitable paranoia gating.

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
belliottsmith: I don't think it should be enabled by default, as we don't *usually* touch this and it has a significant performance penalty. I think we should introduce a lighter-weight version that doesn't perform the intrusive diffs but still optionally loads from the journal that would also catch the majority of problems, that can be on by default.

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
belliottsmith: good catch, looks like a bad (or brain-dead) merge

dcapwell (commented):
need to call it for the day, wrapped up `accord.local.CommandStore#removeRedundantDependencies` which looks fine... hopefully will be able to finish tomorrow or Thursday...

I made sure to push everything to `feedback/cfk-pruning` in C* and Accord repos

accord-core/src/main/java/accord/local/Command.java
belliottsmith: I've adopted all of your suggestions, the only ones I have qualms about are here, as these methods are used in tight loops and want to be as efficient as possible. But, perhaps we can leave it as an exercise for later to facilitate eliding these bounds checks.

belliottsmith:
Thanks, all great suggestions - have merged a squash commit.

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: heh :) 

usually I use conditional breakpoints but sometimes they're too slow

belliottsmith:
I have also pushed an update to manage range transaction _dependencies_ in `CommandsForKey` as this is necessary for `mapReduce` on recovery. That is, when recovering a range transaction we need a way to know whether Key operations have witnessed it. It's not ideal, but I don't see a better all-round approach. 

This change is likely invisible to Cassandra until we implement full range reads, as we only utilise range exclusive sync points - which are invisible to everything except other exclusive sync points (which we never issue as Key operations in Cassandra). *BUT* we need it for correctness of the burn test that _does_ issue range transactions today, and it sets us up to handle it properly in Cassandra later.

accord-core/src/main/java/accord/utils/MergeFewDisjointSortedListsCursor.java
```diff
@@ -0,0 +1,149 @@
...
+    {
+        private final SortedList<? extends T> list;
+        private int itemIdx;
+        private T item;
+
+        public Candidate(@Nonnull SortedList<? extends T> list)
+        {
+            Invariants.checkState(!list.isEmpty());
+            this.list = list;
+            this.item = list.get(0);
```
dcapwell, line 36: the `SortedList` interface does not block `null`, and `null` will trick this class to think the cursor is fully consumed.

* `SortedArrayList` and `ExtendedSortedArrayList` will NPE as it doesn't allow overriding the `Comparator`
* `SortedRelationList` will accept `null` but most methods that touch the data will NPE.  This is public and only using in the Deps classes, so current usage doesn't have `null` (things can always change)

if `SortedList` documented that `null` is not allowed and `SortedRelationList` validates its input, then the lack of `null` check here is fine.  The main concern is not the code as it is today, but how things could evolve in the future... things get reused and assumptions change, so its good to document and/or enforce assumptions early on.

accord-core/src/main/java/accord/primitives/Txn.java
```diff
@@ -397,7 +397,7 @@ default Writes execute(TxnId txnId, Timestamp executeAt, @Nullable Data data)
     default AsyncChain<Data> read(SafeCommandStore safeStore, Timestamp executeAt, Ranges unavailable)
     {
         Ranges ranges = safeStore.ranges().allAt(executeAt).subtract(unavailable);
-        List<AsyncChain<Data>> chains = Routables.foldlMinimal(keys(), ranges, (key, accumulate, index) -> {
+         List<AsyncChain<Data>> chains = Routables.foldlMinimal(keys(), ranges, (key, accumulate, index) -> {
```
dcapwell: style issue... fixed in my branch

accord-core/src/main/java/accord/utils/MergeFewDisjointSortedListsCursor.java
belliottsmith: I don't mind changing this to use something besides null to demarcate the end, and also adding null checks to SortedList. Worth noting (as you allude to) that by definition the `SortedList` cannot safely accept nulls already - `SortedList` does not accept a comparator, it operates on `Comparable` so all of its operations require the inputs be non-null already.

belliottsmith: Hmm actually we already check non-null by checking `isSortedUnique` in the constructors. But I'm updating these validations a little and will note this.

dcapwell: > Hmm actually we already check non-null by checking isSortedUnique in the constructors.

`SortedArrayList` takes `Comparable` and doesn't allow overriding `Comparator`, so `isSortedUnique` will NPE if you have a null element.  `SortedRelationList` has no such validation, so you could create one with null data, it just blows up once you touch any element.

dcapwell:
```
Invariants.checkState(candidate.itemIdx == candidate.list.size());
```

that change you made should be enough, once we see `null` and we didn't consume everything we fail.  Thats fine for me

accord-core/src/main/java/accord/local/SafeCommandStore.java
bdeggleston: ah nice, that's a good idea

bdeggleston:
> quick question: what do you think of splitting `CommandsForKey` into its own package? Don't have a strong feeling, but feel like it might have enough ancillary internal classes and complexity that it would aid in comprehension. I wouldn't want to split them all into a higher class in a shared package, however, as it would badly pollute that. So, could introduce accord.local.cfk to contain all of its inner classes and use package-private visibility rather than private where appropriate.

+1 to that. It's grown to a point where that's reasonable

accord-core/src/main/java/accord/primitives/Deps.java
```diff
@@ -55,12 +79,23 @@ public static abstract class AbstractBuilder<T extends Deps> implements AutoClos
 
         public AbstractBuilder<T> add(Seekable keyOrRange, TxnId txnId)
         {
-            switch (keyOrRange.domain())
+            Invariants.checkArgument(keyOrRange.domain() == txnId.domain(), keyOrRange + " is not same domain as " + txnId);
```
dcapwell, line 82: I don't think this check is correct, and in my new testing I hit it.

```
Caused by: java.lang.IllegalArgumentException: 94851792 is not same domain as [1,157613,1(RR),42]
	at accord.utils.Invariants.illegalArgument(Invariants.java:84)
	at accord.utils.Invariants.checkArgument(Invariants.java:221)
	at accord.primitives.Deps$AbstractBuilder.add(Deps.java:82)
	at accord.messages.PreAccept.lambda$calculatePartialDeps$0(PreAccept.java:257)
	at accord.local.CommandsForKey.mapReduceActive(CommandsForKey.java:1064)
	at accord.impl.InMemoryCommandStore$InMemorySafeStore.lambda$mapReduceActive$4(InMemoryCommandStore.java:854)
	at accord.impl.InMemoryCommandStore.mapReduceForKey(InMemoryCommandStore.java:288)
	at accord.impl.InMemoryCommandStore$InMemorySafeStore.mapReduceActive(InMemoryCommandStore.java:853)
	at accord.messages.PreAccept.calculatePartialDeps(PreAccept.java:254)
```

a range read provided its `Ranges`.  And range transactions find both range and key transactions, only key txn have this specific property that the domain matches

dcapwell: The following check should work better

```
Invariants.checkArgument(txnId.domain().isRange() || keyOrRange.domain() == txnId.domain(), keyOrRange + " is not same domain as " + txnId);
```

accord-core/src/main/java/accord/local/CommandsForKeyUpdate.java
bdeggleston: Got it, the comment you added the class really clarified things, thanks

accord-core/src/main/java/accord/local/SafeCommandStore.java
bdeggleston: yeah that makes sense. I think I was worried about edge cases caused by updating some CFKs for an operation and not others? Thinking about it again though, that seems like it should be ok and anyways pretty well covered by the burn tests.

belliottsmith:
Ok, I've pushed an attempt at cleaning up CFK into separate classes in its own package.

I've split out some of the logic to group by functionality - effectively into modules, where we pass the CFK as a parameter. This isn't optimal, but I think on balance it is better than keeping all of the functionality in the CFK class.

Let me know what you think.

accord-core/src/main/java/accord/local/CommandsForKey.java
bdeggleston: comment on the field is fine

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -594,8 +594,68 @@ public final boolean isRejectedIfNotPreAccepted(TxnId txnId, Unseekables<?> part
...
+                // note that we must exclude all transactions that are pre-bootstrap, and perform the more complicated dance below,
+                // as these transactions may be only partially applied, and we may need to wait for them on another key.
+                if (appliedIdx > bootstrapIdx)
+                {
+                    d.forEach(e.range, bootstrapIdx, appliedIdx, b, s, (b0, s0, txnIdx) -> {
+                        b0.removeWaitingOnDirectKeyTxnId(txnIdx);
+                    });
+                }
+
+                if (bootstrapIdx > 0)
```
dcapwell, line 648: sent you this in slack.  `e.bootstrappedAt` is set *locally* w/e we do a new `Bootstrap.Attempt`, this is called in `store.markBootstrapping(safeStore0, globalSyncId, valid);`

Now, the issue is that we then *start* the `CoordinateSyncPoint.exclusive` after and this logic doesn't know the state of the bootstrap (was it invalidated? Was it applied? is it blocked on dependencies?).

So, lets say we have the following

1) node1 starts a RX but the messages don't reach node2 right away
2) node2 sees a topology change so starts a bootstrap
3) node2 finally sees the RX from node1

in this history the RX from node1 will mark all its range dependencies as no longer waiting as the new bootstrap (which *should* block on this RX) is newer (aka would be placed at the end of the list: `d.txnIds().find(e.bootstrappedAt)`).  This than means the RX from node1 is able to run right away and disregard its dependencies!

In my testing I hit this: `[2,91574,9(RX),42]` took a dependency on `[2,3728,7(RS),42]` (which is in Stable state, waiting on its dependencies), but a Bootstrap.Attempt has started so marks `[2,3728,7(RS),42]` as no longer waiting.  This caused `[2,91574,9(RX),42]` to reach `Apply` which allowed `accord.local.CommandStore#markShardDurable` to be called *before* `[2,3728,7(RS),42]` got unblocked... once it got unblocked it failed with 

```
Caused by: java.lang.IllegalStateException: Loading universally-durable command [2,3728,7(RS),42] that has been PreCommitted but not Applied: Stable
	at accord.utils.Invariants.createIllegalState(Invariants.java:64)
	at accord.utils.Invariants.illegalState(Invariants.java:69)
	at accord.utils.Invariants.illegalState(Invariants.java:74)
	at accord.local.Cleanup.shouldCleanup(Cleanup.java:110)
	at accord.local.Cleanup.shouldCleanup(Cleanup.java:99)
	at accord.local.Cleanup.shouldCleanup(Cleanup.java:93)
	at accord.local.Commands.maybeCleanup(Commands.java:946)
	at accord.local.SafeCommandStore.maybeTruncate(SafeCommandStore.java:134)
```

accord-core/src/main/java/accord/coordinate/Propose.java
```diff
@@ -77,6 +78,7 @@ abstract class Propose<R> implements Callback<AcceptReply>
         this.callback = callback;
         this.acceptOks = new ArrayList<>();
         this.acceptTracker = new QuorumTracker(topologies);
+        Invariants.checkState(txnId.kind().isSyncPoint() || deps.keyDeps.txnIdCount() == 0 || deps.keyDeps.txnId(deps.keyDeps.txnIdCount() - 1).compareTo(executeAt) < 0);
```
dcapwell: can we get a useful error msg to know what happened?

accord-core/src/main/java/accord/local/CommandStore.java
dcapwell: I posted this comment to the wrong line... this is the range logic and I posted to the directKey logic...

accord-core/src/test/java/accord/local/cfk/CommandsForKeyTest.java
```diff
@@ -0,0 +1,1081 @@
...
+import static accord.local.Command.NotDefined.notDefined;
+import static accord.local.Status.Durability.NotDurable;
+
+// TODO (expected): test setting redundant before
+// TODO (expected): test ballot updates
+// TODO (expected): test transition to Erased
+// TODO (expected): ensure execution is not too early
+// TODO (expected): validate mapReduce
+// TODO (expected): insert linearizability violations and detect them
+public class CommandsForKeyTest
```
dcapwell, line 97: this class doesn't have any asserts... so if bad state happens its allowed...

To prove this point to myself I added the following regression....

```
static CommandsForKey Pruning.maybePrune(CommandsForKey cfk, int pruneInterval, long minHlcDelta)
    {
            // prune everything!
            if (cfk.size() == 0)
                return cfk;
            return pruneBefore(cfk, cfk.byId[0], 0);
```

I ran `testMany` and its green... we do a ton of work and spend a ton of time, but we don't validate the state of the CFK...

belliottsmith: We do validate quite a lot actually, though most of it depends on `checkIntegrity` in CFK, which requires command line parameters to be specified. We also verify that all of the transactions execute. What we don't currently check is that they don't execute too early, so I think this particular corruption would go undetected anyway - since an almost empty CFK is perfectly valid in itself.

belliottsmith: I had a TODO to detect early execution, so I just added this and that change now fails the test

belliottsmith: Oh, hold on. This wasn't pruning. We do fail even without updating the test if we do 

        int len = cfk.byId.length;
        return pruneBefore(cfk, cfk.byId[len - 1], len - 1);

bdeggleston (approved):
+1

dcapwell (approved):
Ok did as much as I could before my deadline... sorry for not finishing on time =(

I cleaned up `feedback/cfk-pruning` so now it doesn't have the stateful test I was working on, but does have mostly minor feedback or bug fixes.  Adds a few more tests as well.

I got half way through `accord.local.cfk.Utils`, sadly the half remaining are the more complex methods =(

I didn't have time to closely look at accord-core/src/main/java/accord/local/cfk/Updating.java

PR #107, opened 2024-07-30
CASSANDRA-19805: ExclusiveSyncPoints should execute on all intersecting epochs

belliottsmith:


accord-core/src/main/java/accord/coordinate/ExecuteSyncPoint.java
```diff
@@ -68,17 +77,79 @@ public void start()
         }
     }
 
+    public static class ExecuteExclusiveSyncPoint extends ExecuteSyncPoint<Ranges>
```
dcapwell, line 80: I migrated `ListStore` to use this and it fails with

```
java.lang.IllegalArgumentException: No intersection
		at accord.messages.TxnRequest.computeScope(TxnRequest.java:287)
		at accord.messages.ReadData.<init>(ReadData.java:116)
		at accord.messages.WaitUntilApplied.<init>(WaitUntilApplied.java:51)
		at accord.coordinate.ExecuteSyncPoint$ExecuteExclusiveSyncPoint.lambda$start$0(ExecuteSyncPoint.java:96)
		at accord.local.Node.lambda$send$10(Node.java:501)
		at java.base/java.lang.Iterable.forEach(Iterable.java:75)
		at accord.local.Node.send(Node.java:501)
		at accord.local.Node.send(Node.java:494)
		at accord.coordinate.ExecuteSyncPoint$ExecuteExclusiveSyncPoint.start(ExecuteSyncPoint.java:96)
		at accord.impl.list.ListStore$Await.start(ListStore.java:528)
		at accord.impl.list.ListStore.awaitSyncPoint(ListStore.java:502)
```

I moved `Await.coordinate` into a `awaitSyncPoint` method to make it easier for plug out for me... and with that I have

```
private static AsyncChain<SyncPoint<Ranges>> awaitSyncPoint(Node node, SyncPoint<Ranges> exclusiveSyncPoint)
    {
        Await e = new Await(node, exclusiveSyncPoint);
        e.addCallback(() -> node.configService().reportEpochRedundant(exclusiveSyncPoint.keysOrRanges, exclusiveSyncPoint.syncId.epoch()));
        e.start();
        return e.recover(t -> {
            if (t.getClass() == SyncPointErased.class)
                return AsyncChains.success(null);
            if (t instanceof Timeout ||
                // TODO (expected): why are we not simply handling Insufficient properly?
                t instanceof RuntimeException && "Insufficient".equals(t.getMessage()) ||
                t instanceof SimulatedFault)
                return awaitSyncPoint(node, exclusiveSyncPoint);
            // cannot loop indefinitely
            if (t instanceof RuntimeException && "Redundant".equals(t.getMessage()))
                return AsyncChains.success(null);
            return null;
        });
    }

    private static class Await extends ExecuteSyncPoint.ExecuteExclusiveSyncPoint
    {
        public Await(Node node, SyncPoint<Ranges> syncPoint)
        {
            super(node, syncPoint, AppliedTracker::new);
        }

        @Override
        public void start()
        {
            super.start();
        }
    }
```

This doesn't include table removal, this is just ranges *leaving* the local node

PR #108, opened 2024-08-01
CASSANDRA-19321 Accord: Command to Exclude Replicas from Durability Status Coordination

maedhroz:


accord-core/src/main/java/accord/topology/Topologies.java
```diff
@@ -64,6 +64,11 @@ default boolean isEmpty()
 
     Set<Node.Id> nodes();
 
+    default Set<Node.Id> nonStaleNodes()
+    {
+        throw new UnsupportedOperationException();
```
dcapwell: why?  this can't be accepted for any impl so by not defining it makes it clear to know where to add; no?

dcapwell: if you want to block in `Multi` then we should be more explicit about that... and the callers of this should enforce we don't have multi (which is allowed for RX transactions)

accord-core/src/main/java/accord/topology/Topology.java
```diff
@@ -60,8 +61,10 @@ public class Topology
 {
     public static final long EMPTY_EPOCH = 0;
     private static final int[] EMPTY_SUBSET = new int[0];
-    public static final Topology EMPTY = new Topology(null, EMPTY_EPOCH, new Shard[0], Ranges.EMPTY, Collections.emptyMap(), Ranges.EMPTY, EMPTY_SUBSET);
+    public static final Topology EMPTY = new Topology(null, EMPTY_EPOCH, Collections.emptySet(), new Shard[0], Ranges.EMPTY, Collections.emptyMap(), Ranges.EMPTY, EMPTY_SUBSET);
+
     final long epoch;
+    final Set<Id> staleIds;
```
dcapwell, line 68: to keep the pattern shouldn't this be in `Shard`?  In C* we cache things, so we won't have a ton of duplicates.

dcapwell (changes requested):

accord-core/src/main/java/accord/topology/Topology.java
maedhroz: Talked about this offline. I'm going to see if I can find a way to address both of these comments by implementing `nonStaleNodes()` in `Multi` w/ the stale set from the latest `Topology`...

belliottsmith: This could also quite easily be represented as a bitset per shard (we could even realistically store it as a single long in cases where we have rf<64, but probably not worth optimising that far if we can just have a null bitset when empty)

PR #109, opened 2024-08-02
Fix various bugs, improving burn test stability

belliottsmith:
This patch fixes various bugs as well as deficiencies in some of the abstractions

- Remove concept of non-participating home keys; home keys are required to be a participant in the transaction
- Remove covering/covers concept
- Fix various invalidation/truncation/erase behaviours

### DIFF
--- a/accord-core/src/main/java/accord/coordinate/ExecuteSyncPoint.java
+++ b/accord-core/src/main/java/accord/coordinate/ExecuteSyncPoint.java
@@ -109,7 +109,7 @@ public abstract class ExecuteSyncPoint<S extends Seekables<?, ?>> extends Settab
         }
         else
         {
-            // TODO (required): we also need to handle ranges not being safe to read
+            // TODO (required, consider): do we need to handle ranges not being safe to read
             if (tracker.recordSuccess(from) == RequestStatus.Success)
                 onSuccess();
         }
--- a/accord-core/src/main/java/accord/coordinate/Infer.java
+++ b/accord-core/src/main/java/accord/coordinate/Infer.java
@@ -31,8 +31,10 @@ import accord.local.SafeCommand;
 import accord.local.SafeCommandStore;
 import accord.local.Status;
 import accord.local.Status.Known;
+import accord.messages.CheckStatus.FoundKnown;
+import accord.messages.CheckStatus.FoundKnownMap;
 import accord.primitives.EpochSupplier;
-import accord.primitives.Participants;
+import accord.primitives.Range;
 import accord.primitives.Ranges;
 import accord.primitives.Route;
 import accord.primitives.Timestamp;
@@ -43,6 +45,7 @@ import accord.utils.MapReduceConsume;
 
 import static accord.coordinate.Infer.InvalidIf.IfPreempted;
 import static accord.coordinate.Infer.InvalidIf.IfQuorum;
+import static accord.coordinate.Infer.InvalidIf.NotInvalid;
 import static accord.coordinate.Infer.InvalidIf.NotKnown;
 import static accord.coordinate.Infer.InvalidIfNot.IfUndecided;
 import static accord.coordinate.Infer.InvalidIfNot.IfUnknown;
@@ -52,7 +55,9 @@ import static accord.local.Status.PreApplied;
 import static accord.local.Status.PreCommitted;
 import static accord.primitives.Route.castToRoute;
 import static accord.primitives.Route.isRoute;
+import static accord.utils.Invariants.illegalState;
 
+// TODO (required): dedicated randomised testing of all inferences
 public class Infer
 {
     public enum InvalidIfNot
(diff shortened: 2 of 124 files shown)


accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -538,28 +541,38 @@ public Topologies withUnsyncedEpochs(Unseekables<?> select, Timestamp min, Times
...
 
-    private Topologies withSufficientEpochs(Unseekables<?> select, long minEpoch, long maxEpoch, Function<EpochState, Ranges> isSufficientFor)
+    public Topologies withUncompletedEpochs(Unseekables<?> select, EpochSupplier min, EpochSupplier max, EpochSufficiencyMode mode)
+    {
+        return withSufficientEpochs(select, min.epoch(), max.epoch(), epochState -> epochState.complete, mode);
+    }
+
+    public enum EpochSufficiencyMode { AT_LEAST, AT_MOST }
+
+    private Topologies withSufficientEpochs(Unseekables<?> select, long minEpoch, long maxEpoch, Function<EpochState, Ranges> isSufficientFor, EpochSufficiencyMode mode)
```
bdeggleston: afaict, this is assuming that `mode` will be either `AT_LEAST` or `AT_MOST` and never `null`. We should add a null check (or make it a boolean arg)

accord-core/src/main/java/accord/coordinate/ExecuteSyncPoint.java
```diff
@@ -68,17 +77,79 @@ public void start()
         }
     }
 
+    public static class ExecuteExclusiveSyncPoint extends ExecuteSyncPoint<Ranges>
+    {
+        private long retryInFutureEpoch;
+        public ExecuteExclusiveSyncPoint(Node node, SyncPoint<Ranges> syncPoint, Function<Topologies, AbstractSimpleTracker<?>> trackerSupplier)
+        {
+            super(node, syncPoint, trackerSupplier);
```
bdeggleston: can we add an counterpart Invariant like `Invariants.checkArgument(syncPoint.syncId.kind().awaitsOnlyDeps());` or check the txnId kind against the expected value.

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -1249,193 +1198,94 @@ private static ProgressShard progressShard(Route<?> route, @Nullable RoutingKey
...
-        Set
-    }
-
     @SuppressWarnings({"unchecked", "rawtypes"})
-    private static CommonAttributes set(CommonAttributes attrs,
-                                        Ranges existingRanges, Ranges additionalRanges,
-                                        ProgressShard shard, Route<?> route,
-                                        @Nullable PartialTxn partialTxn, EnsureAction ensurePartialTxn,
-                                        @Nullable PartialDeps partialDeps, EnsureAction ensurePartialDeps)
+    private static CommonAttributes set(SaveStatus newStatus, Command cur, CommonAttributes upd,
```
bdeggleston, line 1203: It would be nice if we wouldn't have to include the status we intend to set here, but there doesn't seem to be a way to make the Known dependent stuff happen as part of the state change method without making it both more complex and less understandable

accord-core/src/main/java/accord/coordinate/Infer.java
```diff
@@ -91,18 +96,26 @@ public enum InvalidIfNot
         /**
          * If the command has not had its execution timestamp agreed on any shard
          */
-        IfUndecided(IfQuorum, IfQuorum);
+        IfUndecided(IfQuorum, IfQuorum),
+
+        /**
+         * If the command has not had its execution timestamp agreed on any shard
```
bdeggleston: this comment is identical to `IfUndecided`

accord-core/src/main/java/accord/coordinate/Invalidate.java
```diff
@@ -148,66 +148,82 @@ private void invalidate()
...
+        Route<?> someRoute = InvalidateReply.mergeRoutes(replies);
+
+        // first look to see if it has already been decided/invalidated
+        // check each shard independently - if we find any that can be invalidated, do so
+        InvalidateReply max = InvalidateReply.max(replies);
+        InvalidateReply maxNotTruncated = max;
+        if (max.status == Status.Truncated)
+            maxNotTruncated = InvalidateReply.maxNotTruncated(replies);
+
+        switch (maxNotTruncated.status)
```
bdeggleston: I think you'd need a null check here in case all replies were truncated?

accord-core/src/main/java/accord/coordinate/Invalidate.java
```diff
@@ -148,66 +148,82 @@ private void invalidate()
...
         isPrepareDone = true;
 
-        // first look to see if it has already been
+        FullRoute<?> fullRoute = InvalidateReply.findRoute(replies);
+        Route<?> someRoute = InvalidateReply.mergeRoutes(replies);
+
+        // first look to see if it has already been decided/invalidated
+        // check each shard independently - if we find any that can be invalidated, do so
+        InvalidateReply max = InvalidateReply.max(replies);
+        InvalidateReply maxNotTruncated = max;
```
bdeggleston: nit: could the assignment and conditional be condensed in to a single ternary assignment?

bdeggleston (changes requested):
just some minor stuff

accord-core/src/main/java/accord/coordinate/Invalidate.java
belliottsmith: yep, good catch

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: yep, agreed. I think this is better than it was though, and perhaps we can think about further refinements later.

bdeggleston (approved):
+1

belliottsmith:
thank you for the review! realise I hadn't opened the companion request in C* to fix the serializers, but have just done so

bdeggleston:
this looks to have broken about 23 jvm-dtests and 27 unit tests. Can you take a look at those? Can you also point the PR at the cep-15-accord branch?

belliottsmith:
I've spliced out the `ExclusiveSyncPoint` changes so I can iterate on them and their testing with David. Will merge this and the Cassandra integration later today. Thanks!

PR #118, opened 2024-08-29
cassandra-19437 ignore Update for read transactions

aweisberg:


belliottsmith:
Should we instead reject coordination of a `Txn` with this defined?

aweisberg:
Don't need this change

PR #121, opened 2024-09-17
Cassandra 19869

ifesdjeen:


accord-core/src/main/java/accord/local/cfk/SafeCommandsForKey.java
```diff
@@ -55,12 +55,13 @@ void updatePruned(SafeCommandStore safeStore, Command nextCommand, NotifySink no
...
         CommandsForKey prevCfk = current();
-        update(safeStore, nextCommand, prevCfk, prevCfk.update(nextCommand));
+        CommandsForKeyUpdate cfkUpdate = prevCfk.update(nextCommand);
+        update(safeStore, nextCommand, prevCfk, cfkUpdate);
     }
 
     private void update(SafeCommandStore safeStore, @Nullable Command command, CommandsForKey prevCfk, CommandsForKeyUpdate updateCfk)
     {
-        update(safeStore, command, prevCfk, updateCfk, NotifySink.DefaultNotifySink.INSTANCE);
+        update(safeStore, command, prevCfk, updateCfk, safeStore.replay() ? NotifySink.NoOpSink.INSTANCE : NotifySink.DefaultNotifySink.INSTANCE);
```
belliottsmith: I think we can switch behaviour more cleanly here, but also we need to either capture the notifications OR we need to actually submit them OR we need to avoid updating unmanaged altogether and batch notify them at the end - otherwise they may never be notified.

accord-core/src/main/java/accord/local/Status.java
```diff
@@ -901,7 +901,7 @@ public static Durability nonNullOrMerge(@Nullable Durability a, @Nullable Durabi
 
         public static Durability merge(Durability a, Durability b)
```
belliottsmith: rename or introduce `nonNullOrMerge`?

accord-core/src/main/java/accord/local/Status.java
```diff
@@ -913,7 +913,7 @@ public static Durability merge(Durability a, Durability b)
 
         public static Durability mergeAtLeast(Durability a, Durability b)
```
belliottsmith: rename or introduce `nonNullOrMergeAtLeast`?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -868,20 +903,15 @@ public static Command purge(SafeCommandStore safeStore, SafeCommand safeCommand,
 
             case TRUNCATE:
                 Invariants.checkState(command.saveStatus().compareTo(TruncatedApply) < 0);
-                Invariants.checkState(command.hasBeen(PreApplied));
-                result = truncatedApply(command, Route.tryCastToFullRoute(maybeFullRoute));
+                if (!command.hasBeen(PreCommitted)) result = Command.Truncated.erasedOrInvalidOrVestigial(command);
```
belliottsmith: I think this is a bad merge

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -568,7 +576,25 @@ private static AsyncChain<Void> applyChain(SafeCommandStore safeStore, PreLoadCo
                }));
     }
 
-    private static void apply(SafeCommandStore safeStore, Command.Executed command)
+    public static AsyncChain<Void> applyFromReplay(SafeCommandStore safeStore, PreLoadContext context, TxnId txnId)
```
belliottsmith: unused

accord-core/src/test/java/accord/impl/basic/Journal.java
```diff
@@ -49,15 +64,19 @@
...
+import static accord.local.Status.Truncated;
 import static accord.utils.Invariants.illegalState;
 
-
 public class Journal implements Runnable
 {
+    private static final Logger logger = LoggerFactory.getLogger(Journal.class);
+
     private final Queue<RequestContext> unframedRequests = new ArrayDeque<>();
     private final LongArrayList waitForEpochs = new LongArrayList();
```
belliottsmith: do we need this anymore?

accord-core/src/main/java/accord/local/Commands.java
ifesdjeen: oh yes it is!

accord-core/src/main/java/accord/local/cfk/SafeCommandsForKey.java
ifesdjeen: Made it that we do submit notifications for now. The problem was that notifications were causing submission of new tasks, which was leading to some exceptions. Should be clean by now thanks to your patches.

accord-core/src/main/java/accord/local/Status.java
ifesdjeen: I think we do not need null checks there. Removed.

accord-core/src/main/java/accord/local/Status.java
ifesdjeen: I think we do not need null checks there. Removed.

accord-core/src/test/java/accord/impl/basic/Cluster.java
```diff
@@ -475,11 +513,40 @@ public static Map<MessageType, Stats> run(Id[] nodes, MessageListener messageLis
             }, 5L, SECONDS);
 
             Scheduled reconfigure = sinks.recurring(configRandomizer::maybeUpdateTopology, 1, SECONDS);
+
+            Scheduled purge = sinks.recurring(() -> {
+                trace.debug("Triggering purge.");
```
belliottsmith, line 519: It would be nice to randomise the frequency of these operations (lots of time in tests is now spent on restart, and anyway would be good to change the cycle), and also to randomise the set of nodes we shutdown for this, so that some transactions can carry on being processed (potentially) during this window, and we can exercise more interesting situations.

Or perhaps just randomise each replica independently.

accord-core/src/test/java/accord/impl/basic/Cluster.java
```diff
@@ -219,18 +220,50 @@ public void processAll()
...
+        Pending next = pending.poll();
         if (next == null)
             return false;
 
         processNext(next);
+
         checkFailures.run();
+
+        // If drain was requested, run until we have no non-recurring tasks are available
+        if (next instanceof RecurringPendingRunnable && ((RecurringPendingRunnable) next).drainAfterRunning)
```
belliottsmith: Might it be cleaner to submit a kind of "drain" task?

belliottsmith: by which I mean, instead of making the recurring task say "drain after me", the task submits a one-shot drain task whenever it wants to perform a drain?

belliottsmith: This drain task could itself submit some other post-drain task to e.g. re-enable messaging (and schedule the next journal replay task)

belliottsmith (approved):

accord-core/src/test/java/accord/impl/basic/Cluster.java
ifesdjeen: Added randomization.

PR #125, opened 2024-09-25
CASSANDRA-18675: (Accord): C* stores keyspace in Range which will cause ranges to be removed from Accord when DROP KEYSPACE is performed

dcapwell:
C* PR: https://github.com/apache/cassandra/pull/3574

### DIFF
--- a/accord-core/src/test/java/accord/utils/Property.java
+++ b/accord-core/src/test/java/accord/utils/Property.java
@@ -708,6 +708,8 @@ public class Property
         private final Map<Setup<State, SystemUnderTest>, Integer> knownWeights = new LinkedHashMap<>();
         @Nullable
         private Set<Setup<State, SystemUnderTest>> unknownWeights = null;
+        @Nullable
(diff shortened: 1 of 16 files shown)


accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -708,6 +708,8 @@ public interface Setup<State, SystemUnderTest>
         private final Map<Setup<State, SystemUnderTest>, Integer> knownWeights = new LinkedHashMap<>();
         @Nullable
         private Set<Setup<State, SystemUnderTest>> unknownWeights = null;
+        @Nullable
+        private Map<Predicate<State>, List<Setup<State, SystemUnderTest>>> conditionalCommands = null;
```
dcapwell, line 712: some of the stateful tests were flakey due to too many selections that were told to be `Ignored`... so avoid that by only including them *when* their predicate matches... this optimization was known at the time of writing this class, but going with a simpler one was selected as things were stable...

PR #129, opened 2024-10-02
CASSANDRA-19975: to make MixedTopologyTest stable we need to know the topology to make sure each change is safe

dcapwell:


accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -467,6 +480,7 @@ public <State, SystemUnderTest> void check(Commands<State, SystemUnderTest> comm
                         }
                         commands.destroySut(sut, null);
                         commands.destroyState(state, null);
+                        commands.onSuccess(state, sut, history);
```
dcapwell, line 483: the test now logs the history for each seed, just for extra logging... I kept putting a break point to see this information

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -837,6 +860,18 @@ public CommandsBuilder<State, SystemUnderTest> unknownWeight(Gen.IntGen unknownW
             return this;
         }
 
+        public CommandsBuilder<State, SystemUnderTest> commandsTransformer(BiFunction<State, Gen<Command<State, SystemUnderTest, ?>>, Gen<Command<State, SystemUnderTest, ?>>> commandsTransformer)
```
dcapwell, line 863: this is to move the hidden changes we do and push them into the history.  stuff like changing replication favor is now going to show up in the history

ifesdjeen (approved):

PR #134, opened 2024-11-04
Cassandra 20054

dcapwell:


accord-core/src/main/java/accord/impl/MajorityDepsFetcher.java
```diff
@@ -208,6 +209,8 @@ private void start(Range slice, int nextIndex)
                                             schedule();
                                         }
                                     }
+                                    if (notifyWaiting)
```
dcapwell: > 3) MajorityDepsFetcher would have a deadlock as it triggers waiting notifications while holding the lock, and the waiting callers then access more locks, such as the config service lock

`trySuccess` will trigger all callbacks, which then leads to a deadlock as we never release this lock

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -470,14 +470,14 @@ final Supplier<EpochReady> bootstrapper(Node node, Ranges newRanges, long epoch)
      */
     protected Supplier<EpochReady> sync(Node node, Ranges ranges, long epoch, boolean isLoad)
     {
-        return () -> syncInternal(node, ranges, epoch, isLoad);
+        return () -> syncInternal(node, ranges, epoch, DONE, isLoad);
     }
 
-    protected EpochReady syncInternal(Node node, Ranges ranges, long epoch, boolean isLoad)
+    protected EpochReady syncInternal(Node node, Ranges ranges, long epoch, AsyncResult<Void> metadata, boolean isLoad)
```
dcapwell: > 2) AccordCommandStore submitted sync requests in a blocking manner, but did this on a CommandStore… this lead to a 5 minute deadlock

needed to let this work...

accord-core/src/test/java/accord/utils/Gens.java
```diff
@@ -90,6 +91,63 @@ public static <T> Gen<T> oneOf(Map<Gen<T>, Integer> values)
         return rs -> gen.next(rs).next(rs);
     }
 
+    public static <T> OneOfBuilder<T> oneOf()
```
dcapwell, line 94: > 8) TCM’s getLogEntries method hit an edge case with snapshots where it assumed the API was inclusive, but its exclusive; this caused a gap in epochs

Added this for the log tests as I added property tests there.  This pattern is super common and figured moving out to a builder (similar to the `CommandsBuilder`) would make it easier for users, and leveraged that in the test

accord-core/src/test/java/accord/utils/Gens.java
```diff
@@ -90,6 +91,63 @@ public static <T> Gen<T> oneOf(Map<Gen<T>, Integer> values)
...
+            return this;
+        }
+
+        public OneOfBuilder<T> unknownWeights(Gen.IntGen gen)
+        {
+            this.unknownWeightGen = gen;
+            return this;
+        }
+
+        public Gen<Gen<T>> buildWithDynamicWeights()
```
dcapwell, line 123: this is currently dead code... I didn't want to expand the usage of this class but figured the `CommandsBuilder` would be a good place as it does this pattern

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -445,6 +446,7 @@ public <State, SystemUnderTest> void check(Commands<State, SystemUnderTest> comm
             {
                 State state = null;
                 List<String> history = new ArrayList<>(steps);
+                LongArrayList historyTiming = stepTimeout == null ? null : new LongArrayList();
```
dcapwell, line 449: this is 100% not required, but it was helpful to know how long things were taking... are we relying on retries?  this will help make that clear

accord-core/src/main/java/accord/impl/MajorityDepsFetcher.java
belliottsmith: this code is all defunct now in trunk

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: is this safe?

dcapwell: Right now we run the sync on the command store and block, which will 100% of the time timeout causing the syncs to keep failing. The calling path blocks then returns `DONE` to the caller, so by sequencing the work then overriding the metadata result we should yield the same result.  So if there is a caller that expects the blocking, then that caller is unsafe but could not work right now.

ifesdjeen:
Apart from the ongoing discussions with Benedict, wanted to resync this patch with the goals stated in the jira:

I'm assuming the last two commits are fixing 

> 2) AccordCommandStore submitted sync requests in a blocking manner, but did this on a CommandStore… this lead to a 5 minute deadlock

and 

> 3) MajorityDepsFetcher would have a deadlock as it triggers waiting notifications while holding the lock, and the waiting callers then access more locks, such as the config service lock

Do I understand correctly, that the following are not fixed, but only found using this tooling?
```
1) vtable showing what txn are blocking the queried table would throw error when txn isn’t known, which is valid (report historic transaction…)
4) when restarting and learning about removed nodes, AccordService is not setup yet, so need to pass this through to avoid startup issues
5) When accord asks TCM for the epoch history, there were no retries which would cause stability issues during startup
6) when learning about min epochs needed for startup, purge all starting epochs that are empty as it isn’t needed and only adds costs for startup
7) when nodes leave the cluster we did not start durability sync (this isn’t working, but thats a different issue… durability sync requires ALL which isn’t possible)
8) TCM’s getLogEntries method hit an edge case with snapshots where it assumed the API was inclusive, but its exclusive; this caused a gap in epochs
9) JVM Dtest now supports startup timeouts, this is to avoid issues where startup will take infinity (due to bugs) causing CI to throw away the logs.
10) fixed a race condition bug in Harry where the TokenPlacementModel could see a partial row causing NPEs down the line
11) Fixed a bug in Harry where Accord timeouts would not retry as they don’t have the expected message
```

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: I don’t follow all of that, but I’m not sure that it’s better to make something appear to pass if it’s still broken.

I am not up to speed on all of this, but my intuition is that my proposed fix to AbstractConfigurationService would resolve all of these problems, so wondering if that has been tried yet?

anyway, syncInternal is I think defunct now, so these changes aren’t obviously needed. We should probably revert them and take another look after rebase.

dcapwell:
> Do I understand correctly, that the following are not fixed, but only found using this tooling?

They are all fixed now, they were detected by existing tests and now fixed.

accord-core/src/main/java/accord/local/CommandStore.java
dcapwell: > I am not up to speed on all of this, but my intuition is that my proposed fix to AbstractConfigurationService would resolve all of these problems, so wondering if that has been tried yet?

The problem here is that C* submits a task to the CommandStore that it currently is running in, then blocks on it for 5 minutes... since its a single threaded executor this times out 100% of the time... this doesn't have anything to do with `AbstractConfigurationService` locks

dcapwell:
@ifesdjeen take a look at the C* side of this patch (linked in JIRA) https://github.com/apache/cassandra/pull/3656.  The bugs are fixed in C*, only small changes in Accord to enable testing or fixing.  

As benedict pointed out a trunk change impacted 2 files here, so likely only the testing changes will remain

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: `syncInternal` is not invoked by the command store thread, or shouldn't be. We should add a check to validate that it is not, as if this is happening somehow it breaks assumptions - although I think we no longer depend on this. Either way, the traces I saw showed a calls via `AccordConfigurationService` callbacks, with a callback blocking while holding the ACS lock so that another thread could not make progress (while waiting on ACS).

PR #136, opened 2024-11-07
CASSANDRA-20065: Changed how Config Service locking works to drastically limit the work done while holding the locks

dcapwell:


accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -45,15 +48,16 @@ public abstract class AbstractConfigurationService<EpochState extends AbstractCo
 
     protected final EpochHistory epochs = createEpochHistory();
 
-    protected final List<Listener> listeners = new ArrayList<>();
+    protected final List<Listener> listeners = new CopyOnWriteArrayList<>();
```
dcapwell, line 51: there are now 2 different types of locking in this class: `listeners` that use the copy-on-write, and `epochs` which does sync on the class.

One of the reasons to make this `CopyOnWrite` is that its accessed by different threads (register, caller) and this is a simple way to make sure its safe without needing this class to manage the locks for it (before we did class level sync)

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -74,38 +78,57 @@ public String toString()
...
     public abstract static class AbstractEpochHistory<EpochState extends AbstractEpochState>
     {
         // TODO (low priority): move pendingEpochs / FetchTopology into here?
         private List<EpochState> epochs = new ArrayList<>();
 
-        protected long lastReceived = 0;
-        protected long lastAcknowledged = 0;
+        /**
+         * {@link #lastReceived} and {@link #lastAcknowledged} help define the visibility by making sure that the update is the very last thing done in their respected logic.
+         * This means that the read of these fields can be done without a lock and can be used to tell the caller that the given epoch has reached a given step.
```
dcapwell, line 94: This semantic was to simplify `accord.impl.AbstractConfigurationService#reportTopology(accord.topology.Topology, boolean, boolean)` to avoid making changes.

`lastReceived` and `lastAcknowledged` are accessed to see if we need to enqueue on `epoch - 1` or if we can receive right away.  To avoid extra locking went with a volatile read at the start of the method and receive/ack do the volatile write at the end to publish.  `receive` and `receiveFuture` both lock on the epochs, which makes the area of the lock just those 2 methods

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -168,15 +196,25 @@ Topology topologyFor(long epoch)
             return getOrCreate(epoch).topology;
         }
 
+        Topology topologyForLastReceived()
```
dcapwell, line 200: I did this as I was planning to just do `synchronized`, but now that `lastReceived` can be accessed without locks it wasn't needed... kept it as I still like it but can revert to `epochs.topologyFor(epochs. lastReceived())` if desired

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -204,44 +242,48 @@ public AbstractConfigurationService(Node.Id localId)
...
 
+    @GuardedBy("this")
     private long maxRequestedEpoch;
     @Override
-    public synchronized void fetchTopologyForEpoch(long epoch)
+    public void fetchTopologyForEpoch(long epoch)
     {
-        if (epoch <= maxRequestedEpoch)
-            return;
+        synchronized (this)
```
dcapwell, line 282: this is the only state that collects the config service lock.  Accord uses this lock, so figure this is small enough that the overlap doesn't matter.

Accord mostly uses the lock for disk state

belliottsmith:
I think we need all invocations of `ACS.reportTopology` to be invoked without the monitor, so that the monitor is not held when we invoke the listeners. I think `AbstractConfigurationService.reportTopology` should be partially synchronised, but should release the lock before adding any callbacks to futures that may immediately be invoked, and before invoking any listeners.

belliottsmith:
Also, I think `AccordConfigurationService.start()` should ideally be reworked, as we are holding the lock when we invoke diskStateManager.loadTopologies, which transitively invokes various listeners and reports topologies.

dcapwell:
we spoke in slack, dumping here

access to epoch and epoch state is done with the lock on `accord.impl.AbstractConfigurationService#epochs`
the lock you are referring to is `AccordConfigurationService.this`, which is separate and doesn't block epochs.

`AccordConfigurationService.this` is used to interact with
`AccordConfigurationService#state`
`AccordConfigurationService#diskState`
and `AccordConfigurationService.EpochState#syncStatus` - this one could use the epoch lock, but the access is 100% with diskState

There is some concern that `AccordConfigurationService#start` holds the lock for too long, and I can try to look into this.  The main issue I see is that Accord can notify about epoch status, which requires we save to the topologies table... but we are reading from that table to populate what the state *should* be... This callback can happen (logically) from 2 places: bootstrap complete (CommandStore), `DONE` (local thread, which is either `main` or `MiscStage`).  This would mean that during startup we need to worry about the `CommandStore` and `MiscStage` threads being blocked (both are single threaded).  For `CommandStore` this feels fine as we have not even started journal reply yet, nor are we accepting Accord messages (they are dropped while we are starting up).  That leaves `MiscStage`

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -252,26 +294,16 @@ public synchronized void fetchTopologyForEpoch(long epoch)
...
-        });
+        ready.metadata.addCallback(() -> epochs.acknowledge(ready));
+        ready.fastPath.addCallback(() ->  localSyncComplete(epochs.getOrCreate(ready.epoch).topology, startSync));
     }
 
     protected void topologyUpdatePreListenerNotify(Topology topology) {}
     protected void topologyUpdatePostListenerNotify(Topology topology) {}
 
-    public synchronized void reportTopology(Topology topology, boolean isLoad, boolean startSync)
+    public void reportTopology(Topology topology, boolean isLoad, boolean startSync)
```
belliottsmith, line 304: it slightly bothers me that we can have the same topology registered multiple times. 

But, this appears to already be a problem, although a new category of race conditions open that permit it.

I don't see any obvious way this causes problems besides leaking memory, so perhaps nothing to worry about here.

belliottsmith: Oh, actually this _is_ a problem. We should at least test it again before we invoke `receive`

dcapwell: > Oh, actually this is a problem. We should at least test it again before we invoke receive

We test in `accord.impl.AbstractConfigurationService.AbstractEpochHistory#receive`

```
Invariants.checkState(lastReceived == epoch - 1 || epoch == 0 || lastReceived == 0,
                                      "Epoch %d != %d + 1", epoch, lastReceived);
```

so we will block double adding.

belliottsmith (approved):

PR #137, opened 2024-11-13
Key recovery sees ranges

belliottsmith:
C* side: https://github.com/apache/cassandra/pull/3679

### DIFF
--- a/accord-core/src/main/java/accord/local/KeyHistory.java
+++ b/accord-core/src/main/java/accord/local/KeyHistory.java
@@ -24,6 +24,7 @@ package accord.local;
  */
 public enum KeyHistory
 {
+    NONE,
     // TODO (required): deprecate
     TIMESTAMPS,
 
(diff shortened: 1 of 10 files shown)


accord-core/src/main/java/accord/utils/CheckpointIntervalArray.java
```diff
@@ -216,4 +217,34 @@ else if ((ri & BIT29) != 0)
         forEachRange.accept(p1, p2, p3, p4, start, end);
         return end;
     }
+
+    @Inline
+    public <P1, P2, P3, P4> int forEachKey(Key key, IndexedQuadConsumer<P1, P2, P3, P4> forEachScanOrCheckpoint, IndexedRangeQuadConsumer<P1, P2, P3, P4> forEachRange, P1 p1, P2 p2, P3 p3, P4 p4, int minIndex)
```
dcapwell, line 222: why `Key` vs `RoutableKey`?  I don't directly see `Key` class being used, so curious on the intent

accord-core/src/main/java/accord/local/KeyHistory.java
```diff
@@ -48,7 +49,33 @@ public enum KeyHistory
...
+            default: throw new AssertionError("Unhandled KeyHistory: " + that);
+            case NONE:
+                return true;
+            case RECOVER:
+            case TIMESTAMPS:
+                return this == that;
+            case ASYNC:
+            case INCR:
+            case SYNC:
+                return this.compareTo(ifSyncRequireAtLeast) >= 0;
```
dcapwell, line 77: so for `satisfies` (only method used in this PR) this is only true for `SYNC` and `RECOVER`?  `ASYNC`, `INCR`, and `SYNC` are all new to me, so need to catch up on the recent changes to better understand this change.

accord-core/src/main/java/accord/utils/CheckpointIntervalArray.java
belliottsmith: this is your code I copied forward, isn't it? I didn't look carefully at the signature here. You're right that this looks odd, I'll take a closer look

belliottsmith: It's your generic type parameter

dcapwell: ah thanks, sorry!

accord-core/src/main/java/accord/local/KeyHistory.java
belliottsmith: yes, ASYNC and INCR basically don't guarantee they have fetched what was requested. They'll accept partial information. That said, this isn't implemented on the load side yet so today it's only declarative of intent/requirements on the task side.

dcapwell (approved):
Accord side LGTM, trying to wrap up the C* now

PR #138, opened 2024-11-15
Investigate mixup test

belliottsmith:


accord-core/src/main/java/accord/coordinate/CoordinateEphemeralRead.java
```diff
@@ -136,6 +136,6 @@ void onPreAccepted(Topologies topologies)
     {
         Deps deps = Deps.merge(oks, oks.domainSize(), SortedListMap::getValue, ok -> ok.deps);
         topologies = node.topology().reselect(topologies, QuorumEpochIntersections.preaccept.include, route, executeAtEpoch, executeAtEpoch, Owned);
-        new ExecuteEphemeralRead(node, topologies, route, txnId, txn, executeAtEpoch, deps, this).start();
+        new ExecuteEphemeralRead(node, topologies, route, txnId.withEpoch(executeAtEpoch), txn, executeAtEpoch, deps, this).start();
```
dcapwell, line 139: why this change?  I am not familiar with the ephemeral read execute_at vs txn_id debate, I thought only sync points did such a thing?

accord-core/src/test/java/accord/KeysTest.java
```diff
@@ -225,6 +237,51 @@ public void slice()
...
+                K[] keys = arrayFactory.apply(rs.nextInt(2, maxSize));
+                IntHashSet used = new IntHashSet();
+                for (int i = 0 ; i < keys.length ; ++i)
+                {
+                    int k = rs.nextInt(maxKey);
+                    while (!used.add(k)) k = rs.nextInt(maxKey);
+                    keys[i] = keyFactory.apply(k);
+                }
+                int addTo = rs.nextInt(1, keys.length);
+                int removeFrom = addTo - rs.nextInt(addTo);
```
dcapwell:
```suggestion
                int removeFrom = addTo - rs.nextInt(addTo + 1);
```

this test doesn't cover the case that the remove set is fully inclusive of the add set, so never reaches the empty set.  By doing `addTo + 1` we allow `removeFrom` to reach `0` which would test the empty set as well.

dcapwell (approved):

accord-core/src/main/java/accord/coordinate/CoordinateEphemeralRead.java
belliottsmith: ephemeral reads chase an epoch they can validly execute in. For simplicity, we define the txnId = executeAt, and so to execute in a future epoch we simply make txnId.epoch == executeAt.epoch. This change was already made, but somehow this codepath lost(?) the update.

PR #141, opened 2024-11-20
Fix infinite loop and notify progress log of sync point durability while waiting to apply

belliottsmith:
C* side: https://github.com/apache/cassandra/pull/3698

### DIFF
--- a/accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
+++ b/accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
@@ -24,7 +24,7 @@ import javax.annotation.Nullable;
 
 import accord.api.ProtocolModifiers;
 import accord.api.Result;
-import accord.coordinate.ExecuteSyncPoint.ExecuteBlocking;
+import accord.coordinate.ExecuteSyncPoint.ExecuteInclusive;
 import accord.coordinate.tracking.FastPathTracker;
 import accord.coordinate.tracking.PreAcceptTracker;
 import accord.coordinate.tracking.QuorumTracker;
@@ -71,6 +71,11 @@ public interface CoordinationAdapter<R>
         @Override
         public <R> CoordinationAdapter<R> get(TxnId txnId, Step step)
         {
+            switch (txnId.kind())
+            {
+                case ExclusiveSyncPoint: return (CoordinationAdapter<R>) Adapters.exclusiveSyncPoint();
+                case SyncPoint: return (CoordinationAdapter<R>) Adapters.inclusiveSyncPoint();
+            }
             switch (step)
             {
                 default: throw new AssertionError("Unhandled step: " + step);
@@ -219,18 +224,17 @@ public interface CoordinationAdapter<R>
             @Override
             public void execute(Node node, Topologies any, FullRoute<?> route, ExecutePath path, TxnId txnId, Txn txn, Timestamp executeAt, Deps deps, BiConsumer<? super SyncPoint<U>, Throwable> callback)
             {
-                Topologies all = forExecution(node, route, txnId, executeAt, deps);
-                persist(node, all, route, txnId, txn, executeAt, deps, txn.execute(txnId, executeAt, null), txn.result(txnId, executeAt, null), callback);
+                persist(node, null, route, txnId, txn, executeAt, deps, txn.execute(txnId, executeAt, null), txn.result(txnId, executeAt, null), callback);
             }
 
             @Override
-            public void persist(Node node, Topologies any, FullRoute<?> route, Route<?> participants, TxnId txnId, Txn txn, Timestamp executeAt, Deps deps, Writes writes, Result result, BiConsumer<? super SyncPoint<U>, Throwable> callback)
+            public void persist(Node node, Topologies ignore, FullRoute<?> route, Route<?> participants, TxnId txnId, Txn txn, Timestamp executeAt, Deps deps, Writes writes, Result result, BiConsumer<? super SyncPoint<U>, Throwable> callback)
             {
                 Topologies all = forExecution(node, route, txnId, executeAt, deps);
 
                 invokeSuccess(node, route, txnId, txn, deps, callback);
                 new PersistSyncPoint(node, all, txnId, route, txn, executeAt, deps, writes, result)
-                    .start(Apply.FACTORY, Maximal, any, writes, result);
+                .start(Apply.FACTORY, Maximal, all, writes, result);
             }
         }
 
@@ -283,18 +287,6 @@ public interface CoordinationAdapter<R>
             {
                 return node.topology().preciseEpochs(route, txnId.epoch(), executeAt.epoch());
             }
-
-            @Override
-            public void execute(Node node, Topologies any, FullRoute<?> route, ExecutePath path, TxnId txnId, Txn txn, Timestamp executeAt, Deps deps, BiConsumer<? super SyncPoint<U>, Throwable> callback)
-            {
-                Topologies all = forExecution(node, route, txnId, executeAt, deps);
-
-                ExecuteBlocking<U> execute = ExecuteBlocking.atQuorum(node, all, new SyncPoint<>(txnId, deps, (FullRoute<U>) route), executeAt);
-                execute.start();
-                addOrExecuteCallback(execute, callback);
-            }
-
-            protected abstract void addOrExecuteCallback(ExecuteBlocking<U> execute, BiConsumer<? super SyncPoint<U>, Throwable> callback);
         }
 
         /*
@@ -310,14 +302,6 @@ public interface CoordinationAdapter<R>
             protected AsyncInclusiveSyncPointAdapter() {
                 super();
             }
-
-            @Override
-            protected void addOrExecuteCallback(ExecuteBlocking<U> execute, BiConsumer<? super SyncPoint<U>, Throwable> callback)
-            {
-                // If this is the async adapter then we want to invoke the callback immediately
-                // and the caller can wait on the txn locally if they want
-                callback.accept(execute.syncPoint, null);
-            }
         }
 
         private static class InclusiveSyncPointBlockingAdapter<U extends Unseekable> extends AbstractInclusiveSyncPointAdapter<U>
@@ -329,9 +313,13 @@ public interface CoordinationAdapter<R>
             }
 
             @Override
-            protected void addOrExecuteCallback(ExecuteBlocking<U> execute, BiConsumer<? super SyncPoint<U>, Throwable> callback)
+            public void execute(Node node, Topologies any, FullRoute<?> route, ExecutePath path, TxnId txnId, Txn txn, Timestamp executeAt, Deps deps, BiConsumer<? super SyncPoint<U>, Throwable> callback)
             {
+                Topologies all = forExecution(node, route, txnId, executeAt, deps);
+
+                ExecuteInclusive<U> execute = ExecuteInclusive.atQuorum(node, all, new SyncPoint<>(txnId, deps, (FullRoute<U>) route), executeAt);
                 execute.addCallback(callback);
+                execute.start();
             }
 
             @Override
--- a/accord-core/src/main/java/accord/coordinate/ExecuteSyncPoint.java
+++ b/accord-core/src/main/java/accord/coordinate/ExecuteSyncPoint.java
@@ -24,17 +24,22 @@ import java.util.function.Function;
 
 import accord.api.Result;
 import accord.coordinate.CoordinationAdapter.Adapters;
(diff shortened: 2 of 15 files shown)


accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -430,8 +430,10 @@ protected Cancellable start(BiConsumer<? super V, Throwable> callback)
 
     // either the thing we start, or the thing we do in follow-up
     BiConsumer<? super V, Throwable> next;
+    private final Head<?> head;
```
dcapwell: you can revert this, I did this to aid in debugger support

accord-core/src/main/java/accord/messages/InformDurable.java
```diff
@@ -63,6 +67,22 @@ private InformDurable(TxnId txnId, Route<?> scope, long waitForEpoch, @Nullable
...
+        int homeShardIndex = homeEpochTopology.indexForKey(route.homeKey());
+        if (homeShardIndex < 0)
+        {
+            homeEpochTopology = node.topology().globalForEpoch(homeEpoch);
+            homeShardIndex = homeEpochTopology.indexForKey(route.homeKey());
+        }
+
+        Shard homeShard = homeEpochTopology.get(homeShardIndex);
+        Topologies homeTopology = new Topologies.Single(any, new Topology(homeEpoch, homeShard));
+        node.send(homeShard.nodes, to -> new InformDurable(to, homeTopology, route.homeKeyOnlyRoute(), txnId, executeAt, Majority));
```
dcapwell: should `Majority` be hard coded or should it be passed as a param?  What about changing the method name to `informHomeMajority`

dcapwell: in `MaybeRecover` we call `accord.primitives.Status.Durability#isDurable` which is

```
public boolean isDurable()
        {
            return this == Majority || this == Universal;
        }
```

so based off what we know, it "could" be `Universal` as well?

accord-core/src/main/java/accord/coordinate/Persist.java
```diff
@@ -86,12 +84,7 @@ public void onSuccess(Id from, ApplyReply reply)
                     if (!isDone)
                     {
                         isDone = true;
-                        Topologies topologies = tracker.topologies();
-                        Topology topology = topologies.forEpoch(txnId.epoch());
-                        int homeShardIndex = topology.indexForKey(route.homeKey());
-                        // we can persist only partially if some shards are already completed; in this case the home shard may not participate
-                        if (homeShardIndex >= 0)
-                            node.send(topology.get(homeShardIndex).nodes, to -> new InformDurable(to, topologies, route, txnId, executeAt, Majority));
+                        InformDurable.informHome(node, topologies, txnId, route, executeAt);
```
dcapwell: we will now send as we fall back to the global topology yet the previous comment implies this isn't safe?

accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
```diff
@@ -219,18 +224,17 @@ public void stabilise(Node node, Topologies any, FullRoute<?> route, Ballot ball
...
             @Override
-            public void persist(Node node, Topologies any, FullRoute<?> route, Participants<?> participants, TxnId txnId, Txn txn, Timestamp executeAt, Deps deps, Writes writes, Result result, BiConsumer<? super SyncPoint<U>, Throwable> callback)
+            public void persist(Node node, Topologies ignore, FullRoute<?> route, Participants<?> participants, TxnId txnId, Txn txn, Timestamp executeAt, Deps deps, Writes writes, Result result, BiConsumer<? super SyncPoint<U>, Throwable> callback)
             {
                 Topologies all = forExecution(node, route, txnId, executeAt, deps);
 
                 invokeSuccess(node, route, txnId, txn, deps, callback);
                 new PersistSyncPoint(node, all, txnId, route, txn, executeAt, deps, writes, result)
-                    .start(Apply.FACTORY, Maximal, any, writes, result);
+                .start(Apply.FACTORY, Maximal, all, writes, result);
```
dcapwell, line 237: for me: lgtm

accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
```diff
@@ -219,18 +224,17 @@ public void stabilise(Node node, Topologies any, FullRoute<?> route, Ballot ball
             @Override
             public void execute(Node node, Topologies any, FullRoute<?> route, ExecutePath path, TxnId txnId, Txn txn, Timestamp executeAt, Deps deps, BiConsumer<? super SyncPoint<U>, Throwable> callback)
             {
-                Topologies all = forExecution(node, route, txnId, executeAt, deps);
-                persist(node, all, route, txnId, txn, executeAt, deps, txn.execute(txnId, executeAt, null), txn.result(txnId, executeAt, null), callback);
+                persist(node, null, route, txnId, txn, executeAt, deps, txn.execute(txnId, executeAt, null), txn.result(txnId, executeAt, null), callback);
```
dcapwell, line 227: just double checking, this is because every impl recalculates `all`, so just avoid double effort?

accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
```diff
@@ -310,14 +302,6 @@ private static class AsyncInclusiveSyncPointAdapter<U extends Unseekable> extend
             protected AsyncInclusiveSyncPointAdapter() {
                 super();
             }
-
-            @Override
-            protected void addOrExecuteCallback(ExecuteBlocking<U> execute, BiConsumer<? super SyncPoint<U>, Throwable> callback)
-            {
-                // If this is the async adapter then we want to invoke the callback immediately
```
dcapwell, line 317: what is the reason to make this async rather than immediate?

accord-core/src/main/java/accord/coordinate/ExecuteSyncPoint.java
```diff
@@ -195,6 +257,11 @@ protected void onSuccess()
         trySuccess(syncPoint);
     }
 
+    protected void sendApply(Node.Id to)
```
dcapwell, line 267: this appears to be dead code, is this used by C*?

accord-core/src/main/java/accord/coordinate/MaybeRecover.java
```diff
@@ -108,7 +108,7 @@ protected void onDone(Success success, Throwable fail)
                     if (hasMadeProgress(full))
                     {
                         if (full.durability.isDurable())
-                            node.send(topologies.forEpoch(txnId.epoch()).forKey(route.homeKey()).nodes, to -> new InformDurable(to, topologies, route, txnId, full.executeAtIfKnown(), full.durability));
+                            InformDurable.informHome(node, topologies, txnId, route, full.executeAtIfKnown());
```
dcapwell: commented in `InformDurable.informHome` but durability "could" be `Universal` yet this method hard codes `Majority`

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -513,10 +513,17 @@ public enum CommitOrReadNack implements ReadData.ReadReply
...
+
         /**
          * Either not committed, or not stable
          */
         Insufficient("CommitInsufficient"),
+
+        /**
+         * PreApplied successfully, but the request is blocking so waiting to reply
+         */
+        Waiting("ApplyWaiting"),
```
dcapwell, line 532: I don't see this sent in this patch, is this done in C*?

accord-core/src/main/java/accord/coordinate/ExecuteSyncPoint.java
dcapwell: I don't see it used in https://github.com/apache/cassandra/pull/3698

accord-core/src/main/java/accord/messages/ReadData.java
dcapwell: I don't see it used in https://github.com/apache/cassandra/pull/3698

accord-core/src/main/java/accord/utils/async/AsyncChains.java
belliottsmith: ah, I'll rebase against trunk

accord-core/src/main/java/accord/messages/InformDurable.java
belliottsmith: I guess `durable` is a slightly overloaded concept. But, something is _durable_ once it is known to a majority. Basically it means can survive our tolerated number of failures. We have higher levels of durability, and confusingly we also have `Local` durability (that perhaps we should remove, since it isn't _durable_). Anyway, I'm happy to parameterise it, but `Majority` is the base level at which you can call something _durable_, so if unspecified should normally be assumed.

accord-core/src/main/java/accord/coordinate/Persist.java
belliottsmith: it doesn't say it's unsafe, it says we might not have the information in our topologies. It's fine to inform the home shard (but perhaps redundant). I thought it was simpler to merge the different callers, but happy to split them again and introduce a `tryInformHome` (I did have this initially)

accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
belliottsmith: yeah, could do with some more cleaning up really but just noticed this redundancy here

accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
belliottsmith: I just properly separated the logic for async/blocking, since we shouldn't be executing a blocking coordination act all for async, rather than simply sharing everything except deciding in `addOrExecuteCallback` whether a/sync

accord-core/src/main/java/accord/coordinate/ExecuteSyncPoint.java
belliottsmith: probably my mistake, I meant to extract and call it directly. will fix

accord-core/src/main/java/accord/coordinate/MaybeRecover.java
belliottsmith: I'll parameterise, but as mentioned in other location if unspecified the lowest level of _durable_ durability is reasonable to assume (should perhaps remove `Local` durability to make it clearer this is the lowest _durable_ durability).

accord-core/src/main/java/accord/messages/ReadData.java
belliottsmith: hmm, let me see where it went - it should be here as part of the new sync point logic.

accord-core/src/main/java/accord/utils/async/AsyncChains.java
dcapwell: its not in trunk.  The patch I sent you I needed to know the state of the `allOf` to see which shard was not replying, so doing this let me get access.  This was debugging code

accord-core/src/main/java/accord/coordinate/Persist.java
dcapwell: its fine, I just wanted to get your feedback on it.  Redundant is fine by me, just making sure things were intentional (I have caused a lot of bugs due to refactoring missing subtle details)

accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
dcapwell: ok ill need to look closer, I missed that detail

accord-core/src/main/java/accord/messages/ReadData.java
belliottsmith: It's used in `ApplyThenWaitUntilApplied`

accord-core/src/main/java/accord/coordinate/ExecuteSyncPoint.java
belliottsmith: yep, forgot to invoke it instead of CoordinateSyncPoint.sendApply

accord-core/src/main/java/accord/coordinate/Persist.java
```diff
@@ -83,18 +78,12 @@ public void onSuccess(Id from, ApplyReply reply)
             default: throw new IllegalStateException();
             case Redundant:
             case Applied:
-                persistedOn.add(from);
```
dcapwell, line 86: for me: confirmed this is existing dead code, so this patch just cleans it up

dcapwell (approved):

PR #144, opened 2024-11-27
Expose Journal-related files to allow external Journal implementation integration.

ifesdjeen:
CASSANDRA-20112


accord-core/src/main/java/accord/impl/AbstractLoader.java
```diff
@@ -0,0 +1,77 @@
...
+                case TRUNCATE_WITH_OUTCOME:
+                case TRUNCATE:
+                case ERASE:
+                    command = Commands.purge(command, command.participants(), cleanup);
+            }
+        }
+
+        command = safeStore.unsafeGet(txnId).update(safeStore, command);
+        if (command.status() == Truncated)
+            safeStore.progressLog().clear(command.txnId());
```
belliottsmith: is this something we need for both implementations? I'm not sure, it looks like something hacky we would do for the in memory implementation, that perhaps should be done more cleanly (perhaps tied to `Cleanup` invocation). But also perhaps this is fine.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -1377,99 +1375,86 @@ public void clearForTesting()
...
 
+        @Override
+        public void load(Command command, Journal.OnDone onDone)
+        {
+            try
+            {
+                commandStore.executeInContext(commandStore,
+                                              context(command, ASYNC),
+                                              safeStore -> loadInternal(command, safeStore));
+                onDone.success();
```
belliottsmith: not super important, but perhaps this should be invoked outside of the try block so we don't invoke onDone.failure(t) if success callback fails

accord-core/src/test/java/accord/burn/BurnTest.java
```diff
@@ -18,675 +18,13 @@
...
-
-    private static Verifier createVerifier(int keyCount)
-    {
-        if (!ElleVerifier.Support.allowed())
-            return new StrictSerializabilityVerifier("", keyCount);
-        return CompositeVerifier.create(new StrictSerializabilityVerifier("", keyCount),
-                                        new ElleVerifier());
-    }
-
-    public static void main(String[] args)
```
belliottsmith, line 531: Should the main method stay here perhaps?

accord-core/src/main/java/accord/impl/AbstractLoader.java
ifesdjeen: Double checking.

accord-core/src/test/java/accord/impl/basic/Cluster.java
```diff
@@ -778,7 +780,7 @@ private static void verifyConsistentRestore(Int2ObjectHashMap<NavigableMap<TxnId
                     if (!store.unsafeCommands().containsKey(txnId))
                     {
                         Command beforeCommand = before.get(txnId);
-                        if (beforeCommand.saveStatus() == SaveStatus.Erased)
```
belliottsmith: Should we permit this? I don't think we should even *have* any commands where `saveStatus` is `Uninitialised` - might indicate some other problem? If the command is `Uninitialised` it means we didn't touch it, so we shouldn't have saved it.

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
```diff
@@ -153,7 +155,8 @@ public void validateRead(Command current)
                 return;
 
             // Journal will not have result persisted. This part is here for test purposes and ensuring that we have strict object equality.
-            Command reconstructed = journal.reconstruct(id, current.txnId());
+            // TODO: redundant/durable before
```
belliottsmith: This TODO looks important

accord-core/src/test/java/accord/impl/mock/MockConfigurationService.java
```diff
@@ -18,27 +18,27 @@
...
-import org.junit.jupiter.api.Assertions;
-
 import java.util.*;
 
 public class MockConfigurationService implements TestableConfigurationService
 {
     private final MessageSink messageSink;
     private final List<Topology> epochs = new ArrayList<>();
-    private final Map<Long, EpochReady> acks = new HashMap<>();
+    private final Map<Long, ConfigurationService.EpochReady> acks = new HashMap<>();
```
belliottsmith: static import class?

accord-core/src/test/java/accord/impl/mock/MockConfigurationService.java
```diff
@@ -18,27 +18,27 @@
...
     private final List<AsyncResult<Void>> syncs = new ArrayList<>();
-    private final List<Listener> listeners = new ArrayList<>();
+    private final List<ConfigurationService.Listener> listeners = new ArrayList<>();
```
belliottsmith: static import class?

belliottsmith (approved):
Left some minor feedback, otherwise LGTM +1

accord-core/src/test/java/accord/impl/mock/MockConfigurationService.java
ifesdjeen: I will double check on commit / when sequashing, I think these are artifact of moving files back and forth.

accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
ifesdjeen: Fixed this, you are right, should not have left it for TODO.

accord-core/src/test/java/accord/burn/BurnTest.java
ifesdjeen: Moved back.

accord-core/src/test/java/accord/impl/mock/MockConfigurationService.java
ifesdjeen: Reverted all changes that are not related directly to the patch, including this one.

accord-core/src/test/java/accord/impl/basic/Cluster.java
ifesdjeen: Agreed, this should not be the case; fixed now

accord-core/src/main/java/accord/impl/AbstractLoader.java
ifesdjeen: Removed and ran several seeds with 10k operations, which are passing.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
ifesdjeen: did that in both methods

ifesdjeen:
Committed manually

PR #147, opened 2024-11-29
Sorted arrays tests sometimes falls into infinite re-generate loop

ifesdjeen:
This only does the simplest possible thing: limits the number of cycles to 10K. We can improve this in the future by introducing a variable-entropy zero-cycle Integer generators. I have made a sketch, but I'm not comfortable releasing it yet, needs a bit more thought.


belliottsmith (approved):

ifesdjeen:
Committed manually: https://github.com/apache/cassandra-accord/commit/356fee54d56e1349465fddd20653129338f9ad7b

PR #150, opened 2024-12-12
Fix tests 3

ifesdjeen:


accord-core/src/main/java/accord/coordinate/CollectLatestDeps.java
```diff
@@ -0,0 +1,126 @@
...
+import static accord.coordinate.tracking.RequestStatus.Success;
+
+public class CollectLatestDeps implements Callback<GetLatestDepsOk>
+{
+    final Node node;
+    final TxnId txnId;
+    final RoutingKey homeKey;
+    final Timestamp executeAt;
+
+    private final SortedListMap<Id, GetLatestDepsOk> oks;
```
ifesdjeen, line 52: could you elaborate why it is important for this to be sorted?

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
```diff
@@ -104,7 +106,7 @@ private Commit.Kind commitKind()
     @Override
     public void contact(Id to)
     {
-        if (SEND_MINIMUM_STABLE_MESSAGES) Commit.stableAndRead(to, node, allTopologies, commitKind(), txnId, txn, route, readScope, executeAt, stableDeps, this, SEND_MINIMUM_STABLE_MESSAGES);
+        if (SEND_MINIMUM_STABLE_MESSAGES && path != RECOVER) Commit.stableAndRead(to, node, allTopologies, commitKind(), txnId, txn, route, readScope, executeAt, stableDeps, this, false);
```
ifesdjeen: Is it intended that we e passing `false` here? Above we seem to be passing `true` under same condition (send minimum and path not recover).

accord-core/src/main/java/accord/coordinate/tracking/QuorumIdTracker.java
```diff
@@ -0,0 +1,122 @@
...
+import accord.local.Node;
+import accord.topology.Shard;
+import accord.topology.Topologies;
+import org.agrona.collections.ObjectHashSet;
+
+import static accord.coordinate.tracking.AbstractTracker.ShardOutcomes.Fail;
+import static accord.coordinate.tracking.AbstractTracker.ShardOutcomes.NoChange;
+import static accord.coordinate.tracking.AbstractTracker.ShardOutcomes.Success;
+
+public class QuorumIdTracker extends SimpleTracker<QuorumIdTracker.QuorumIdShardTracker> implements ResponseTracker
```
ifesdjeen: Should we maybe always use `QuorumIdTracker`? I have been in a situation recently, where I had to reconstruct IDs of successes and failures.

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -289,13 +289,16 @@ public void fetchTopologyForEpoch(long epoch)
         fetchTopologyInternal(epoch);
     }
 
+    // TODO (expected): rename, sync is too ambiguous
```
ifesdjeen, line 292: Oh yes please. I was looking into whether we need to move this into Listener because of this name, too.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -331,6 +332,17 @@ protected void updatedRedundantBefore(SafeCommandStore safeStore, TxnId syncId,
                 }
             });
         });
+        TxnId clearProgressLogBefore = unsafeGetRedundantBefore().minShardRedundantBefore();
+        List<TxnId> clearing = ((DefaultProgressLog) progressLog).activeBefore(clearProgressLogBefore);
+        for (TxnId txnId : clearing)
+        {
+            GlobalCommand globalCommand = commands.get(txnId);
+;            Invariants.checkState(globalCommand != null && !globalCommand.isEmpty());
```
ifesdjeen, line 295: nit: `;`

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -479,11 +479,12 @@ final Supplier<EpochReady> bootstrapper(Node node, Ranges newRanges, long epoch)
                 Bootstrap bootstrap = new Bootstrap(node, this, epoch, newRanges);
                 bootstraps.add(bootstrap);
                 bootstrap.start(safeStore);
-                return new EpochReady(epoch, null, bootstrap.coordination, bootstrap.data, bootstrap.reads);
+                return new EpochReady(epoch, null, null, bootstrap.data, bootstrap.reads);
```
ifesdjeen, line 483: I might be missing something, but if the node can process reads, it should be ready for coordination, too or?

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -498,15 +499,21 @@ final Supplier<EpochReady> bootstrapper(Node node, Ranges newRanges, long epoch)
...
     {
         return () -> {
-            if (redundantBefore.min(ranges, RedundantBefore.Entry::locallyWitnessedBefore).epoch() >= epoch)
-                return new EpochReady(epoch, DONE, DONE, DONE, DONE);
-
-            AsyncResults.SettableResult<Void> whenDone = new AsyncResults.SettableResult<>();
-            waitingOnSync.put(epoch, new WaitingOnSync(whenDone, ranges));
-            return new EpochReady(epoch, DONE, whenDone, DONE, DONE);
+            AsyncResult<Void> readyToCoordinate = readyToCoordinate(ranges, epoch);
+            return new EpochReady(epoch, DONE, readyToCoordinate, DONE, DONE);
```
ifesdjeen, line 504: hm, looks like I was wrong above (can serve reads -> can coordinate), so it would be nice to have this documented somewhere.

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -91,7 +102,7 @@ private Commands()
     {
     }
 
-    public enum AcceptOutcome { Success, Redundant, RejectedBallot, Truncated }
+    public enum AcceptOutcome { Success, Redundant, RejectedBallot, Retired, Truncated }
```
ifesdjeen, line 104: Is this a correctness issue or optimizaiton? In either case would be great to document.

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -107,11 +118,11 @@ public static AcceptOutcome recover(SafeCommandStore safeStore, SafeCommand safe
     private static AcceptOutcome preacceptOrRecover(SafeCommandStore safeStore, SafeCommand safeCommand, StoreParticipants participants, TxnId txnId, long acceptEpoch, PartialTxn partialTxn, FullRoute<?> route, Ballot ballot)
     {
         final Command command = safeCommand.current();
-
         if (command.hasBeen(Truncated))
         {
             logger.trace("{}: skipping preaccept - command is truncated", txnId);
-            return command.is(Invalidated) ? AcceptOutcome.RejectedBallot : AcceptOutcome.Truncated;
+            return command.is(Invalidated) ? AcceptOutcome.RejectedBallot : participants.owns().isEmpty()
```
ifesdjeen, line 123: nit: would you mind to expand double ternary?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -598,9 +582,11 @@ public static boolean maybeExecute(SafeCommandStore safeStore, SafeCommand safeC
         {
             case Stable:
                 // TODO (desirable, efficiency): maintain distinct ReadyToRead and ReadyToWrite states
-                // TODO (required): we can have dangling transactions in some cases when proposing in a future epoch but
-                //   later deciding on an earlier epoch. We should probably turn this into an erased vestigial command,
-                //   but we should tighten up our semantics there in general.
+                // TODO (required): we can have dangling transactions if we don't execute them,
+                //  but we don't want to erase the transaction until
```
ifesdjeen, line 589: Sentence seems to be finished abruptly?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -718,7 +706,7 @@ else if (safeStore.isFullyPreBootstrapOrStale(dependency, waiting.partialDeps().
     static void updateDependencyAndMaybeExecute(SafeCommandStore safeStore, SafeCommand safeCommand, SafeCommand predecessor, boolean notifyWaitingOn)
     {
         Command.Committed command = safeCommand.current().asCommitted();
-        if (command.hasBeen(Applied))
+        if (command.hasBeen(Status.Applied))
```
ifesdjeen: nit: `Status` added by accident?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -1162,41 +1246,27 @@ static Command removeNoLongerOwnedDependency(SafeCommandStore safeStore, SafeCom
...
 
-    public static Command updateParticipants(SafeCommandStore safeStore, SafeCommand safeCommand, StoreParticipants participants)
+    public static Command supplementParticipants(SafeCommandStore safeStore, SafeCommand safeCommand, StoreParticipants participants)
     {
         Command current = safeCommand.current();
         if (current.saveStatus().compareTo(Erased) >= 0)
             return current;
-
-        CommonAttributes updated = updateParticipants(current, participants);
+        // TODO (required): invoke updateParticipants directly
```
ifesdjeen: nit: supplement?

accord-core/src/main/java/accord/local/RedundantBefore.java
```diff
@@ -161,8 +185,7 @@ public Entry(Range range, long startOwnershipEpoch, long endOwnershipEpoch, @Non
             this.bootstrappedAt = bootstrappedAt;
             this.staleUntilAtLeast = staleUntilAtLeast;
             checkNoneOrRX(locallyWitnessedOrInvalidatedBefore, locallyAppliedOrInvalidatedBefore, locallyDecidedAndAppliedOrInvalidatedBefore,
-                          shardAppliedOrInvalidatedBefore, gcBefore);
-            checkPartiallyOrdered(locallyDecidedAndAppliedOrInvalidatedBefore, locallyAppliedOrInvalidatedBefore);
+                          shardAppliedOrInvalidatedBefore, gcBefore);checkPartiallyOrdered(locallyDecidedAndAppliedOrInvalidatedBefore, locallyAppliedOrInvalidatedBefore);
```
ifesdjeen: nit: line break here?

accord-core/src/main/java/accord/local/RedundantBefore.java
```diff
@@ -712,37 +857,95 @@ public Ranges removePreBootstrap(TxnId txnId, Ranges ranges)
...
+        if (txnId.is(ExclusiveSyncPoint))
+        {
+            if (!mayFilterStaleOrPreBootstrapOrRetired(txnId, participants))
+                return participants;
+
+            return foldl(participants, Entry::withoutStaleOrPreBootstrapOrLocallyRetired, participants, txnId, i -> false);
+        }
+        else
+        {
+            if (!mayFilterStaleOrPreBootstrap(txnId, participants))
```
ifesdjeen, line 918: nit: do I understand right that we can get here only for non-exclusive syncpoints?

accord-core/src/main/java/accord/local/RedundantBefore.java
```diff
@@ -712,37 +857,95 @@ public Ranges removePreBootstrap(TxnId txnId, Ranges ranges)
...
+                return participants;
+
+            return foldl(participants, Entry::participantsWithoutStaleOrPreBootstrap, participants, txnId, executeAt, r -> false);
+        }
+    }
+
+    public boolean mayFilter(TxnId txnId, Participants<?> participants)
```
ifesdjeen, line 925: nit: why a shortcut name for this one?

accord-core/src/main/java/accord/local/RedundantStatus.java
```diff
@@ -80,6 +92,26 @@ public enum RedundantStatus
      */
     LOCALLY_REDUNDANT,
 
+    /**
+     * The relevant owned ranges are all shard-redundant AND (pre-bootstrap or stale), meaning any intersecting transaction
```
ifesdjeen, line 96: Are these new states required for correctness?

accord-core/src/main/java/accord/local/cfk/CommandsForKey.java
```diff
@@ -740,7 +753,7 @@ public enum InternalStatus
         STABLE(true, false),
         APPLIED(true, false),
         INVALIDATED(false, false),
-        TRUNCATED_OR_PRUNED(false, false)
```
ifesdjeen, line 743: could you elaborate a difference between truncated and pruned here?

accord-core/src/main/java/accord/local/cfk/PostProcess.java
```diff
@@ -275,30 +275,54 @@ static NotifyUnmanagedResult notifyUnmanaged(Unmanaged[] unmanageds,
...
+                while (end < unmanageds.length && j < committedByExecuteAt.length)
+                {
+                    int c = committedByExecuteAt[j].executeAt.compareTo(unmanageds[end].waitingUntil);
+                    if (c == 0)
+                    {
+                        if (start != end)
+                        {
+                            TxnId[] notifyNotWaiting = selectUnmanaged(unmanageds, start, end);
+                            unmanageds = removeUnmanaged(unmanageds, start, end);
+                            end -= (end - start);
```
ifesdjeen: Had to re-read this several times. Isn't this just `end = start` ?

accord-core/src/main/java/accord/local/cfk/Pruning.java
```diff
@@ -393,23 +393,23 @@ private static Long2ObjectHashMap<TxnInfo> buildEpochPrunedBefores(TxnInfo[] byI
             return null;
 
         Long2ObjectHashMap<TxnInfo> epochPrunedBefores = new Long2ObjectHashMap<>();
-        for (long epoch = newPrunedBefore.epoch() ; epoch <= newPrunedBefore.executeAt.epoch(); ++epoch)
-            epochPrunedBefores.put(epoch, newPrunedBefore);
+        epochPrunedBefores.put(newPrunedBefore.epoch(), newPrunedBefore);
 
-        int maxi = Arrays.binarySearch(committedByExecuteAt, newPrunedBefore, TxnInfo::compareExecuteAt);
+        int maxi = -1 - Arrays.binarySearch(committedByExecuteAt, newPrunedBefore, (a, b) -> a.compareExecuteAtEpoch(b) >= 0 ? 1 : -1);
```
ifesdjeen, line 460: do we need to check for < 0 here?

accord-core/src/main/java/accord/messages/AbstractRequest.java
```diff
@@ -81,7 +81,7 @@ public final void process(Node on, Node.Id replyTo, ReplyContext replyContext)
         {
             long expiresAt = node.agent().expiresAt(replyContext, MICROSECONDS);
             if (expiresAt > 0)
-            {
+            {   // TODO (required): now!! this should be registerAt (not changing to avoid messing up seed)
```
ifesdjeen: Reminding of this one

accord-core/src/main/java/accord/messages/Commit.java
```diff
@@ -173,64 +174,65 @@ public static void stableAndRead(Node node, Topologies all, Kind kind, TxnId txn
...
         Invariants.checkState(all.oldestEpoch() == txnId.epoch());
         Invariants.checkState(all.currentEpoch() == executeAt.epoch());
-        Topology executes = all.forEpoch(executeAt.epoch());
-        Topology coordinates = all.forEpoch(txnId.epoch());
+        Topology executes = all.getEpoch(executeAt.epoch());
+        Topology coordinates = all.getEpoch(txnId.epoch());
 
         SortedArrayList<Id> contact = all.nodes().without(all::isFaulty);
-        sendTo(contact, readSet, (set, id) -> set.contains(id.id), readScope, node, coordinates, executes, all, kind, Ballot.ZERO,
+        sendTo(contact, readSet, (set, id) -> set.contains(id.id), (set, id) -> false, readScope, node, coordinates, executes, all, kind, Ballot.ZERO,
```
ifesdjeen, line 181: Did you intend to add this _after_ `(set, id) -> set.contains(id.id)`?

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -53,12 +53,15 @@
...
 import static accord.messages.MessageType.READ_RSP;
 import static accord.messages.ReadData.CommitOrReadNack.Insufficient;
+import static accord.messages.ReadData.CommitOrReadNack.Invalid;
 import static accord.messages.ReadData.CommitOrReadNack.Redundant;
 import static accord.messages.TxnRequest.latestRelevantEpochIndex;
 import static accord.primitives.Routables.Slice.Minimal;
 import static accord.utils.Invariants.illegalState;
 import static java.util.concurrent.TimeUnit.MICROSECONDS;
 
+// TODO (expected): if one shard timesout waiting to reply, but another shard produces a reply, return a partial response (or response with suitably populated unavailable)
```
ifesdjeen, line 62: nit: times out

accord-core/src/main/java/accord/coordinate/CollectLatestDeps.java
belliottsmith: It isn't, this is just an efficient and debuggable map when you already have a sorted list of keys (which we have via topologies). You just need an array of the same size for storing the values. The alternative is a plain list, but this is not debuggable as it loses the source node id. I think it is a nice balance of clarity and efficiency, though I'll grant it might be nice in future to introduce an immutable hash collection that could be used to index an array of values.

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
belliottsmith: I think it is irrelevant which parameter we provide here, as we should always be a read contacting the primary topology. I can add a comment.

accord-core/src/main/java/accord/coordinate/tracking/QuorumIdTracker.java
belliottsmith: Wherever we actually need that sure. Though in a lot of cases the SortedListMap should work.

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: being able to serve reads and being able to coordinate are distinct properties, one does not necessarily happen before the other. We only create the async read result before we create the async coordination result.

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: We expect coordination to be ready much earlier than reads, to restore our quorum properties as quickly as possible in the event of a topology change. Being able to serve reads does not actually guarantee we can safely serve coordination requests anyway, as some of the read results come from a snapshot and do not have associated transaction information for computed dependencies.

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: It's used to refine the handling of the state machine for command stores that no longer own a range. This helps more clearly support a change that ensures prompt forward progress (it would happen eventually, but might be delayed without this change). I will add a comment explaining the rationale. This is primarily used in PreAccept to skip certain PreAccept steps that might not work if the command has been erased due to the range being retired.

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: you mean convert to ifs, or...?

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: not sure what happened 🤷

accord-core/src/main/java/accord/local/RedundantBefore.java
belliottsmith: right, the first branch is explicitly if `is(ExclusiveSyncPoint)`

accord-core/src/main/java/accord/local/RedundantBefore.java
belliottsmith: Unqualified version, i.e. can any of the qualified filters apply. Equivalent to checking the weakest criteria, but to make clearer the intent.

accord-core/src/main/java/accord/local/RedundantStatus.java
belliottsmith: Broadly speaking, yes. I am not sure we need all of the distinctions, I suspect we can collapse some (and have added some TODOs to revisit), but I don't have time this second to refine them all. Perhaps when we have no failing seeds we can look at some simplifications.

belliottsmith: The main thing is making sure we can distinguish between locally redundant, shard redundant and no longer owned, plus cases where this is completely the case or only partially the case. But in different cases we care about different facets, e.g. when cleaning up locally we care about fully shard redundant (plus variations thereof for lost ownership), but when we are executing we care about locally partially redundant (plus variations thereof for lost ownership), and when we are fetching data from remote replicas that have erased their state we care only about _partial_ local redundancy (which means the transaction must have executed locally if it executes, so if we don't know it we can invalidate it). This is further complicated by bootstrap, where the distinction between whether it is fully or partially pre-bootstrap (and fully or partially shard redundant) matter. 

I hope to formalise the GC state machine once we're otherwise broadly where we want to be.

I did look at encoding each dimension independently but the refactor looked too big and didn't clearly simplify things - and would no doubt have introduced new problems to diagnose.

accord-core/src/main/java/accord/local/cfk/CommandsForKey.java
belliottsmith: Truncation is normal truncation, and I have stopped encoding truncation in the CFK. Basically the CFK continues retaining any truncated transaction as its pre-truncation state. 

Pruned on the other hand means we have decided the CFK probably doesn't need to know about the transaction anymore, so we remove it. This may happen to a truncated or non-truncated transaction. If we later witness it as a dependency, we may have to load it from disk to see what we knew about it. So we trade smaller memory footprint for occasionally having to consult command state.

accord-core/src/main/java/accord/local/cfk/PostProcess.java
belliottsmith: Yes, I suppose it is. Logically it made more sense to me to represent it as deducting the number of removed items. But it also makes sense that we would reset to the start if we remove from there to our current pointer.

accord-core/src/main/java/accord/local/cfk/Pruning.java
belliottsmith: The comparator never returns 0, so it is always a negative result.

accord-core/src/main/java/accord/messages/AbstractRequest.java
belliottsmith: thanks!

accord-core/src/main/java/accord/local/RedundantStatus.java
belliottsmith: It occurs to me a good approach might be to create separate (but very similar) enums, for each purpose. I'll add a TODO and explore this in a later patch.

accord-core/src/main/java/accord/messages/Commit.java
belliottsmith: No, but perhaps it is counter-intuitive. If isRead is true, then we always register a callback (as anything else would make no sense). So we only invoke the `false` test for non-reads.

accord-core/src/main/java/accord/local/Commands.java
ifesdjeen: Yeah, I thought maybe one top level if and only 1 level of ternary.

PR #154, opened 2025-01-06
Accord: Bootstrap should detect insufficient source

belliottsmith:


accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -285,16 +286,30 @@ protected void unsafeUpsertRedundantBefore(RedundantBefore addRedundantBefore)
...
-    protected synchronized void unsafeSetSafeToRead(NavigableMap<Timestamp, Ranges> newSafeToRead)
+    final void unsafeSetSafeToRead(NavigableMap<Timestamp, Ranges> newSafeToRead)
     {
         this.safeToRead = newSafeToRead;
     }
 
-    protected void unsafeSetBootstrapBeganAt(NavigableMap<TxnId, Ranges> newBootstrapBeganAt)
+    protected void loadSafeToRead(NavigableMap<Timestamp, Ranges> newSafeToRead)
+    {
+        Invariants.checkState(safeToRead == null || safeToRead.equals(ImmutableSortedMap.of(Timestamp.NONE, Ranges.EMPTY)));
```
ifesdjeen: nit: Been meaning to do this for a while, but never got to: should we put default values for safe to read and bootstrap began at somewhere handy? I keep repeating them here and there.

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -311,8 +326,34 @@ protected void updateMaxConflicts(Command prev, Command updated)
...
+    {
+        Timestamp max = Timestamp.NONE;
+        MaxConflicts updated = maxConflicts;
+        for (Map.Entry<? extends Timestamp, Ranges> e : map.entrySet())
+        {
+            Timestamp at = e.getKey();
+            if (at.compareTo(Timestamp.NONE) > 0)
+            {
+                updated = updated.update(e.getValue(), at);
+                max = Timestamp.max(max, at);
```
ifesdjeen, line 373: I might be missing something, but do we have a guarantee that existing `maxConflicts` does not have a higher timestamp? In other words, we collect max timestamp only from the provided map, but not from the existing `maxConflict`

accord-core/src/main/java/accord/local/cfk/CommandsForKey.java
```diff
@@ -1433,52 +1497,58 @@ private CommandsForKeyUpdate update(InternalStatus newStatus, Command next, bool
...
-                    }
+                    // and only when the new ballot is strictly greater
+                    if (updated.acceptedOrCommitted().compareTo(cur.ballot()) <= 0)
+                        return this;
+                }
+                else if (c == 0)
+                {
+                    // we're updating to the same state; we only do this with a strictly greater ballot;
+                    // even so, if we have no executeAt or deps there's nothing to record
+                    if (updated.acceptedOrCommitted().compareTo(cur.ballot()) <= 0 && !newStatus.hasExecuteAtOrDeps())
```
ifesdjeen: don't we need to consider `(!newStatus.hasExecuteAtAndDeps)` here?

accord-core/src/main/java/accord/local/cfk/CommandsForKey.java
```diff
@@ -1819,13 +1893,28 @@ public CommandsForKeyUpdate withRedundantBeforeAtLeast(RedundantBefore.Entry new
      */
     public CommandsForKeyUpdate withRedundantBeforeAtLeast(RedundantBefore.Entry newBoundsInfo, boolean force)
     {
-        // TODO (required): handle receiving an entry from the past, e.g. on reload (OR expunge all CFK on restart)
-        if (!force && newBoundsInfo.gcBefore.equals(boundsInfo.gcBefore)
-            && newBoundsInfo.bootstrappedAt.equals(boundsInfo.bootstrappedAt)
-            && newBoundsInfo.locallyDecidedAndAppliedOrInvalidatedBefore.equals(boundsInfo.locallyDecidedAndAppliedOrInvalidatedBefore)
-            && newBoundsInfo.endOwnershipEpoch == boundsInfo.endOwnershipEpoch)
+        // we can't let HLC epoch go backwards as this breaks assumptions around maxHlc tracking
```
ifesdjeen, line 1888: I could not find any usages where `force` would be passed as `true`, do we have any plans for this? If not, maybe worth considering to remove to simplify logic a bit?

accord-core/src/main/java/accord/local/cfk/Utils.java
```diff
@@ -261,6 +262,28 @@ static TxnId[] removePrunedAdditions(TxnId[] additions, int additionCount, TxnId
         return prunedIds;
     }
 
+    static TxnId[] removeUnmanaged(TxnId[] ids)
+    {
+        int count = 0;
```
ifesdjeen, line 269: Nit: if we have mostly-managed, we could just pre-allocate an array with same size as ids, and copy as we go, then system.arrays copy the `count` elements. Feels it might be a bit more efficient than iterating twice.

ifesdjeen (approved):

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: I don't follow. The goal here is to ensure all of the timestamps in the map are represented in `MaxConflicts`

accord-core/src/main/java/accord/local/cfk/Utils.java
belliottsmith: I will make a note to consider this more later. My intuition is that counting will have negligible cost, as these arrays should normally be small. But equally in C* we expect to have very few unmanaged transactions.

belliottsmith: Hmm, actually no I think the current approach is optimal in the case we expect it to be mostly managed, as we don't copy anything if we don't remove anything?

PR #156, opened 2025-01-14
Coordinator optimisation + medium path; do not restart recovery, dire…

belliottsmith:
…ctly calculate continuation criteria

fix: - isOutOfRange when invoked from callback for some txn < Committed

fix: - handle loadingPruned that is pre-bootstrap on update to bootstrappedAt

fix - journal replay can overwrite in memory state loaded for a command not yet replayed


iamaleksey (approved):
With a few comments addressed out of band, +1

PR #157, opened 2025-01-15
Make sure accord burn test flushes synchronously; fix problem with it…

ifesdjeen:
…eration order

### DIFF
--- a/accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
+++ b/accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
@@ -61,6 +61,7 @@ import accord.utils.Invariants;
 import accord.utils.RandomSource;
 import accord.utils.async.AsyncChain;
 import accord.utils.async.AsyncChains;
+import accord.utils.async.AsyncResults;
 import accord.utils.async.Cancellable;
 
 import static accord.api.Journal.CommandUpdate;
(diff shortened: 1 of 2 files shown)


accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
```diff
@@ -351,7 +352,9 @@ public void postExecute()
 
                 Command before = safe.original();
                 Command after = safe.current();
-                commandStore.journal.saveCommand(commandStore.id(), new CommandUpdate(before, after), () -> {});
+                AsyncResults.SettableResult<Void> flush = new AsyncResults.SettableResult<>();
+                commandStore.journal.saveCommand(commandStore.id(), new CommandUpdate(before, after), () -> flush.setSuccess(null));
+                AsyncChains.getUnchecked(flush);
```
dcapwell, line 357: isn't this problematic for places like burn test where blocking logic isn't allowed?

ifesdjeen: In memory journal is already synchornous; in Cassandra integration this simply ensures we wait for flush, so we simply make journal write synchronous. Since BurnTest does not control journal threads, it does not really know whether the call is synchronous or no, so from its perspective, `postExecute` will just be executed slightly longer.

In retrospect, I actually think for _burn test_ this is actually optional, because we are using mmap, so by the time we exit the write function, data is already readable for the remainder of the test. I still felt like it is important to exercise flush callbacks. 

A much more important part of this PR is on Cassandra side.

PR #158, opened 2025-01-22
Agree a distributed uniqueHlc to use for Apply

belliottsmith:
- detect and save whether RX can be used as a GC bound for HLCs
 - have RX durably record if they witnessed any superseding epoch and use this for HLC GC
 - rework GC to retain applyAt for writes

Also improve:
 - PreCommitted overwrites fast path vote, then loses tie-break with NotAccept, so we don't recover
 - track executeAtLeast and uniqueHlc separately from WaitingOn in CommandChange
 - Always propose executeAt.hlc() >= txnId.hlc() Also fix:
 - Dangling recovery when Commit rejected
 - Failed epoch fetch handling in DefaultProgressLog
 - CommandsForKey must track Accept executeAt for computing earlierWait collection on recovery


accord-core/src/main/java/accord/impl/AbstractFetchCoordinator.java
```diff
@@ -323,7 +323,7 @@ public static class FetchResponse extends ReadOk
 
         public FetchResponse(@Nullable Ranges unavailable, @Nullable Data data, @Nonnull Timestamp safeToReadAfter)
         {
-            super(unavailable, data);
+            super(unavailable, data, 0);
```
ifesdjeen: Nit: should we introduce a NO_HLC magic constant? Just seen 0 on 2 occasions.

accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -503,41 +523,48 @@ public static int getFlags(Command before, Command after)
...
+        return setChanged(field, flags);
+    }
 
-        return setFieldChanged(field, flags);
+    private static <OBJ> int collectFlags(OBJ lo, OBJ ro, ToLongFunction<OBJ> convert, Field field, int flags)
+    {
+        long l = 0, r = 0;
+        if (lo != null) l = convert.applyAsLong(lo);
+        if (ro != null) r = convert.applyAsLong(ro);
+        return l == r ? flags : r == 0 ? setFieldIsNullAndChanged(field, flags) : setChanged(field, flags);
```
ifesdjeen: nit: would you mind to replace this line with

```
        if (l == r) return flags;
        return r == 0 ? setFieldIsNullAndChanged(field, flags) : setChanged(field, flags);
```

just to break up double-ternary

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
ifesdjeen, line 1: nit: checkstyle complains about this file's imports

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -763,7 +766,12 @@ CommandSummaries commandsForRanges()
                     summaries.put(summary.txnId, summary);
             }
 
-            return commandsForRanges = () -> summaries;
+            final Kinds kinds = new Kinds(Read, ExclusiveSyncPoint, SyncPoint);
```
ifesdjeen, line 768: nit: should we turn this into a static constant?

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -457,23 +464,37 @@ private void addToRunBuffer(TxnState add)
         runBuffer[runBufferCount++] = add;
     }
 
+    private void rerunWithPendingEpoch()
+    {
+        long minEpoch = Long.MAX_VALUE;
+        for (int i = 0 ; i < awaitingEpochBufferCount ; ++i)
+            minEpoch = Math.min(awaitingEpochBuffer[i].run.txnId.epoch(), minEpoch);
+        Invariants.checkArgument(minEpoch != Long.MAX_VALUE);
+        isAwaitingEpoch = true;
```
ifesdjeen, line 476: nit: isAwaitingEpoch seems to be unused outside this method; was there a check you intended to add?

accord-core/src/test/java/accord/impl/basic/InMemoryJournal.java
```diff
@@ -497,6 +501,9 @@ private Diff(int flags, CommandUpdate update)
                     case EXECUTES_AT_LEAST:
                         changes.put(EXECUTES_AT_LEAST, after.executesAtLeast());
                         break;
+                    case MIN_UNIQUE_HLC:
```
ifesdjeen, line 504: If we already store HLC in waitingOn, do we need to also save it separately?

ifesdjeen (approved):

accord-core/src/test/java/accord/impl/basic/InMemoryJournal.java
belliottsmith: We aren’t, I’ve stopped WaitingOn from persisting either executesAtLeast or UniqueHLC. This also means an update to one of these components only requires updating that component. The WaitingOn provider accepts them as inputs now.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
belliottsmith: I opted not to because it's test-only, but can if you have a strong preference

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
belliottsmith: yes, good catch!

accord-core/src/main/java/accord/impl/CommandChange.java
belliottsmith: I've opted to update the formatting. I'm not a huge fan of mixing return types, particularly at this level of one of each kind. I would be OK with switching to three return statements if you prefer though.

PR #159, opened 2025-01-22
Avoid double loading in Cassandra side of journal

ifesdjeen:


accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -414,6 +415,8 @@ private static Command.Truncated truncated(TxnId txnId, SaveStatus status, Statu
                 case TruncatedApplyWithOutcome:
                 case TruncatedApplyWithDeps:
                 case TruncatedApply:
+                    if (status != TruncatedApplyWithOutcome)
```
belliottsmith: Can we add a TODO to tidy this up when we have a better solution?

belliottsmith: Would be good to mention the reason this arises. Unfortunately I've forgotten since we spoke, I think it may be because of compaction? Because if we write out a truncate record to the log it should null out this field.

ifesdjeen: Reverted for now; I'm sure it'll show up in burn test. When it does, we will document and/or fix it, wdyt?

belliottsmith (approved):

PR #161, opened 2025-01-28
CASSANDRA-20256: Fix 0xe299233a8efdcefdl in AccordSimulationRunner

ifesdjeen:


accord-core/src/main/java/accord/local/cfk/Updating.java
```diff
@@ -788,7 +788,13 @@ static CommandsForKeyUpdate updateUnmanaged(CommandsForKey cfk, SafeCommand safe
         if (safeCommand.current().hasBeen(Status.Truncated))
             return cfk;
 
-        Command.Committed command = safeCommand.current().asCommitted();
+        Command orig = safeCommand.current();
+        if (orig.saveStatus() == SaveStatus.Uninitialised)
```
belliottsmith: maybe static import, and inline the two conditions? Also, good to use `orig.txnId().is(EphemeralRead)`

belliottsmith (approved):

PR #162, opened 2025-01-28
Refine optimisations

belliottsmith:


accord-core/src/main/java/accord/coordinate/FetchData.java
```diff
@@ -79,6 +79,18 @@ private static class FetchRequest
         final BiConsumer<? super FetchResult, Throwable> callback;
 
         public FetchRequest(Known fetch, TxnId txnId, InvalidIf invalidIf, @Nullable Timestamp executeAt, Participants<?> fetchKeys, EpochSupplier lowEpoch, EpochSupplier highEpoch, BiConsumer<? super FetchResult, Throwable> callback)
+        {
```
ifesdjeen, line 82: nit: unused constructor; but could be used in C*?

ifesdjeen: nit: unused constructor; but could be used in C*?

accord-core/src/main/java/accord/primitives/Timestamp.java
```diff
@@ -277,6 +277,18 @@ public int compareTo(@Nonnull Timestamp that)
         return c;
     }
 
+    public boolean isAtLeastByEpochAndHlc(@Nonnull Timestamp that)
+    {
+        if (this == that) return true;
+        int cmpEpoch = Long.compare(this.epoch(), that.epoch());
+        int cmpHlc = Long.compare(this.hlc(), that.hlc());
+        if (cmpEpoch < 0 || cmpHlc < 0) return false;
+        if (cmpEpoch > 0 || cmpHlc > 0) return true;
```
ifesdjeen, line 286: Can we be in a situation where epoch is ahead and HLC is behind?

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -301,23 +303,40 @@ public List<TxnId> activeBefore(TxnId before)
...
+                return;
+
+            boolean notify = state.waiting().blockedUntil()  != NotBlocked
+                          && state.waiting().homeSatisfies() == NotBlocked;
+
+            if (notify)
+            {
+                // the command might be invalidated, which should be established on load, so simply load the command
+                TxnId txnId = state.txnId;
+                safeStore.commandStore().execute(PreLoadContext.contextFor(txnId), safeStore0 -> {
```
ifesdjeen: Maybe a silly optimization, and definitely outside scope of this patch, but should we maybe make txnid implement preloadcontext for itself? This seems like a rather common pattern.

accord-core/src/main/java/accord/messages/RecoverAwait.java
```diff
@@ -18,24 +18,31 @@
...
 
+import static accord.api.ProgressLog.BlockedUntil.CommittedOrNotFastPathCommit;
 import static accord.primitives.Routables.Slice.Minimal;
 import static accord.primitives.Timestamp.Flag.HLC_BOUND;
 
 public class RecoverAwait extends Await
 {
     final TxnId recoverId;
     private transient boolean rejects;
+    private transient boolean cannotAccept;
```
ifesdjeen, line 52: Nit: unused constructor on L47

accord-core/src/main/java/accord/utils/ArrayBuffers.java
```diff
@@ -169,6 +174,73 @@ default int[] completeAndDiscard(int[] buffer, int usedSize)
...
+         * Return a {@code long[]} of size at least {@code minSize}, possibly from a pool.
+         * This array may not be zero initialized, and its contents should be treated as random.
+         */
+        long[] getLongs(int minSize);
+    }
+
+    public interface LongBuffers extends LongBufferAllocator
+    {
+        /**
+         * Return an {@code int[]} of size at least {@code minSize}, possibly from a pool,
```
ifesdjeen: nit: probably `long` here

accord-core/src/main/java/accord/utils/ArrayBuffers.java
```diff
@@ -169,6 +174,73 @@ default int[] completeAndDiscard(int[] buffer, int usedSize)
...
+         * However, it is guaranteed that a failure to do so does not leak memory or pool space, only produces some
+         * additional garbage.
+         *
+         * @return true if the buffer is discarded (and discard-able), false if it was retained or is believed to be in use
+         */
+        boolean discard(long[] buffer, int usedSize);
+
+        /**
+         * Equivalent to
+         *   int[] result = complete(buffer, usedSize);
```
ifesdjeen: nit: probably `long[]` here

ifesdjeen (approved):

accord-core/src/main/java/accord/primitives/Timestamp.java
belliottsmith: yes, it's possibly to agree two different transactions to execute at (2,1) and (1,2)

accord-core/src/main/java/accord/coordinate/FetchData.java
belliottsmith: removed

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
belliottsmith: I like this idea. It's a small change, will fold it in

PR #163, opened 2025-01-30
Refine redundantstatus

belliottsmith:


accord-core/src/main/java/accord/api/ProtocolModifiers.java
```diff
@@ -203,6 +204,15 @@ public static class Toggles
         public static boolean requiresUniqueHlcs() { return requiresUniqueHlcs; }
         public static void setRequiresUniqueHlcs(boolean newRequiresUniqueHlcs) { requiresUniqueHlcs = newRequiresUniqueHlcs; }
 
+        private static int markStaleIfCannotExecute = TinyEnumSet.encode(Txn.Kind.Write);
+        public static boolean markStaleIfCannotExecute(Txn.Kind kind) { return TinyEnumSet.contains(markStaleIfCannotExecute, kind); }
+        public static void setMarkStaleIfCannotExecute(Txn.Kind ... kinds)
```
ifesdjeen, line 209: nit: space before `...`

accord-core/src/main/java/accord/coordinate/CoordinateGloballyDurable.java
```diff
@@ -31,12 +31,12 @@
 import accord.utils.async.AsyncResult;
 import accord.utils.async.AsyncResults.SettableResult;
 
-// TODO (expected): this does not need to query every shard; can disseminate globally any sub-range of the ring
-//  (indeed, we could slice both the query and dissemination only so that they always overlap)
+// TODO (expected): this does not need to query every shard OR more than one replica per shard;
+//  can disseminate globally any sub-range of the ring.
+//  Infact, we could simply autonomously disseminate our latest information to some subset of replicas
```
ifesdjeen, line 36: nit: honest question: is `infact` commonly spelled as one word?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -923,51 +913,45 @@ private static Command purge(Command command, @Nonnull StoreParticipants newPart
 
             case TRUNCATE:
                 Invariants.require(command.saveStatus().compareTo(TruncatedApply) < 0);
-                Invariants.requireArgument(command.known().is(ApplyAtKnown));
+                Invariants.requireArgument(command.known().is(ApplyAtKnown) || (command.known().is(ExecuteAtKnown) && !command.txnId().is(Write)));
```
ifesdjeen: nit: could you add brackets to signify operator precedence here?

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLogs.java
```diff
@@ -54,7 +56,7 @@ static boolean pauseForTest(DefaultProgressLog progressLog)
             return false;
 
         if (!paused.containsKey(progressLog))
-            paused.putIfAbsent(progressLog, () -> progressLog.commandStore.execute(progressLog));
+            paused.putIfAbsent(progressLog, () -> progressLog.commandStore.execute(empty(), progressLog));
```
ifesdjeen: discussed offline: looks like we may want to add `begin` here, since when runnable runs, we only create an async chain but do not start it.

ifesdjeen (approved):

accord-core/src/main/java/accord/coordinate/CoordinateGloballyDurable.java
belliottsmith: Good question. I think perhaps, otherwise I probably wouldn't have done it? But I didn't consciously decide to, so who knows. It looks like "official" versions have a space, but English doesn't actually have any official official version, so 🤷

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: it looks like there are already brackets?

accord-core/src/main/java/accord/api/ProtocolModifiers.java
belliottsmith: Also looks like already a space before, and I don't think I consciously edited it. Not sure what's going on.

PR #164, opened 2025-02-04
Fixes

belliottsmith:
- Even if we can decide autonomously that we took the fast path, we must still wait for earlier transactions to decide themselves
 - We must update a command that is in CFK.loadingPruned whether or not it is outOfRange
 - We must visit pruned commands that are transitive dependencies of RX to ensure dependencies are propagated
 - flagsWithoutDomainAndKind() -> flagsWithoutDomainOrKindOrCardinality() [to fix integration regression]
 - Don't invoke uniqueNow() twice when allocate nextTxnId
 - BTreeReducingRangeMap edge case that corrupts map when merge function does not return one of its inputs

### DIFF
--- a/accord-core/src/main/java/accord/local/Node.java
+++ b/accord-core/src/main/java/accord/local/Node.java
@@ -704,8 +704,13 @@ public class Node implements ConfigurationService.Listener, NodeCommandStoreServ
      */
     public TxnId nextTxnId(Txn.Kind rw, Domain domain, Cardinality cardinality, int flags)
     {
+        return nextTxnId(uniqueNow(), rw, domain, cardinality, flags);
+    }
+
+    private static TxnId nextTxnId(Timestamp now, Txn.Kind rw, Domain domain, Cardinality cardinality, int flags)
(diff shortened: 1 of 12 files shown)


accord-core/src/main/java/accord/local/Node.java
```diff
@@ -703,9 +703,14 @@ public TxnId nextTxnId(Txn.Kind rw, Domain domain, int flags)
      * TODO (required): Don't use a new epoch for the TxnId at least until we know its definition
      */
     public TxnId nextTxnId(Txn.Kind rw, Domain domain, Cardinality cardinality, int flags)
+    {
+        return nextTxnId(uniqueNow(), rw, domain, cardinality, flags);
+    }
+
+    private TxnId nextTxnId(Timestamp now, Txn.Kind rw, Domain domain, Cardinality cardinality, int flags)
```
ifesdjeen: nit: can also be static to preclude double-calling in future

ifesdjeen (approved):

PR #167, opened 2025-02-06
CASSANDRA-20297: Fix null field accounting

ifesdjeen:


accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -541,10 +541,16 @@ private static <OBJ, VAL> int collectFlags(OBJ lo, OBJ ro, Function<OBJ, VAL> co
...
+        if (lo == null)
+        {
+            if (r == null)
+                return setFieldIsNullAndChanged(field, flags);
+            else
+                return setChanged(field, flags);
+        }
+        VAL l = convert.apply(lo);
 
         if (l == r) return flags; // no change
```
belliottsmith: This could be handled here in these return flags, but could you perhaps explain more (also in a comment) why this is necessary? I think it might introduce bugs with compaction, as we seem to be _setting_ to null any field that is null in both input and output, but for compaction of partial results we can expect fields to just not participate in the compaction (i.e. be null in both sides) but be *not null* in any constructed object, and this could cause them to be shadowed and nulled out, no?

Is this to handle the `TruncatedApply` case, and erasing `Result`? If so, I think it is better to special-case truncation, and to produce a null field mask directly.

belliottsmith (approved):
This works for me, although I might slightly prefer that we precomputed some flags and saved them in `SaveStatus` and just did |= saveStatus.commandChangeFlags. This would permit us to clear other registers too (e.g. DepsErased, ExecuteAtErased etc), and do it in a simple declarative way.

ifesdjeen:
@belliottsmith great idea! Pushed the implementation and test for this.

belliottsmith (approved):
Only one suggestion: there's an edge case with `Vestigial` for which - for distributed inference - we don't declare that any of the values are Erased, even though we have erased them. Perhaps we could reify this in the SaveStatus somehow, but for now at least it would be most robust to special-case this SaveStatus and set all of the null/changed flags for `Vestigial`.

PR #168, opened 2025-02-06
Cleanup 6feb25

belliottsmith:


accord-core/src/main/java/accord/local/Command.java
```diff
@@ -685,12 +688,16 @@ public static Accepted accepted(ICommand common, SaveStatus status)
         {
             super(copy, status);
             validateDeps(participants(), partialDeps());
+            if (txnId().toString().equals("[3,23124,17(RR),2]") && status() == PreCommitted && executeAt().equals(txnId()))
```
ifesdjeen: nit: looks like remaining debug logging in this file, both here and below

accord-core/src/main/java/accord/local/Command.java
```diff
@@ -685,12 +688,16 @@ public static Accepted accepted(ICommand common, SaveStatus status)
...
+                System.out.println();
         }
 
         Accepted(TxnId txnId, SaveStatus status, Durability durability, @Nonnull StoreParticipants participants, Ballot promised, Timestamp executeAt, PartialTxn partialTxn, PartialDeps partialDeps, Ballot acceptedOrCommitted)
         {
             super(txnId, status, durability, participants, promised, executeAt, partialTxn, partialDeps, acceptedOrCommitted);
             validateDeps(participants(), partialDeps);
+            if (txnId().toString().equals("[3,23124,17(RR),2]") && status() == PreCommitted && executeAt().equals(txnId()))
```
ifesdjeen: nit: looks like remaining debug logging in this file

accord-core/src/main/java/accord/coordinate/CollectLatestDeps.java
```diff
@@ -95,10 +109,15 @@ public void onFailure(Id from, Throwable failure)
...
             return;
 
+        if (failure != null)
+            this.failure = FailureAccumulator.append(this.failure, failure);
+
         if (tracker.recordFailure(from) == Failed)
         {
             isDone = true;
-            callback.accept(null, new Timeout(txnId, route.homeKey()));
+            if (this.failure == null)
```
ifesdjeen, line 118: looks like above we can report Failure as a replay, so it not necessarily is a timeout, or?

accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
```diff
@@ -133,16 +136,33 @@ public static CoordinationAdapter<Result> recoverExclusiveSyncPoint()
...
+                                                             route, txnId, executeAt, SHARE, QuorumEpochIntersections.accept);
                 new ProposeTxn(node, accept, route, kind, ballot, txnId, txn, executeAt, deps, callback).start();
             }
 
+            @Override
+            public void proposeOnly(Node node, Route<?> require, Route<?> sendTo, SelectNodeOwnership selectNodeOwnership, FullRoute<?> route, Accept.Kind kind, Ballot ballot, TxnId txnId, Txn txn, Timestamp executeAt, Deps deps, BiConsumer<? super Deps, Throwable> callback)
+            {
+                Topologies accept = node.topology().reselect(null, QuorumEpochIntersections.preacceptOrRecover,
+                                                             sendTo, txnId, executeAt, selectNodeOwnership, QuorumEpochIntersections.accept);
+                new ProposeOnly(node, accept, sendTo, route, kind, ballot, txnId, txn, executeAt, deps, callback).start();
```
ifesdjeen, line 165: nit: looks like this `start` might be quite easy to miss. If we always call it (which we seem to do), should we maybe move the start call to constructor, or maybe make a static method?

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
```diff
@@ -192,11 +198,25 @@ protected Action process(Id from, ReadReply reply)
         CommitOrReadNack nack = (CommitOrReadNack) reply;
         switch (nack)
         {
-            default: throw new IllegalStateException();
+            default: throw new UnhandledEnum(nack);
+            case Waiting:
+                if (from.id != node.id().id)
+                    throw new UnhandledEnum(nack);
```
ifesdjeen: nit: should we throw unhandled enum here, or maybe something more informative, like node id mismatch?

accord-core/src/main/java/accord/coordinate/Propose.java
```diff
@@ -102,32 +108,37 @@ public void onSuccess(Id from, AcceptReply reply)
...
                 break;
 
+            case Truncated:
             case Redundant:
-                if (reply.supersededBy != null || ballot.equals(Ballot.ZERO))
+                if (sendTo == route || !isSufficientPartialReply(reply, from))
                 {
-                    isDone = true;
-                    callback.accept(null, new Preempted(txnId, route.homeKey()));
+                    Throwable failNow = reply.outcome == AcceptOutcome.Truncated
```
ifesdjeen: could we unwrap this ternary?

```
            if (reply.outcome == AcceptOutcome.Truncated)
                failNow = new Truncated(txnId, route.homeKey());
            else if (reply.supersededBy != null || ballot.equals(Ballot.ZERO))
                failNow = new Preempted(txnId, route.homeKey());
            else if (reply.committedExecuteAt != null)
                failNow = new Redundant(txnId, route.homeKey(), reply.committedExecuteAt);
            else
                failNow = null;
```

(please feel free to ignore if i am the only person who struggles parsing long ternaries)

accord-core/src/main/java/accord/coordinate/Propose.java
```diff
@@ -140,17 +151,27 @@ else if (reply.committedExecuteAt != null)
...
-        // TODO (required): we aren't tracking the specific failure here to report
+        if (failure != null)
+            this.failure = FailureAccumulator.append(this.failure, failure);
+
         if (acceptTracker.recordFailure(from) == Failed)
         {
             isDone = true;
-            callback.accept(null, new Timeout(txnId, route.homeKey()));
+            if (this.failure == null)
+                this.failure = new Timeout(txnId, route.homeKey());
```
ifesdjeen: same question about timeout here: since we are calling `onFailure` with `null` failure from success on nack, is it right to replace failure with timeout here?

accord-core/src/main/java/accord/coordinate/ProposeOnly.java
```diff
@@ -0,0 +1,59 @@
...
+    void onAccepted()
+    {
+        Deps deps = mergeDeps();
+        callback.accept(deps, null);
+    }
+
+    @Override
+    CoordinationAdapter<Deps> adapter()
+    {
+        throw new UnsupportedOperationException();
```
ifesdjeen, line 57: I might be missing context, but could you elaborate why we do not need an adapter for this one?

accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -264,84 +271,53 @@ public Cleanup shouldCleanup(Input input, Agent agent, RedundantBefore redundant
...
-            Builder builder = new Builder(txnId, ALL);
-
-            builder.count++;
-            builder.nextCalled = true;
-
-            if (saveStatus != null)
+            int mask = saveStatusMasks[newSaveStatus.ordinal()];
+            // low flag bits represent fields already nulled out, so no need to visit them again
+            int iterable = toIterableSetFields(mask) & ~flags;
+            for (Field next = nextSetField(iterable); next != null; iterable = unsetIterable(next, iterable), next = nextSetField(iterable))
```
ifesdjeen, line 309: nice, i really like this new implementation of `truncate`

accord-core/src/main/java/accord/coordinate/CollectLatestDeps.java
belliottsmith: We don't really distinguish between Timeout and having exhausted our options in most places. We do technically have `Exhausted`, I can switch to that if you like? It is probably more correct. We should at some point go through all our exception logic and make sure it is consistent, I noticed already a number of inconsistencies.

accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
belliottsmith: Normally I introduce such a static method, but in this case we only invoke from an accessor class which basically serves that static method functionality. But, I can add the extra hop if you like, pretty neutral on it.

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
belliottsmith: It means we aren't expecting this reply from anyone but ourselves, and so we don't know how to handle the reply. I am happy to use another exception, but I don't think node id mismatch is correct, it's that the enum shouldn't have come from another node, and it did, so we don't know what to do.

belliottsmith: (That said, I could just `break` as it is totally fine to ignore this reply from another node really)

accord-core/src/main/java/accord/coordinate/Propose.java
belliottsmith: Sure. I think you have more of a problem with ternaries than me, but I am not adequately enamoured with them to prefer them strongly.

accord-core/src/main/java/accord/coordinate/Propose.java
belliottsmith: `Exhausted` again?

accord-core/src/main/java/accord/coordinate/ProposeOnly.java
belliottsmith: The "Only" means only propose, don't continue to the rest of the state machine (callback once done and let the caller decide how to proceed). So, there's no reason to invoke the adapter to start the next coordination that advances the state machine.

accord-core/src/main/java/accord/coordinate/CoordinationAdapter.java
ifesdjeen: i have inspected all adjacent classes and there seem to be no problems as of now. Just making sure we do not miss this in the future.

accord-core/src/main/java/accord/coordinate/CollectLatestDeps.java
ifesdjeen: Maybe add a TODO for now and we can do all of them in one sweep?

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
ifesdjeen: I am mostly concerned that in case codebase diverges from the deployed version we might mistake this unhandled enum from the one on L201. I don't mind throwing here, just maybe assertion error, or some other exception? I don't mind swallowing the response in this case, either.

accord-core/src/main/java/accord/coordinate/Propose.java
ifesdjeen: Yes

accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -264,84 +271,53 @@ public Cleanup shouldCleanup(Input input, Agent agent, RedundantBefore redundant
...
             {
-                builder.flags = setChanged(SAVE_STATUS, builder.flags);
-                builder.saveStatus = saveStatus;
+                switch (next)
+                {
+                    default: throw new UnhandledEnum(next);
+                    case PARTICIPANTS:      participants = null;                     break;
```
ifesdjeen, line 314: definitely a nit and can be done in a separate patch, but would it make sense to make `clear` use this `switch` will `all fields` mask, and then simply unset `flags`? In this case we only have one place to update when adding a new field.

accord-core/src/main/java/accord/primitives/LatestDeps.java
```diff
@@ -60,29 +63,110 @@ public static LatestDeps create(boolean inclusiveEnds, RoutingKey[] starts, Late
...
             }
             else
             {
-                CollectLatestDeps.withLatestDeps(node, txnId, route, missing, executeAt, (extraDeps, fail) -> {
-                    if (fail != null)
-                    {
-                        failureCallback.accept(null, fail);
-                    }
+                // we merge with committed to make sure we can send a full Commit to any replica that overlaps the stable and unstable ranges
+                Deps notaccepted = committed.with(merge.mergeProposal(propose));
```
ifesdjeen, line 119: nit: `notAccepted` for consistency with other places?

accord-core/src/main/java/accord/topology/Topologies.java
```diff
@@ -39,6 +39,8 @@
 //  (e.g. at least implementing Topologies by Topology)
 public interface Topologies extends TopologySorter
 {
+    enum SelectNodeOwnership { RESELECT, SHARE }
```
ifesdjeen: could you add 1-line comments to both cases? They are obvious in context of the patch, but might be hard to understand after a while.

Also, maybe `RECOMPUTE` or `SLICE` or `RESLICE`?

ifesdjeen (approved):

accord-core/src/main/java/accord/impl/CommandChange.java
belliottsmith: I like the idea, but I think I would want to double check hotspot does the sensible thing and collapses that into direct field updates, else we could do a lot of wasted work.

accord-core/src/main/java/accord/coordinate/CollectLatestDeps.java
belliottsmith: I've just updated to Exhausted for now. There's already TODOs littered about the place to look at our exception handling, but I'll add another here as well to be sure.

PR #169, opened 2025-02-06
CASSANDRA-20302: Expose epoch ready state as a vtable for operators to inspect things

dcapwell:


accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -806,6 +806,8 @@ public interface Setup<State, SystemUnderTest>
         private Set<Setup<State, SystemUnderTest>> unknownWeights = null;
         @Nullable
         private Map<Predicate<State>, List<Setup<State, SystemUnderTest>>> conditionalCommands = null;
+        @Nullable
+        private Map<Predicate<State>, List<Pair<Setup<State, SystemUnderTest>, Integer>>> conditionalCommandsKnownWeights = null;
```
dcapwell, line 825: this is here so we can disable repair when there are down nodes while running with Accord.  `nodetool repair --force` is timing out and looks to be durability scheduling related according to the logs.  I disabled running repairs when there are downed nodes and now we are mostly stable

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -486,15 +486,10 @@ public <State, SystemUnderTest> void check(Commands<State, SystemUnderTest> comm
...
-                        {
-                            commands.destroySut(sut, t);
-                            commands.destroyState(state, t);
-                        }
-                        catch (Throwable t2)
-                        {
-                            t.addSuppressed(t2);
-                        }
+                        State finalState = state;
+                        safeHandle(t, () -> commands.onFailure(finalState, sut, maybeRewriteHistory(history, historyTiming), t));
```
dcapwell, line 490: `onFailure` is mostly there for when i do tests in a loop i want to get a summary of the history i am looking at.  For example while working on CASSANDRA-20156 i want to find a shorter history, so getting a summary of whats going on helps me see if i should ignore the failing seed.

Also, if `destorySut` failed `destoryState` wouldn't get called, so this logic is now properly handled

belliottsmith (approved):

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -382,7 +382,7 @@ public void onEpochClosed(Ranges ranges, long epoch)
     @Override
     public void onEpochRetired(Ranges ranges, long epoch)
     {
-        topology.onEpochRedundant(ranges, epoch);
+        topology.onEpochRetired(ranges, epoch);
```
dcapwell, line 385: we spoke about this in slack.  This should have been renamed a long time ago, and since im touching this logic i am doing that rename now

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -442,6 +448,102 @@ private void timeOutCurrentListeners(long newDeadline, Agent agent)
...
+                this.global = global;
+                this.addedRanges = addedRanges;
+                this.removedRanges = removedRanges;
+                this.synced = synced;
+                this.closed = closed;
+                this.retired = retired;
+            }
+        }
+
+        private static class Builder
```
dcapwell: could inline, the first draft of this wasn't defining each field at once

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -469,6 +554,35 @@ public TopologyManager(TopologySorter.Supplier sorter, Agent agent, Id self, Sch
         this.localConfig = localConfig;
     }
 
+    public EpochsSnapshot epochsSnapshot()
+    {
+        ImmutableList.Builder<EpochsSnapshot.Epoch> builder = ImmutableList.builder();
```
belliottsmith: Suggestion: when we know the size of the contents and are already wrapping the contents in some object, I prefer to use a raw array. If not, we should provide the size to the builder.

dcapwell: TIL: `ImmutableList.builderWithExpectedSize`

dcapwell:
merged 74a3b81ca9d9e1ce7ddfd117682fc7c310f0cd99

PR #171, opened 2025-02-18
CASSANDRA-20349

belliottsmith:


accord-core/src/main/java/accord/primitives/Range.java
```diff
@@ -208,10 +207,10 @@ public int compareEndTo(RoutableKey key)
 
     private Range(RoutingKey start, RoutingKey end)
     {
-        if (start.compareTo(end) >= 0)
-            throw new IllegalArgumentException(start + " >= " + end);
-        if (startInclusive() == endInclusive())
-            throw illegalState("Range must have one side inclusive, and the other exclusive. Range of different types should not be mixed.");
+        // TODO (expected): should we at least relax to permit an empty Range?
+        Invariants.requireArgument(start.compareTo(end) < 0, start + " >= " + end);
```
ifesdjeen, line 211: nit string concatenation to format? (pre-existing though)

ifesdjeen (approved):

PR #172, opened 2025-02-21
Cleanup 19feb25

belliottsmith:


ifesdjeen (approved):
+1 
  * checked the boundary conditions everywhere where we have logic with inclusive/exclusive min/max bounds, all seems to be correct
  * it is unclear to me if we should also add an invariant for range intersection in reduce; doesn't seem to be a problem: added it locally and all tests pass, but i do not see a whole lot value in it
  * generally, tried to look over the code searching for off-by-one's / out-of-boundary conditions, and things like inclusive/exclusive bounds inconsistencies with <= etc, but can not spot anything

PR #173, opened 2025-02-24
Accord cleanup 23feb25

belliottsmith:


accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -554,15 +556,18 @@ public static void listenerUpdate(SafeCommandStore safeStore, SafeCommand safeLi
         }
     }
 
-    protected static void postApply(SafeCommandStore safeStore, Command command)
+    protected static boolean postApply(SafeCommandStore safeStore, TxnId txnId, long t0)
```
ifesdjeen: nit: looks like the result here is still unused, intended?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -604,14 +606,15 @@ public static AsyncChain<Void> applyWrites(SafeCommandStore safeStore, PreLoadCo
...
-                              return null;
-                          }));
-        else
+        if (executes.isEmpty())
             return AsyncChains.success(null);
+
+        TxnId txnId = command.txnId();
+        return command.writes().apply(safeStore, executes, command.partialTxn())
+                      .flatMap(unused -> unsafeStore.build(context, ss -> {
+                          postApply(ss, txnId, 0);
```
ifesdjeen: Could you elaborate why we won't pass a timestamp here? I checked the impl where this is used, and we would seemingly apply the negative number to metrics. I might need to check out new integration patch if there is one.

accord-core/src/main/java/accord/primitives/Deps.java
```diff
@@ -326,13 +304,76 @@ public int size()
...
+    }
+
+    public boolean isStable(int index)
+    {
+        return !applyToIndex(index, KeyOrRangeDeps::txnIdWithFlags).is(UNSTABLE);
+    }
+
+    private <T> T applyToIndex(int index, IndexedFunction<KeyOrRangeDeps, T> apply)
+    {
+        {
```
ifesdjeen: nit: are these blocks intentional / for visual separation? Just wondering

accord-core/src/main/java/accord/primitives/Range.java
```diff
@@ -215,6 +228,15 @@ private Range(RoutingKey start, RoutingKey end)
         this.end = end;
     }
 
+    private Range(RoutingKey start, RoutingKey end, AntiRangeMarker antiRange)
```
ifesdjeen, line 231: intended to be unused here?

ifesdjeen (approved):

accord-core/src/main/java/accord/primitives/Range.java
ifesdjeen: looks like it is

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: We didn't previously update the timestamps here, this is used by journal replay. I've added a check to only invoke the callback if > 0.

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: I think I previously was using the boolean to decide whether to invoke the metric callback in the caller, but decided to move it inside the method instead.

accord-core/src/main/java/accord/primitives/Deps.java
belliottsmith: for scoping, but probably date from an earlier method formulation where it mattered more. removing them...

accord-core/src/main/java/accord/primitives/Range.java
belliottsmith: yeah, it's a marker for invoking the right constructor

PR #174, opened 2025-02-28
CASSANDRA-20347: Accord: Epoch/Topology Garbage Collection

ifesdjeen:


accord-core/src/main/java/accord/api/ConfigurationService.java
```diff
@@ -196,5 +190,7 @@ default long currentEpoch()
 
     void reportEpochClosed(Ranges ranges, long epoch);
 
-    void reportEpochRedundant(Ranges ranges, long epoch);
+    void reportEpochRetired(Ranges ranges, long epoch);
+
+    void reportEpochGcd(long epoch);
```
belliottsmith: Maybe `reportEpochRemoved`?

accord-core/src/main/java/accord/impl/AbstractConfigurationService.java
```diff
@@ -237,15 +248,22 @@ public AsyncResult<Void> acknowledgeFuture(long epoch)
...
-            if (toTrim <= 0)
-                return;
 
-            epochs = new ArrayList<>(epochs.subList(toTrim, epochs.size()));
+            int count = epochs.size();
+
+            if (toTrim == 0 ||
+                // Always leave least 1 ready epoch after truncation
+                toTrim >= epochs.size() - 1 ||
+                !epochs.get(toTrim + 1).isReady())
```
belliottsmith: maybe loop to find the last isReady epoch if the one at the end of the requested range isn't suitable?

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -411,6 +407,12 @@ public void withEpoch(EpochSupplier epochSupplier, BiConsumer<Void, Throwable> c
 
     public void withEpoch(long epoch, BiConsumer<Void, Throwable> callback)
     {
+        if (epoch < topology.minEpoch())
```
belliottsmith, line 396: I think we maybe need to replace `withEpoch` with `withEpochAtLeast` and `withEpochExact`. Or maybe we need to add an additional parameter to confirm the ranges we want. This would probably be annoying, but we already have some inconsistency here - one method checks `hasEpoch` the other `hasAtLeastEpoch`, and in one case it would be in principle fine to not `hasEpoch` anymore and we probably _don't_ want to fail in this case if we otherwise have enough information.

Alternatively, we just pick one or the other behaviour (atLeast or has), and we rely on the caller to simply fail if that isn't enough information. 

So... perhaps for now for path of least resistance we _introduce_ both methods, but we _default to_ hasEpochAtLeast and migrate as we find time to a variant of hasEpoch which can supply Unseekables to verify exist in the event the epoch has been removed.

accord-core/src/main/java/accord/primitives/AbstractRanges.java
```diff
@@ -127,6 +127,11 @@ public final boolean containsAll(AbstractRanges that)
         return ((int) supersetLinearMerge(this.ranges, that.ranges)) == that.size();
     }
 
+    public boolean contains(Range range)
+    {
+        return ((int) supersetLinearMerge(this.ranges, new Range[] { range} )) == 1;
```
belliottsmith:
```suggestion
            return indexOf(range, FAST) >= 0;
```

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -231,8 +254,10 @@ static class Notifications
             Ranges closed = Ranges.EMPTY, retired = Ranges.EMPTY;
         }
 
-        private static final Epochs EMPTY = new Epochs(new EpochState[0]);
+        private static final Epochs EMPTY = new Epochs(new EpochState[0], new ArrayList<>(), new ArrayList<>(), -1);
```
belliottsmith: Collections.emptyList?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -754,9 +845,21 @@ public long epoch()
         return current().epoch;
     }
 
+    public long firstNonEmpty()
```
belliottsmith, line 867: unused - presumably for integration, should we be using it in burn test? Could leave TODO to expand testing.

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -231,8 +254,10 @@ static class Notifications
             Ranges closed = Ranges.EMPTY, retired = Ranges.EMPTY;
         }
 
-        private static final Epochs EMPTY = new Epochs(new EpochState[0]);
+        private static final Epochs EMPTY = new Epochs(new EpochState[0], new ArrayList<>(), new ArrayList<>(), -1);
         private final long currentEpoch;
+        private final long firstNonEmptyEpoch;
```
belliottsmith, line 263: do we need to precompute this? can we calculate it on the fly? I think we don't query it often

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -258,12 +290,45 @@ private Epochs(EpochState[] epochs, List<Notifications> pending, List<FutureEpoc
                 Invariants.requireArgument(futureEpochs.get(i).epoch == futureEpochs.get(i - 1).epoch + 1);
             for (int i = 1; i < epochs.length; i++)
                 Invariants.requireArgument(epochs[i].epoch() == epochs[i - 1].epoch() - 1);
-            this.epochs = epochs;
-        }
+            int truncateFrom = -1;
```
belliottsmith, line 297: Maybe this can be in a specific method, rather than being imposed in the constructor? Would perhaps be clearer to invoke it deliberately, even if we do it in all the same places.

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -754,9 +845,21 @@ public long epoch()
...
+    }
+
     public long minEpoch()
     {
-        return epochs.minEpoch();
+        Epochs epochs = this.epochs;
+        long minEpoch = epochs.minEpoch();
+
+        // We _never_ consider current epoch retired
+        while (minEpoch < epochs.currentEpoch && epochs.get(minEpoch).allRetired())
```
belliottsmith: maybe introduce a new `minLiveEpoch()`?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -892,8 +1055,11 @@ private <C, K extends Routables<?>, T> T atLeast(K select, long minEpoch, long m
 
         if (i == snapshot.epochs.length)
         {
-            if (!select.isEmpty())
-                throw new IllegalArgumentException("Ranges " + select + " could not be found");
+            // Epochs earlier than minEpoch might have been GC'd, so we can not collect
+            // matching ranges for them. However, if ranges were still present in the min epoch,
+            // we have reported them.
+            if (!select.isEmpty() && !select.without(snapshot.get(minEpoch).global.ranges).isEmpty())
```
belliottsmith, line 1078: maybe move to `Invariants.requireArgument`, or creating the exception with `Invariants.illegalArgument`?

accord-core/src/main/java/accord/local/Node.java
ifesdjeen: I have switched both paths to `atLeast` variant, I personally think this is what is expected in both cases. Also, `hasEpoch` and `hasEpochAtLeast` are more or less synonymous, since we already throw retired exception.

accord-core/src/main/java/accord/topology/TopologyManager.java
ifesdjeen: talked off-github: we can't avoid precomputing this field, since it is a "historical" information. In other words, we can not infer it form `Epochs`. 

Tried moving it to a different class, but unfortunately alternatives are only worse than the current version.

accord-core/src/main/java/accord/local/Node.java
belliottsmith: We might need to be careful about using retired, or perhaps we at least need to offer a two param option for messages, as we ask for an earlier epoch than we actually need - passing the min epoch with expect the node to have witnessed but executing in a later epoch. This only affects message arrival, so should be an easy edge case to address by providing the actual intended epoch for testing against retired

PR #175, opened 2025-03-02
Rx repair

ifesdjeen:


accord-core/src/test/java/accord/coordinate/CoordinateTransactionTest.java
```diff
@@ -164,152 +130,6 @@ void exclusiveSyncTest() throws Throwable
         }
     }
 
-    @Test
-    void barrierTest() throws Throwable
```
ifesdjeen, line 168: Was this one not very useful, or just hard to maintain?

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
```diff
@@ -0,0 +1,375 @@
...
+
+import static accord.coordinate.ExecuteSyncPoint.coordinateIncluding;
+import static accord.local.durability.DurabilityService.SyncRemote.All;
+import static java.util.concurrent.TimeUnit.MICROSECONDS;
+import static java.util.concurrent.TimeUnit.MINUTES;
+
+/**
+ * Helper methods and classes to invoke coordination to propagate information about durability.
+ *
+ * Both CoordinateShardDurable and CoordinateGloballyDurable use wall time to loosely coordinate between nodes
```
ifesdjeen: Looks like this comment got mostly copied from `ShardDurability`

accord-core/src/main/java/accord/local/durability/ShardDurability.java
```diff
@@ -0,0 +1,672 @@
...
+ * so that they take non-overlapping (hopefully) turns doing the coordination.
+ *
+ * Both of these have a concept of rounds where rounds have a known duration in wall time, and the current round is known
+ * based on time since the epoch, and the point in time where a node should do something in a given round is known based
+ * on its index in the sorted list of nodes ids in the current epoch.
+ *
+ * Coordinate globally durable is simpler because they just need to take turns so nodes just calculate when it is their
+ * turn and invoke CoordinateGloballyDurable.
+ *
+ * CoordinateShardDurable needs nodes to not overlap on the ranges they operate on or the exlusive sync points overlap
```
ifesdjeen: nit: exlusive

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
```diff
@@ -0,0 +1,375 @@
...
+                return true;
+        }
+        return false;
+    }
+
+    private void registerInProgress(SyncPoint<Range> syncPoint, AsyncResult<DurabilityResult> submitted)
+    {
+        inProgress.add(submitted);
+        for (Range range : syncPoint.route)
+            inProgressRanges.put(range.start(), range.end());
```
ifesdjeen, line 140: nit: maybe add invariant that previous value is null here?

accord-core/src/main/java/accord/local/durability/ConcurrencyControl.java
```diff
@@ -0,0 +1,90 @@
...
+            Pending<?> next = pending.poll();
+            if (next != null)
+            {
+                ++inflight;
+                next.start(this);
+            }
+        }
+    }
+
+    public synchronized <V> AsyncChain<V> submit(Supplier<? extends AsyncChain<V>> supplier)
```
ifesdjeen: nit/optional: probably unimportant but we could limit the syncronized scope here a bit by reading and incrementing inflight if required, and doing the rest outside it

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
```diff
@@ -0,0 +1,375 @@
...
+        AgentExecutor executor = node.commandStores().someExecutor();
+        if (executor != null && inProgress.size() < maxConcurrency && !isInProgress(syncPoint.route))
+        {
+            start(syncPoint, request, attempt, executor);
+        }
+        else
+        {
+            ++pendingCounter;
+            pending.add(new Pending(syncPoint, request, attempt));
+            if (pending.size() > 100 && pendingCounter > prunedAt + pending.size())
```
ifesdjeen: nit: maybe put it to final constant in the header?

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
```diff
@@ -0,0 +1,375 @@
...
+    {
+        logger.debug("{}: Awaiting durability for {}", exclusiveSyncPoint.syncId, exclusiveSyncPoint.route.toRanges());
+        AsyncResult<DurabilityResult> coordinate = coordinateIncluding(node, exclusiveSyncPoint, request == null ? null : request.including, executor, attempt);
+        registerInProgress(exclusiveSyncPoint, coordinate);
+        if (request != null)
+            request.reportAttempt(exclusiveSyncPoint.syncId, node.elapsed(MICROSECONDS), coordinate);
+
+        coordinate.addCallback((success, fail) -> {
+            inProgress.remove(coordinate);
+            unregisterInProgress(exclusiveSyncPoint, coordinate);
```
ifesdjeen: `inProgress.remove(coordinate);` is called twice: once in the callback inself, and once in the `unregisterInProgress`, probably only the latter one suffices.

accord-core/src/main/java/accord/local/durability/GlobalDurability.java
```diff
@@ -0,0 +1,237 @@
...
+        {
+            logger.warn("Exception invoking withEpoch to start coordination for global durability", e);
+        }
+    }
+
+    synchronized void updateTopology(Topology latestGlobal)
+    {
+        Invariants.require(currentGlobalTopology == null || latestGlobal.epoch() > currentGlobalTopology.epoch());
+        currentGlobalTopology = latestGlobal;
+        List<Node.Id> ids = new ArrayList<>(latestGlobal.nodes());
```
ifesdjeen, line 179: Should we check if our global index has changed and cancel outstanding job, or is it OK to finish the current round?

accord-core/src/main/java/accord/utils/SortedListSet.java
```diff
@@ -0,0 +1,195 @@
...
+ * limitations under the License.
+ */
+
+package accord.utils;
+
+import java.util.AbstractSet;
+import java.util.Iterator;
+
+// TODO (expected): make a hashing version of this class; have Node.Id support both approaches
+public abstract class SortedListSet<K extends Comparable<? super K>> extends AbstractSet<K>
```
ifesdjeen, line 27: Would you be opposed to adding this test?

```
    @Test
    public void setTest()
    {
        for (int iteration = 0; iteration < 1000; iteration++)
        {
            RandomTestRunner.test().check(rs -> {
                List<Integer> input = new ArrayList<>();
                int size = rs.nextInt(1, 10);
                for (int i = 0; i < size; i++)
                {
                    int v = rs.nextInt();
                    input.add(v);
                }

                input.sort(Integer::compare);
                Integer[] sorted = input.toArray(new Integer[0]);
                SortedArrays.assertSorted(sorted);
                SortedListSet<Integer> sut = SortedListSet.create(new SortedArrays.SortedArrayList<>(sorted));
                Set<Integer> model = new HashSet<>();
                for (int i = 0; i < 10_000; i++)
                {
                    int v = input.get(rs.nextInt(input.size()));
                    assertThat(model.size()).isEqualTo(sut.size());
                    assertThat(model.contains(v)).isEqualTo(sut.contains(v));
                    assertThat(setFromIter(model.iterator())).isEqualTo(setFromIter(sut.iterator()));
                    if (rs.nextBoolean())
                    {
                        model.add(v);
                        sut.add(v);
                    }
                    else
                    {
                        model.remove(v);
                        sut.remove(v);
                    }
                    assertThat(model.size()).isEqualTo(sut.size());
                    assertThat(model.contains(v)).isEqualTo(sut.contains(v));
                    assertThat(setFromIter(model.iterator())).isEqualTo(setFromIter(sut.iterator()));
                }
            });
        }
    }
```

also, currently API is a bit confusing: usually if you pass a collection during set creation, all of its elements become contained. I understand that this class's intention is more like one of `EnumSet`, so this does not apply. Maybe close the constructor and add a factory method with a different name?

accord-core/src/main/java/accord/utils/SortedListSet.java
```diff
@@ -0,0 +1,195 @@
...
+        public void remove()
+        {
+            if (cur >= 0 && unset(cur))
+                --size;
+        }
+
+        @Override
+        public boolean hasNext()
+        {
+            while (++cur < list.size())
```
ifesdjeen: if `hasNext` is called more than once, we'll advance the iterator inadvertently.

accord-core/src/main/java/accord/coordinate/SynchronousAwait.java
```diff
@@ -0,0 +1,97 @@
...
+    public static AsyncChain<Boolean> awaitQuorum(Node node, Topologies topologies, TxnId txnId, BlockedUntil blockedUntil, boolean notifyProgressLog, Participants<?> participants)
+    {
+        // TODO (expected): copy this pattern elsewhere; should also make it easier to share exception handling logic etc
+        return new AsyncChains.Head<>()
+        {
+            protected @Nullable @Override Cancellable start(BiConsumer<? super Boolean, Throwable> callback)
+            {
+                SynchronousAwait await = new SynchronousAwait(new QuorumTracker(topologies), callback);
+                node.send(topologies.nodes(), to -> new Await(to, topologies, txnId, participants, blockedUntil, notifyProgressLog), await);
+                return null;
```
ifesdjeen, line 67: do we want to do anything with the result of `await` once its set? Or is it intended as fire-and-forget? I'm a bit confused about the name of the class if so.

ifesdjeen:
+1 

I think we should think deeply about how we want to test durability scheduling. I tried disabling it entirely and burn test would still pass (which makes sense), so we need to find other means, perhaps ensuring that we do make progress on RX in general, and that nodes do manage to satisfy their slots, and have invariants that would notify us about this.

Also, I think we may benefit from more localized testing of durability scheduling: a simple test where given a topology, we ensure that nodes' participation is non-overlapping, and that step-taking logic is correct, outside of burn test.

But I do not think it would be reasonable to block this patch on either one of two points.

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
```diff
@@ -0,0 +1,375 @@
...
+ * responsible for rotates backwards every round so that a down node doesn't prevent a subrange from being processed.
+ *
+ * The work for CoordinateShardDurable is further subdivided where each subrange a node operates on is divided a fixed
+ * number of times and then processed one at a time with a fixed wait between them.
+ *
+ * TODO (expected): cap number of coordinations we can have in flight at once
+ * TODO (expected): do not start new ExclusiveSyncPoint if we have more than X already agreed and not yet applied
+ * Didn't go with recurring because it doesn't play well with async execution of these tasks
+ */
+public class DurabilityQueue
```
ifesdjeen, line 67: If you find it useful, here's what I left in my notes:

```
/**
 * Durability queue tracks and schedules exclusive durability requests after sync point is complete.
 *
 * Up to {@code maxConcurrency} tasks are allowed to be in flight. All subsequent tasks are put into the {@code pending}
 * queue, which new tasks are pulled from after the current in-progress ones complete. To prevent unbounded growth of the
 * pending queue, pending tasks are periodically pruned.
 */
```

accord-core/src/test/java/accord/coordinate/CoordinateTransactionTest.java
belliottsmith: It didn't make much sense after removing the BarrierTransaction concept, and was hard to port to an equivalent.

accord-core/src/main/java/accord/coordinate/SynchronousAwait.java
belliottsmith: There's a callback. I'm not sure why we're returning null though.

accord-core/src/main/java/accord/utils/SortedListSet.java
belliottsmith: Yep, as we discussed offline I'll add EnumSet like `noneOf` and `allOf` methods for initialisation.

accord-core/src/main/java/accord/local/durability/GlobalDurability.java
belliottsmith: global durability doesn't contend or cause any problems, so this is completely best effort just improve the distribution and timeliness of work

accord-core/src/main/java/accord/coordinate/SynchronousAwait.java
belliottsmith: Oh right yes, we aren't Cancellable. So yes, this is all correct. We pass through the callback and leave Await to invoke it.

belliottsmith:
> So we need to find other means, perhaps ensuring that we do make progress on RX in general, and that nodes do manage to satisfy their slots

If it was properly disabled, we should have seen topologies all get snarled up, so perhaps we want to add some validation that this doesn't happen as well. But, I agree we probably also want some validation of the request logic, both in the burn test and separately. I have added a TODO to the long tail.

PR #176, opened 2025-03-03
CASSANDRA-20390: Improve partial compaction

ifesdjeen:


accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -109,33 +110,44 @@ public enum Field
...
         {
             SaveStatus saveStatus = SaveStatus.forOrdinal(i);
+
             int mask = 0;
             if (forceFieldChangedToNullFlag(saveStatus, saveStatus.known::is, DepsErased))
-                mask |= setFieldIsNullAndChanged(PARTIAL_DEPS, mask)
-                     |  setFieldIsNullAndChanged(WAITING_ON, mask)
-                     |  setFieldIsNullAndChanged(MIN_UNIQUE_HLC, mask);
+                mask |= setIsNullAndChanged(PARTIAL_DEPS, mask)
+                        | setIsNullAndChanged(WAITING_ON, mask)
```
belliottsmith, line 135: nit: alignment

accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -359,6 +397,8 @@ public Command construct(RedundantBefore redundantBefore)
             if (this.waitingOn != null)
                 waitingOn = this.waitingOn.provide(txnId, partialDeps, executesAtLeast, minUniqueHlc);
 
+            if (saveStatus == null)
```
belliottsmith, line 400: cleanup before commit

accord-core/src/test/java/accord/impl/basic/InMemoryJournal.java
```diff
@@ -308,29 +325,77 @@ public void purge(CommandStores commandStores)
...
-                        e2.setValue(new PurgedList(e2.getValue()));
-                        break;
+                    Invariants.require(cleanup != NO);
+                    if (cleanup == EXPUNGE)
+                        e2.setValue(new PurgedList());
+                    else
+                    {
+                        int flags = CommandChange.flagsForPurge(cleanup.appliesIfNot);
+                        EnumMap<Field, Object> values = new EnumMap<>(Field.class);
+                        for (Field field : Field.values())
```
belliottsmith, line 369:
```suggestion
                        int iterator = toIterableSetFields(~flags);
                        for (Field field = nextSetField(iterator); field != null; iterator = unsetIterable(field, iterator), field = nextSetField(iterator))
                        {
                            if (field == CLEANUP || field == SAVE_STATUS)
                                continue;
```

accord-core/src/test/java/accord/impl/basic/InMemoryJournal.java
```diff
@@ -308,29 +325,77 @@ public void purge(CommandStores commandStores)
...
-
-                    case EXPUNGE:
-                        e2.setValue(new PurgedList(e2.getValue()));
-                        break;
+                    Invariants.require(cleanup != NO);
+                    if (cleanup == EXPUNGE)
+                        e2.setValue(new PurgedList());
+                    else
+                    {
+                        int flags = CommandChange.flagsForPurge(cleanup.appliesIfNot);
```
belliottsmith, line 367: since we're calling setChanged(SAVE_STATUS, flags) below, I think we can simply use the `eraseKnownFieldsMask` directly?

accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -503,7 +543,7 @@ public static int getFlags(Command before, Command after)
         }
 
         if (after.saveStatus() != null)
```
belliottsmith, line 545: not changed in this patch, but I think this condition is unnecessary as it is always non-null?

PR #179, opened 2025-03-10
Fix serializers

ifesdjeen:


accord-core/src/main/java/accord/utils/SortedArrays.java
```diff
@@ -447,6 +447,93 @@ else if (cmp < 0)
...
+                }
+
+                if (cmp < 0) rightKey = null;
+                else ++rightIdx;
+                ++leftIdx;
+
+                result[resultSize++] = update.apply(leftKey, rightKey);
+            }
+
+            while (leftIdx < leftEnd)
```
ifesdjeen: Just to make sure: do we always know that left is longer than right?

ifesdjeen (commented):
+1

accord-core/src/main/java/accord/utils/SortedArrays.java
belliottsmith: we don't, we just don't care if right is longer than left - this is meant to be like a left join; we only care about those items on the right that match those on the left.

PR #181, opened 2025-03-13
Cleanup mar25

belliottsmith:


accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -329,50 +341,142 @@ public Cleanup maybeCleanup(Input input, Cleanup cleanup)
...
+            if (addCleanup == null || addCleanup == NO)
+                return false;
+
+            if (cleanup != null && addCleanup.compareTo(cleanup) <= 0)
+                return false;
+
+            if (saveCleanup)
+                cleanup = addCleanup;
+
+            if (saveStatus != null && saveStatus.compareTo(cleanup.appliesIfNot) >= 0)
```
ifesdjeen: can't we end up with null for `cleanup`? (full disclosure: Intellij has highlighted that cleanup might be null, but it seems to be a valid point)

accord-core/src/test/java/accord/impl/basic/InMemoryJournal.java
```diff
@@ -331,31 +330,50 @@ public void purge(CommandStores commandStores, EpochSupplier minEpoch)
...
+                    builder.apply(subset.get(i));
+                    builders[i] = builder;
+                }
 
-                if (builder.saveStatus().status == Truncated || builder.saveStatus().status == Invalidated)
-                    continue; // Already truncated
+                Builder builder = new Builder(e2.getKey(), ALL);
+                for (int i = builders.length - 1; i >= 0 ; --i)
+                {
+                    Builder diffBuilder = builders[i];
```
ifesdjeen: nit: maybe rename to `current`?

ifesdjeen (approved):

accord-core/src/main/java/accord/impl/CommandChange.java
belliottsmith: Hmm, not sure what's going on in UX, but this is not up to date - we always assign cleanup now, so this is not null

PR #182, opened 2025-03-14
Cassandra 20443

dcapwell:


accord-core/src/main/java/accord/primitives/Keys.java
```diff
@@ -94,7 +94,10 @@ public Keys slice(Ranges ranges, Slice slice)
     @Override
     public final boolean containsAll(AbstractUnseekableKeys that)
     {
-        return that.size() == SortedArrays.foldlIntersection(0, RoutableKey::compareAsRoutingKey, that.keys, 0, that.keys.length, keys, 0, keys.length, (k, p, v, l, r) -> v + 1, 0, 0, 0);
+        return that.size() == SortedArrays.foldlIntersection(0, RoutableKey::compareAsRoutingKey,
+                                                             keys, 0, keys.length,
+                                                             that.keys, 0, that.keys.length,
```
dcapwell, line 99: swapped keys and that.keys

accord-core/src/main/java/accord/primitives/AbstractKeys.java
```diff
@@ -304,17 +304,10 @@ interface ToRoutingKeysFactory<T>
 
     protected <T> T toRoutingKeysArray(RoutingKey withKey, boolean permitInsert, ToRoutingKeysFactory<T> toRoutingKeysFactory)
     {
-        RoutingKey[] copy;
-        if (keys.getClass() == RoutingKey[].class)
-        {
-            copy = (RoutingKey[])keys;
```
belliottsmith: why deleting this branch?

belliottsmith (changes requested):
Only change the second branch in AbstractKeys, as the first is safe if the current object has been constructed correctly, and avoids unnecessary allocations. Otherwise +1, feel free to commit once sorted.

accord-core/src/main/java/accord/primitives/AbstractKeys.java
dcapwell: I’ll revert

belliottsmith (approved):

PR #183, opened 2025-03-15
Cleanup 15mar25

belliottsmith:


dcapwell:
starting review but there is a lot going on and don't see a JIRA that i can read more information about, i just have the description `Cleanup 15mar25`

How does this patch relate to the SSTable issue that its supposed to fix, and what is done that is unrelated?

accord-core/src/main/java/accord/api/Journal.java
```diff
@@ -42,6 +43,8 @@
  */
 public interface Journal
 {
+    Journal start(Node node);
```
dcapwell: Nothing consumes the result of this function, maybe drop `Journal` in favor of `void`?  `void` makes the semantics clear where as `Journal` "could" be a different version.

Screenshot 2025-03-27 at 3 23 11 PM

accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -90,18 +91,18 @@ public enum Field
     {
         PARTICIPANTS, // stored first so we can index it
         SAVE_STATUS,
-        PARTIAL_DEPS,
```
dcapwell: is there a reason to shuffle this around?  confirmed nothing was lost, but journal does rely on `field.ordinal()` which makes this a breaking change; so would like to know why we should reorder the fields

accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -226,17 +229,17 @@ public Object get(Field field)
         {
             switch (field)
             {
-                case EXECUTE_AT: return executeAt;
```
dcapwell: was there a reason for this refactor?  looks like nothing was lost, but also looks arbitrary?

accord-core/src/test/java/accord/impl/basic/LoggingJournal.java
```diff
@@ -73,6 +74,13 @@ private synchronized void log(String format, Object... objects)
         }
     }
 
+    @Override
+
+    public Journal start(Node node)
+    {
+        return this;
+    }
```
dcapwell:
```suggestion
    @Override
    public Journal start(Node node)
    {
        delegate.start(node);
        return this;
    }
```

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -701,8 +701,8 @@ public void markShardDurable(SafeCommandStore safeStore, TxnId globalSyncId, Ran
     protected void updatedRedundantBefore(SafeCommandStore safeStore, TxnId syncId, Ranges ranges)
     {
         TxnId clearWaitingBefore = redundantBefore.minShardAndLocallyAppliedBefore();
-        TxnId clearAnyBefore = durableBefore().min.majorityBefore;
-        progressLog.clearBefore(safeStore, clearWaitingBefore, clearAnyBefore);
+        TxnId clearAllBefore = TxnId.min(clearWaitingBefore, durableBefore().min.majorityBefore);
```
dcapwell: why push this change here?  it looked to happen in default progress log before now its an assert and the logic moved here.

accord-core/src/main/java/accord/local/cfk/CommandsForKey.java
```diff
@@ -1603,35 +1603,33 @@ private CommandsForKeyUpdate update(InternalStatus newStatus, Command updated, @
         {
             // update
             TxnInfo cur = byId[pos];
-            if (cur != null)
```
dcapwell, line 1606: for me: 

`pos >= 0` so its known that the field isn't null, so this just removes the `null` check; i couldn't find any lines that differ, just removed the if

accord-core/src/main/java/accord/coordinate/tracking/FastPathTracker.java
```diff
@@ -116,7 +116,7 @@ public final ShardOutcome<? super FastPathTracker> onDelayed(@Nonnull Node.Id fr
                 ++fastPathDelayed;
 
                 if (isFastPathDelayed() && hasReachedQuorum())
-                    return complete(Success);
+                    return complete(hasMetMediumPathCriteria() ? NewMediumPathSuccess : Success);
```
dcapwell: this logic is duplicated in `accord.coordinate.tracking.FastPathTracker.FastPathShardTracker#quorumIfHasRejectedFastPath`

can we cleanup so we don't have it scattered?

dcapwell: What about `accord.coordinate.tracking.FastPathTracker.FastPathShardTracker#onFailure`?  It had the same logic but doesn't include `NewMediumPathSuccess`, why is that?

belliottsmith:
The commit messages provide brief descriptions of the problems addressed

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: It’s clearer to have the caller declare what they actually want rather than have the callee reinterpret, which can be surprising

accord-core/src/test/java/accord/impl/basic/LoggingJournal.java
belliottsmith: I’ll have to look at this again next week. I think @ifesdjeen may have said this startup sequencing issue (I forgot the precise problem) may have been addressed in another patch of his anyway.

accord-core/src/main/java/accord/impl/CommandChange.java
belliottsmith: To make the field order consistent throughout this class and also broadly consistent with the rest of the codebase (except where otherwise necessary for efficiency, ie participants)

accord-core/src/main/java/accord/impl/CommandChange.java
belliottsmith: To make the field order consistent throughout this class and also broadly consistent with the rest of the codebase (except where otherwise necessary for efficiency, ie participants)

Nothing has been released yet, so this is not a breaking change, which is why it is important to make this kind of consistency change before that happens

accord-core/src/main/java/accord/api/Journal.java
belliottsmith: I’ll have to look at this again next week. I think @ifesdjeen may have said this startup sequencing issue (I forgot the precise problem) may have been addressed in another patch of his anyway.

belliottsmith:
The sstable issue is addressed by the change to not use InfiniteLoopExecutor which must signal using interrupts.

accord-core/src/main/java/accord/coordinate/tracking/FastPathTracker.java
belliottsmith: > why is that

I think an oversight; have fixed. Have also factored out the logic to do this as requested.

accord-core/src/main/java/accord/coordinate/ExecuteSyncPoint.java
```diff
@@ -249,6 +249,6 @@ protected void start()
     {
         SortedArrayList<Node.Id> contact = tracker.filterAndRecordFaulty();
         if (contact == null) tryFailure(new Exhausted(syncPoint.syncId, syncPoint.route.homeKey(), null));
-        else node.send(contact, to -> new WaitUntilApplied(to, tracker.topologies(), syncPoint.syncId, syncPoint.route, syncPoint.syncId.epoch()), executor, this);
+        else node.send(contact, to -> new WaitUntilApplied(to, tracker.topologies(), syncPoint.syncId, syncPoint.route, tracker.topologies().currentEpoch()), executor, this);
```
dcapwell, line 252: why this change?  `currentEpoch` is the `execute_at` epoch where as syncId.epoch was txn_id right?  this is sync so shouldn't `execute_at == txn_id`?

belliottsmith: To enforce the invariant that all transactions with lower txnid have executed we sometimes have to visit future epochs to witness earlier txnid executed.

dcapwell (approved):

PR #184, opened 2025-04-02
CASSANDRA-20520: Improve debuggability of AsyncChain calls

ifesdjeen:


accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -178,6 +222,54 @@ public void accept(V v, Throwable throwable)
             // we implement here just to simplify logic a little
             throw new UnsupportedOperationException();
         }
+
+        /**
+         * A method to be called upon Head construction to wrap it into a specialized Debug object that will
+         * preserve the call site  for async chain debugging.
+         */
+        public Head<V> maybeWrapDebug()
```
belliottsmith, line 230: does this need to be public? could it be private static?

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -226,6 +336,41 @@ public void accept(I i, Throwable throwable)
         }
     }
 
+    private static abstract class DebugMap<I, O> extends Link.Debug<I, O> implements Function<I, O>
```
belliottsmith, line 339: would it be better to extend `Map` and have a `Debug` interface we implement?

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -253,8 +415,37 @@ public static abstract class FlatMap<I, O> extends Link<I, O> implements Functio
...
+                else                   apply(i).begin(next);
+            }
+            catch (Throwable t)
+            {
+                throw t;
+            }
+        }
+    }
+
+    private static abstract class DebugFlatMap<I, O> extends Link.Debug<I, O> implements Function<I, AsyncChain<O>>
```
belliottsmith, line 430: similarly: wouldn't it be better to extend `FlatMap` since it has more bespoke logic we may modify in future and want to inherit any changes?

belliottsmith: it might even be better to just have a single DebugWrapper object, that wraps all of the results of each call? We pay an additional object overhead per link but only when debug is enabled, and don't have to extend anything?

PR #187, opened 2025-04-23
Add testing of consensus live migration to simulator

aweisberg:
Patch by Ariel Weisberg; Reviewed by ??? for CASSANDRA-20587

### DIFF
--- a/accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
+++ b/accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
@@ -223,8 +223,11 @@ public class ExecuteTxn extends ReadCoordinator<ReadReply>
                 executeAt = new TimestampWithUniqueHlc(executeAt, uniqueHlc);
             }
 
-            Writes writes = txnId.is(Txn.Kind.Write) ? txn.execute(txnId, executeAt, data) : null;
+            // Always compute Result before Write to provide integrations with a predictable invocation order
+            // in case there is shared state between Result and Update. This can change if really needed
+            // just make sure to check the integrations to make sure it won't break anything
             Result result = txn.result(txnId, executeAt, data);
+            Writes writes = txnId.is(Txn.Kind.Write) ? txn.execute(txnId, executeAt, data) : null;
             adapter().persist(node, executor, allTopologies, route, ballot, flags, txnId, txn, executeAt, stableDeps, writes, result, callback);
         }
         else
--- a/accord-core/src/main/java/accord/local/CommandStores.java
+++ b/accord-core/src/main/java/accord/local/CommandStores.java
@@ -819,6 +819,10 @@ public abstract class CommandStores implements AsyncExecutorFactory
         Snapshot snapshot = current;
         for (ShardHolder shard : snapshot.shards)
             list.add(shard.store.build(empty(), forEach));
+
+        if (list.isEmpty())
+            return AsyncResults.success(null);
+
         return AsyncChains.reduce(list, Reduce.toNull());
     }
 
--- a/accord-core/src/main/java/accord/local/Node.java
+++ b/accord-core/src/main/java/accord/local/Node.java
@@ -711,6 +711,11 @@ public class Node implements ConfigurationService.Listener, NodeCommandStoreServ
         return nextTxnId(rw, domain, cardinality, defaultMediumPath().bit());
     }
 
+    public TxnId nextTxnId(long minHlc, Txn.Kind rw, Domain domain, Cardinality cardinality)
+    {
+        return newTxnId(epoch(), uniqueNow(minHlc), rw, domain, cardinality, defaultMediumPath().bit(), id);
+    }
+
     public TxnId nextTxnId(Timestamp min, Txn.Kind rw, Domain domain, Cardinality cardinality)
     {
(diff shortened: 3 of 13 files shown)


accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
```diff
@@ -236,8 +236,9 @@ protected void onDone(Success success, Throwable failure)
                 executeAt = new TimestampWithUniqueHlc(executeAt, uniqueHlc);
             }
 
-            Writes writes = txnId.is(Txn.Kind.Write) ? txn.execute(txnId, executeAt, data) : null;
+            // Always compute Result before Write
```
belliottsmith: Explain why?

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -625,6 +625,10 @@ public AsyncChain<Void> forEach(Consumer<SafeCommandStore> forEach)
         Snapshot snapshot = current;
         for (ShardHolder shard : snapshot.shards)
             list.add(shard.store.build(empty(), forEach));
+
+        if (list.isEmpty())
```
belliottsmith, line 823: I wonder if we should remove the variants of `AsyncChains.reduce` that don't accept `identity` and use this version here. It's a bit brittle, the lack of properly handling no inputs.

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -894,6 +900,13 @@ public <R> CoordinationAdapter<R> coordinationAdapter(TxnId txnId, Kind kind)
         return coordinationAdapters.get(txnId, kind);
     }
 
+    public void updateMinHlc(long minHlc)
+    {
+        commandStores().forEach(safeStore -> {
```
belliottsmith, line 927: do we care that this won't affect new command stores that are added later?

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -687,6 +687,12 @@ public TxnId nextTxnId(Txn.Kind rw, Domain domain, Cardinality cardinality)
         return nextTxnId(rw, domain, cardinality, defaultMediumPath().bit());
     }
 
+    public TxnId nextTxnId(Txn.Kind rw, Domain domain, Cardinality cardinality, long minHlc)
+    {
+        Timestamp minTimestamp = Timestamp.fromValues(topology.epoch(), minHlc, 0, id());
```
belliottsmith: might be cleaner to invoke `newTxnId` directly with the epoch and uniqueNow(minHlc)?

belliottsmith: might also be more consistent to accept minHlc as first parameter to match Timestamp min variant below? Or could perhaps introduce a nextTxnId(minEpoch, minHlc, ... to accept both of these variants

accord-core/src/main/java/accord/local/SafeCommandStore.java
```diff
@@ -303,6 +305,11 @@ public void updateMaxConflicts(Command prev, Command updated)
         commandStore().updateMaxConflicts(prev, updated);
     }
 
+    public long maxHLCForSeekables(Routables<?> keysOrRanges)
```
belliottsmith: simply `maxHlc`? You're accepting `Routables` not `Seekables` and including the type isn't probably necessary anyway. Mostly we have lower cased the `lc` in `Hlc` so far in the codebase.

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
aweisberg: It makes it easier for the integration to do things like evaluate part of the result and logic related to what will happen. The order was switched and that caused the integration to apply transactions when the condition failed. I fixed it so that the integration will tolerate either order, but since this is an easy bug to re-introduce and there isn't a huge cost in saying that `Result` and `Write` will be evaluated in a specific order that it would be better to do that.

belliottsmith: I meant in the code comment for posterity, but thanks I did also want to know 🙂

accord-core/src/main/java/accord/local/CommandStores.java
aweisberg: I think that makes sense to an extent, but it does require that the reducer accept `null` and handle it by returning the non-null parameter? Or for everything to have a non-null identity that can be handled correctly. WDYT?

accord-core/src/main/java/accord/local/Node.java
aweisberg: Looking into this now, but yes it is a problem if new ones are constructed without getting the min HLC from somewhere.

accord-core/src/main/java/accord/local/SafeCommandStore.java
aweisberg: Unused, removed

accord-core/src/main/java/accord/local/Node.java
aweisberg: I checked and it is updating the min HLC from Paxos state tracker as part of updating ranges for epoch in the Accord integration so newly constructed command stores always get it.

PR #188, opened 2025-04-28
CASSANDRA-20608: Get accord compiling on JDK 21

dcapwell:


accord-core/src/main/java/accord/utils/btree/BTreeSet.java
```diff
@@ -34,7 +33,7 @@
 import static accord.utils.btree.BTree.Dir;
 import static accord.utils.btree.BTree.findIndex;
 
-public class BTreeSet<V> extends AbstractSet<V> implements NavigableSet<V>, List<V>
+public class BTreeSet<V> extends AbstractSet<V> implements NavigableSet<V>
```
dcapwell, line 36: `Set` and `List` conflict, so can't be both

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -16,7 +16,7 @@
 
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-bin.zip
```
dcapwell, line 19: 8.0 doesn't work on jdk21, but 8.5 does

maedhroz (approved):

PR #189, opened 2025-05-02
CASSANDRA-20577: add new utility method based off review feedback

dcapwell:


accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -944,6 +944,19 @@ public CommandsBuilder<State, SystemUnderTest> addIf(Predicate<State> predicate,
             return this;
         }
 
+        @SafeVarargs
+        public final CommandsBuilder<State, SystemUnderTest> addAllIf(Predicate<State> predicate,
```
yifan-c, line 948: To me, `addAllIf` is hard to understand. My mental model when processing it reads as 
```
if (predicate) 
{
   first;
   second;
   rest...;
}
```
So, maybe just `addIf`. 
I guess you mean "add all the if statements". Since they are using the same predicate, it is really one single `if` statement.

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -944,6 +944,19 @@ public CommandsBuilder<State, SystemUnderTest> addIf(Predicate<State> predicate,
...
+                                                                      Setup<State, SystemUnderTest> first,
+                                                                      Setup<State, SystemUnderTest> second,
+                                                                      Setup<State, SystemUnderTest>... reset)
```
yifan-c: `rest`?

dcapwell: the reason for this style is to force you to use `addIf` when there is only 1 `Setup`, so can only call if-and-only-if there are at least 2.

accord-core/src/test/java/accord/utils/Property.java
dcapwell: > My mental model when processing it reads as

That is what it is doing, but it doesn't do it right away, it does this later on.

`addIf(predicte, stmt)`. this doesn't add the `stmt` to the builder if the predicate is true.  What this actually does is the following (logically)

```
conditionals.put(predicate, stmt); // multiple stmt can share the same predicate
```

then when the stateful builder executes it calls `public Gen<Command<State, SystemUnderTest, ?>> commands(State state) throws Throwable`

What we do here is 

```
Map<Gen, Weight> weights = ...
weights.addAll(unconditional);
for (predicate, stmts) in conditionals:
  if predicate.test(state):
    stmts.forEach(stmt ->     weights.put(stmt, weightGen.next(rs)));

Gen gen = Gens.pick(weights);
```

So we eval the `predicate` each "step" and add the commands then.  

So given this, your mental model is in fact correct, its just delayed.

yifan-c (approved):

PR #193, opened 2025-05-15
Initialise CommandStores.nextId on restart

belliottsmith:


accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -827,14 +827,17 @@ public synchronized void initializeTopologyUnsafe(Journal.TopologyUpdate update)
...
         for (Map.Entry<Integer, RangesForEpoch> e : update.commandStores.entrySet())
         {
             Invariants.require(e.getValue() != null);
             EpochUpdateHolder holder = new EpochUpdateHolder();
             holder.add(1, e.getValue(), e.getValue().all());
             shards[i++] = new ShardHolder(supplier.create(e.getKey(), holder), e.getValue());
+            maxId = Math.max(maxId, e.getKey());
         }
 
+        nextId = maxId + 1;
```
dcapwell, line 840: so rather than starting at `0` and only bumping in `accord.local.CommandStores#updateTopology(accord.local.Node, accord.local.CommandStores.Snapshot, accord.topology.Topology, boolean)`, we figure out what the id is for each topology from the journal so the *next* `updateTopology` happens-after?

Makes sense to me, and locking also matches

dcapwell (approved):

PR #194, opened 2025-05-19
Fixes:

belliottsmith:
- cfk pruning+prebootstrap=invalid future dependency
 - exclude retired ranges when filtering RX stillTouches
 - propagate uses incorrect lowEpoch when fetch finds additional owned/touched ranges
 - node.withEpoch should callback with TopologyRetiredException, not throw
 - Recovery can race with durable-applied pruning; must not send durable unless latest ballot on apply
 - removeRedundantDependencies was not slicing pre-bootstrap range calculation to participating ranges
 - NPE in TopologyManager.atLeast caused by referencing an epoch that has been GC'd
 - use journal durableBeforePersister in burn test, not NOOP_PERSISTER
 - Fix failing invariant check(s), and burn test initialisation (TODO: confirm MaybeRecover check doesn't break forward progress)

### DIFF
--- a/accord-core/src/main/java/accord/local/StoreParticipants.java
+++ b/accord-core/src/main/java/accord/local/StoreParticipants.java
@@ -73,29 +73,40 @@ public class StoreParticipants
             }
         }
 
-        @Nullable public final Participants<?> executes()
+        @Override
+        public final @Nullable Participants<?> executes()
         {
             return executes;
         }
 
-        @Nullable public final Participants<?> waitsOn()
+        @Override
+        public final @Nullable Participants<?> waitsOn()
         {
             return waitsOn;
         }
 
+        @Override
         public final Participants<?> touches()
         {
             return touches;
         }
 
+        @Override
         public final Participants<?> hasTouched()
         {
             return hasTouched;
         }
 
+        @Override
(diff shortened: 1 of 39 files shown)


accord-core/src/main/java/accord/local/StoreParticipants.java
```diff
@@ -639,4 +653,42 @@ public static StoreParticipants all(Route<?> route)
...
+        Participants<?> unsynced = safeStore.node().topology().unsyncedOnly(participants, txnIdEpoch);
+        if (unsynced == null || unsynced.isEmpty())
+            return txnIdEpoch;
+
+        return computeCoveringEpoch(safeStore, txnIdEpoch, unsynced);
+    }
+
+    private static long computeCoveringEpoch(SafeCommandStore safeStore, long txnIdEpoch, Participants<?> participants)
+    {
+        long lowEpoch = safeStore.ranges().latestEarlierEpochThatFullyCovers(txnIdEpoch, participants);
```
ifesdjeen, line 691: `latestEarlier` feels a bit hard to parse, do we need "earlier" here?

ifesdjeen (approved):

PR #198, opened 2025-05-29
Fixes

belliottsmith:
- Don't assume stillExecutes applies to remote request - must mark unavailable any keys we don't have the txn definition for
 - Don't exit notifyManagedPreBootstrap notify loop early, as could have later transactions with lower txnId
 - CoordinateEphemeralRead must retry if insufficient responses from replicas that still own the range

### DIFF
--- a/accord-core/src/main/java/accord/coordinate/CoordinateEphemeralRead.java
+++ b/accord-core/src/main/java/accord/coordinate/CoordinateEphemeralRead.java
@@ -19,6 +19,7 @@
 package accord.coordinate;
 
 import java.util.Collection;
+import java.util.function.BiConsumer;
 
 import accord.api.Result;
 import accord.coordinate.tracking.QuorumTracker;
@@ -35,7 +36,7 @@ import accord.topology.Topologies;
 import accord.utils.Invariants;
 import accord.utils.SortedListMap;
 import accord.utils.async.AsyncResult;
-import accord.utils.async.AsyncResults;
+import accord.utils.async.AsyncResults.SettableByCallback;
 
 import static accord.api.ProtocolModifiers.QuorumEpochIntersections;
 import static accord.api.ProtocolModifiers.QuorumEpochIntersections.Include.Owned;
@@ -60,14 +61,23 @@ public class CoordinateEphemeralRead extends AbstractCoordinatePreAccept<Result,
 {
     public static AsyncResult<Result> coordinate(Node node, FullRoute<?> route, TxnId txnId, Txn txn)
     {
+        SettableByCallback<Result> result = new SettableByCallback<>();
+        coordinate(node, route, txnId, txn, result);
+        return result;
+    }
+
+    public static void coordinate(Node node, FullRoute<?> route, TxnId txnId, Txn txn, BiConsumer<Result, Throwable> callback)
+    {
         TopologyMismatch mismatch = TopologyMismatch.checkForMismatchOrPendingRemoval(node.topology().globalForEpoch(txnId.epoch()), txnId, route.homeKey(), route);
         if (mismatch != null)
-            return AsyncResults.failure(mismatch);
+        {
+            callback.accept(null, mismatch);
+            return;
+        }
 
         Topologies topologies = node.topology().withUnsyncedEpochs(route, txnId, txnId);
-        CoordinateEphemeralRead coordinate = new CoordinateEphemeralRead(node, topologies, route, txnId, txn);
+        CoordinateEphemeralRead coordinate = new CoordinateEphemeralRead(node, topologies, route, txnId, txn, callback);
         coordinate.start();
-        return coordinate;
     }
 
     private final Txn txn;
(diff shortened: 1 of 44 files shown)


accord-core/src/main/java/accord/coordinate/CoordinateEphemeralRead.java
```diff
@@ -65,20 +67,36 @@ public static AsyncResult<Result> coordinate(Node node, FullRoute<?> route, TxnI
...
 
         Topologies topologies = node.topology().withUnsyncedEpochs(route, txnId, txnId);
-        CoordinateEphemeralRead coordinate = new CoordinateEphemeralRead(node, topologies, route, txnId, txn);
+        SettableByCallback<Result> result = new SettableByCallback<>();
+        CoordinateEphemeralRead coordinate = new CoordinateEphemeralRead(node, topologies, route, txnId, txn, result);
+        coordinate.start();
+        return result;
+    }
+
+    public static void coordinate(Node node, FullRoute<?> route, TxnId txnId, Txn txn, BiConsumer<Result, Throwable> callback)
```
ifesdjeen, line 69: should we call this one coordinateAsync maybe?

accord-core/src/main/java/accord/coordinate/CoordinateEphemeralRead.java
```diff
@@ -65,20 +67,36 @@ public static AsyncResult<Result> coordinate(Node node, FullRoute<?> route, TxnI
             return AsyncResults.failure(mismatch);
 
         Topologies topologies = node.topology().withUnsyncedEpochs(route, txnId, txnId);
-        CoordinateEphemeralRead coordinate = new CoordinateEphemeralRead(node, topologies, route, txnId, txn);
+        SettableByCallback<Result> result = new SettableByCallback<>();
```
ifesdjeen, line 64: would something like this make sense? 

```
    public static AsyncResult<Result> coordinate(Node node, FullRoute<?> route, TxnId txnId, Txn txn)
    {
        SettableByCallback<Result> result = new SettableByCallback<>();
        coordinate(node, route, txnId, txn, result);
        return result;
    }
```

accord-core/src/test/java/accord/impl/basic/MonitoredPendingQueue.java
```diff
@@ -75,6 +75,30 @@ public void add(Pending item, long delay, TimeUnit units)
         wrapped.add(item, delay, units);
     }
 
+    @Override
+    public void preregister(Pending item)
```
ifesdjeen, line 79: Could you add a small comment on distinction between reregister and addPreregistered? 
looks like this was done for cancellation, but thought it might be useful for explicit. Maybe we could change name for addPreregistered to have larger distinction? Maybe `addPreregistered` could be something like schedule?

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -364,13 +365,25 @@ protected AsyncChain<Data> beginRead(SafeCommandStore safeStore, Timestamp execu
...
-        // TODO (required): for awaitsOnlyDeps commands, if we cannot infer an actual executeAtLeast we should confirm no situation where txnId is not an adequately conservative value for unavailable/unsafeToRead
-        Ranges ranges = safeStore.unsafeToReadAt(executeAt);
-        if (ranges.size() > command.route().size())
-            ranges = ranges.intersecting(command.route(), Minimal);
-        return ranges;
+        // note: syncpoints and ephemeral reads simply consume the latest information (whatever it is),
+        //  which is represented by the latest possible safeToRead entry (which is only updated on successful bootstrap)
+        //  We DO NOT use executesAtLeast here, this is used only to compute retryInFutureEpoch which reports ranges
+        //  we aren't able to serve information newer than executesAtLeast for.
+        //  executeAtLeast can yield false answers for safeToRead even for single keys by picking a time
```
ifesdjeen: I am afraid I can not parse (or am missing context to do so) this comment, particularly this last line.

accord-core/src/main/java/accord/messages/ReadData.java
```diff
@@ -364,13 +365,25 @@ protected AsyncChain<Data> beginRead(SafeCommandStore safeStore, Timestamp execu
...
+        {
+            Invariants.require(txnId.awaitsOnlyDeps());
+            executeAt = txnId;
+        }
+        Timestamp readAt = txnId.awaitsOnlyDeps() ? Timestamp.MAX : executeAt;
+        Ranges safeToRead = safeStore.safeToReadAt(readAt);
+        Ranges reads = safeStore.ranges().allAt(executeAt);
+        Ranges unsafeToRead = reads.without(safeToRead);
+        if (unsafeToRead.size() > command.route().size())
+                unsafeToRead = unsafeToRead.intersecting(command.route(), Minimal);
```
ifesdjeen: nit: indentation

ifesdjeen (approved):

PR #203, opened 2025-06-11
Fetch propagate

belliottsmith:


accord-core/src/main/java/accord/coordinate/CoordinatePreAccept.java
```diff
@@ -142,7 +142,7 @@ void onNewEpochTopologyMismatch(TopologyMismatch mismatch)
         proposeInvalidate(node, node.uniqueTimestamp(Ballot::fromValues), txnId, route.homeKey(), (outcome, failure) -> {
             if (failure != null)
                 mismatch.addSuppressed(failure);
-            setFailure(mismatch);
+            callback.accept(null, mismatch);
```
dcapwell, line 145: This patch doesn't remove `setFailure` and intellij is telling me that there are still callers to `setFailure`, why would you call `setFailure` in one code path and another directly notify the callback?  For example, this no longer updates the `eventListener` and doesn't set `isDone` (if we trigger the callback are we not actually done?)

belliottsmith: This call-site was mechanically updated without consideration previously, I assume (corroborated by the fact I don't remember it). It is a callback from another process that occurs _after_ isDone is already set.

belliottsmith: That is to say more specifically, this is invoked by `onPreAcceptedOrNewEpoch` and is the parallel work for rejecting the query, but the `CoordinatePreAccept` phase is logically done by this point and ownership is passed to the `notAccept` / `acceptInvalidate` phase.

dcapwell: > It is a callback from another process that occurs after isDone is already set.

that doesn't answer my question.  This change is taking 1 line from `setFailure` and ignoring the rest of the logic (such as monitoring).  If this change actually matters it implies that it was called *after* `isDone == true` *or* it isn't desirable to have `isDone = true` here (which feels weird as we are notifying the callback of a result).

Which is it?  If we feel its best to special case can we document the *why*?

belliottsmith: >  If this change actually matters it implies that it was called after isDone == true

I think I answered this already unless I'm missing something? Sorry if it was ambiguous, or if I have misunderstood something additional. To quote: "It is a callback ... that occurs _after isDone is already set_."

That is, it occurs after `isDone == true`

dcapwell: > but the CoordinatePreAccept phase is logically done by this point and ownership is passed to the notAccept / acceptInvalidate phase.

"logically" or is `isDone` already set to `true`?  Based off the current logic when `isDone` is set to `true` we also notify the callback.  So is this block of code trying to double notify the callback?

> "It is a callback ... that occurs after isDone is already set."

This implies to me we are trying to double notify the callback; is this correct?  Can we document this behavior?  Can we also keep the `eventListener` logic?

belliottsmith: > "logically" or is isDone already set to true? 

both....

> Based off the current logic when isDone is set to true we also notify the callback. 

No, that is not true. `isDone` does not mean the callback is invoked, it means we have completed the `PreAccept` phase and have passed the callback onto another phase, normally either Execute, Accept or AcceptInvalidate. It _could_ mean we invoke the callback if there were no other phase to invoke; indeed, invoking AcceptInvalidate is logically optional, so we could choose to invoke the callback immediately.

> Can we also keep the eventListener logic?

It's a no-op? We know what the type is, and it doesn't get invoked. We should revisit all of this in future, as maybe this isn't ideal behaviour, but I intend to look at observability holistically once stability is resolved, so better considered then.

dcapwell: thanks, i just found it.  `accord.coordinate.AbstractCoordinatePreAccept#onPreAcceptedOrNewEpoch` sets `isDone` but doesn't notify the callback.

accord-core/src/main/java/accord/coordinate/FetchSomeRoute.java
```diff
@@ -0,0 +1,153 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package accord.coordinate;
+
+import java.util.function.BiConsumer;
+
+import javax.annotation.Nullable;
```
dcapwell: unused import

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -715,49 +760,144 @@ protected <O> Cancellable mapReduceConsume(PreLoadContext context, Unseekables<?
...
+        {
+            return snapshot -> StoreFinder.find(snapshot, participants)
+                                          .filter(snapshot, participants, txnId.epoch(), (executeAt != null ? executeAt : txnId).epoch())
+                                          .iterator(snapshot);
+        }
+
+        @Override
+        public Iterator<CommandStore> select(Snapshot snapshot)
+        {
+            throw new UnsupportedOperationException();
```
dcapwell: been tracing the code and not found 1 place that calls `none` then `select` without `refine`.  Its nested so might have missed something

dcapwell (commented):
```
> Task :accord-core:checkstyleMain
[ant:checkstyle] [ERROR] /Users/dcapwell/src/apple/aci/cassandra/cie-5.1/modules/accord/accord-core/src/main/java/accord/impl/progresslog/WaitingState.java:30:8: Unused import - accord.local.CommandStores. [UnusedImports]
[ant:checkstyle] [ERROR] /Users/dcapwell/src/apple/aci/cassandra/cie-5.1/modules/accord/accord-core/src/main/java/accord/impl/progresslog/WaitingState.java:33:8: Unused import - accord.local.CommandStores.StoreSelector. [UnusedImports]
[ant:checkstyle] [ERROR] /Users/dcapwell/src/apple/aci/cassandra/cie-5.1/modules/accord/accord-core/src/main/java/accord/coordinate/Invalidate.java:52:15: Unused import - accord.utils.Invariants.nonNull. [UnusedImports]
[ant:checkstyle] [ERROR] /Users/dcapwell/src/apple/aci/cassandra/cie-5.1/modules/accord/accord-core/src/main/java/accord/coordinate/MaybeRecover.java:23:8: Unused import - accord.local.CommandStores. [UnusedImports]
[ant:checkstyle] [ERROR] /Users/dcapwell/src/apple/aci/cassandra/cie-5.1/modules/accord/accord-core/src/main/java/accord/coordinate/FetchSomeRoute.java:23:8: Unused import - javax.annotation.Nullable. [UnusedImports]
```

dcapwell (approved):
please fix the checkstyle issues before merge

accord-core/src/main/java/accord/local/CommandStores.java
belliottsmith: I have made a last minute change to split out these into two interfaces, so we make the transition explicit. We can bike shed naming in post.

PR #204, opened 2025-06-13
Bugs 250612

belliottsmith:


accord-core/src/main/java/accord/impl/progresslog/WaitingState.java
```diff
@@ -712,15 +712,15 @@ static void fetchRoute(DefaultProgressLog owner, BlockedUntil blockedUntil, TxnI
         // TODO (desired): fetch only the route
         // we MUSt allocate before calling withEpoch to register cancellation, as async
         BiConsumer<Route<?>, Throwable> invoker = invokeWaitingCallback(owner, txnId, blockedUntil, WaitingState::fetchRouteCallback);
-        FetchSomeRoute.fetchSomeRoute(owner.node(), txnId, contactable, new IncludingSpecificStoreSelector(owner.commandStore.id()), invoker);
+        accord.coordinate.FetchRoute.fetchRoute(owner.node(), txnId, contactable, new IncludingSpecificStoreSelector(owner.commandStore.id()), invoker);
```
ifesdjeen: Nit: maybe it would be better to use callback kind's FetchRoute scoped with `CallbackKind`, like `CallbackKind#FetchRoute` so that we do not have to use fully qualified name here?

accord-core/src/main/java/accord/impl/progresslog/WaitingState.java
```diff
@@ -23,9 +23,9 @@
 import javax.annotation.Nullable;
 
 import accord.api.ProgressLog.BlockedUntil;
+import accord.api.ProtocolModifiers;
```
ifesdjeen: nit: seemingly unused

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -1191,7 +1191,8 @@ private static long validateMax(long maxEpoch, Epochs snapshot)
             return snapshot.currentEpoch;
 
         Invariants.require(snapshot.currentEpoch >= maxEpoch, "current epoch %d < provided max %d", snapshot.currentEpoch, maxEpoch);
-        Invariants.require(snapshot.minEpoch() <= maxEpoch, "minimum known epoch %d > provided max %d", snapshot.minEpoch(), maxEpoch);
+        if (snapshot.minEpoch() > maxEpoch)
```
ifesdjeen: I think maxEpoch can still be equal to minEpoch. 

I think we should use the ```if (maxEpoch < snapshot.minEpoch())` condition instead.

ifesdjeen (approved):

accord-core/src/main/java/accord/topology/TopologyManager.java
ifesdjeen: (my mistake, flipped arguments)

PR #205, opened 2025-06-16
Fast coordination optimisations

belliottsmith:


accord-core/src/main/java/accord/coordinate/CoordinateTransaction.java
```diff
@@ -111,11 +113,18 @@ void onPreAccepted(Topologies topologies, Timestamp executeAt, SortedListMap<Nod
             Deps deps = mergeFastOrMediumDeps(oks);
             if (deps != null)
             {
-                ExecuteFlags executeFlags = Functions.foldl(oks.valuesAsNullableList(), (ok, v) -> ok == null ? v : v.and(ok.flags), ExecuteFlags.all());
+                CoordinationFlags executeFlags = oks.foldlNonNull((d, k, v, out) -> {
```
ifesdjeen: Took me a bit to understand this; maybe this warrants a comment. Also, should we deduplicate this with `CoordinateEphemeralRead`, in case this logic gets changed in future?

accord-core/src/main/java/accord/coordinate/ExecuteFlag.java
```diff
@@ -38,10 +43,107 @@ public static final class ExecuteFlags extends TinyEnumSet<ExecuteFlag>
...
+        @Override
+        public boolean hasUniqueHlc()
+        {
+            return hasUniqueHlc;
+        }
+
+        @Override
+        public void add(Node.Id node, ExecuteFlags flags)
+        {
+            if (hasUniqueHlc && !flags.contains(HAS_UNIQUE_HLC))
```
ifesdjeen: nit: 

```
            hasUniqueHlc &= flags.contains(HAS_UNIQUE_HLC);
```

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
```diff
@@ -262,28 +247,21 @@ public String toString()
...
-        onSlowResponse(node.id());
-        localTimeout = null;
-    }
-
-    @Override
-    public int stripe()
+    boolean mayFastExecute(ExecuteFlags flags)
     {
-        return txnId.hashCode();
+        return flags.contains(READY_TO_EXECUTE) && (!txnId.hasPrivilegedCoordinator() || path != FAST) && fastReadsMayBypassSafeStore(txnId);
```
ifesdjeen, line 263: I am very likely misunderstanding something but `(!txnId.hasPrivilegedCoordinator() || path != FAST`: if has no privileged coordinator, or path isn't fast, allow fast execute?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -519,19 +527,55 @@ else if (command.hasBeen(PreCommitted) && !executeAt.equals(command.executeAt())
...
+                logger.trace("{}: applying", executed.txnId());
+                applyChain(safeStore, executed).begin(safeStore.agent());
+                break;
+            }
+            case Applied:
+            {
+                Command.Executed executed = safeCommand.applied(safeStore, participants, executeAt, partialTxn, partialDeps, waitingOn, writes, result);
+                safeStore.agent().eventListener().onPreApplied(executed);
+                safeStore.agent().eventListener().onApplied(executed, -1);
+                safeStore.notifyListeners(safeCommand, command);
```
ifesdjeen, line 572: nit: I would still add `break` here just for clarity / in case we move things around

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -963,6 +1017,12 @@ private static Command purge(Command command, @Nonnull StoreParticipants newPart
         return result;
     }
 
+    private static boolean validateSafeToFastApply(SafeCommandStore safeStore, SafeCommand safeCommand)
+    {
+        // TODO (now): implement validation!!
+        return true;
```
ifesdjeen: Just to make sure: did now mean in this patch or we're ok to leave it out for now?

ifesdjeen (approved):

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: I will aim to get this done before merge, thanks for pointing it out (though, it should already be covered by existing validation, this would provide tighter/faster coverage)

PR #207, opened 2025-06-16
Changed staleIds signature to SortedArrayList so its easier to use in the serializer

dcapwell:
C*: https://github.com/apache/cassandra/pull/4202

### DIFF
--- a/accord-core/src/main/java/accord/topology/Topology.java
+++ b/accord-core/src/main/java/accord/topology/Topology.java
@@ -26,7 +26,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
@@ -659,7 +658,7 @@ public class Topology
         return subsetOfRanges;
     }
 
-    public Set<Id> staleIds()
+    public SortedArrayList<Id> staleIds()
     {
         return staleNodes;
     }
--- a/accord-core/src/main/java/accord/topology/Topology.java
+++ b/accord-core/src/main/java/accord/topology/Topology.java
@@ -182,6 +182,12 @@ public class Topology
         this.supersetIndexes = supersetIndexes;
     }
 
+    @VisibleForTesting
+    public Topology withEpoch(long epoch)
+    {
+        return new Topology(global, epoch, shards, ranges, staleNodes, nodeIds, nodeLookup, subsetOfRanges, supersetIndexes);
+    }
+
     public Topology global()
     {
         return global == null ? this : global;
--- a/accord-core/src/main/java/accord/utils/SortedArrays.java
+++ b/accord-core/src/main/java/accord/utils/SortedArrays.java
@@ -21,6 +21,7 @@ package accord.utils;
 import java.lang.reflect.Array;
 import java.util.AbstractList;
 import java.util.Arrays;
+import java.util.BitSet;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.Iterator;
@@ -1792,5 +1793,31 @@ public class SortedArrays
         int t = values[i];
         values[i] = values[j];
         values[j] = t;
+    }
+
+    public static <T extends Comparable<? super T>> BitSet toBitSet(SortedArrays.SortedArrayList<T> src,
+                                                                    SortedArrays.SortedArrayList<T> subset)
+    {
+        //TODO (now): create unit test
+        BitSet bitSet = new BitSet(src.size());
+        for (int i = 0; i < src.size(); i++)
+        {
+            if (subset.contains(src.get(i)))
+                bitSet.set(i);
+        }
+        return bitSet;
+    }
+
+    public static <T extends Comparable<? super T>> SortedArrays.SortedArrayList<T> fromBitSet(SortedArrays.SortedArrayList<T> src,
+                                                                                               BitSet bitSet,
+                                                                                               IntFunction<T[]> alloc)
+    {
+        SortedArrays.SortedArrayList.Builder<T> builder = new SortedArrays.SortedArrayList.Builder<>(alloc.apply(bitSet.cardinality()));
+        for (int i = 0; i < src.size(); i++)
+        {
+            if (bitSet.get(i))
+                builder.add(src.get(i));
+        }
+        return builder.build();
     }
 }
--- a/accord-core/src/main/java/accord/utils/SortedArrays.java
+++ b/accord-core/src/main/java/accord/utils/SortedArrays.java
@@ -21,7 +21,6 @@ package accord.utils;
 import java.lang.reflect.Array;
 import java.util.AbstractList;
 import java.util.Arrays;
-import java.util.BitSet;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.Iterator;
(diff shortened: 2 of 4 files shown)


accord-core/src/main/java/accord/topology/Topology.java
```diff
@@ -638,7 +643,7 @@ public Ranges ranges()
         return subsetOfRanges;
     }
 
-    public Set<Id> staleIds()
+    public SortedArrayList<Id> staleIds()
```
dcapwell, line 667: `accord.topology.Topology#Topology(long, accord.utils.SortedArrays.SortedArrayList<accord.local.Node.Id>, accord.topology.Shard...)`

we take a `SortedArrayList` as part of our public api, so this just makes it easier for C* to work with it and avoid having to cast

accord-core/src/main/java/accord/utils/SortedArrays.java
```diff
@@ -1793,4 +1794,29 @@ private static void swap(int[] values, int i, int j)
...
+        for (int i = 0; i < src.size(); i++)
+        {
+            if (subset.contains(src.get(i)))
+                bitSet.set(i);
+        }
+        return bitSet;
+    }
+
+    public static <T extends Comparable<? super T>> SortedArrays.SortedArrayList<T> fromBitSet(SortedArrays.SortedArrayList<T> src,
+                                                                                               BitSet bitSet,
```
belliottsmith: More consistent to use `SimpleBitSet` here as well, since this is what this codebase uses primarily

accord-core/src/main/java/accord/utils/SortedArrays.java
```diff
@@ -1793,4 +1794,29 @@ private static void swap(int[] values, int i, int j)
         values[i] = values[j];
         values[j] = t;
     }
+
+    public static <T extends Comparable<? super T>> BitSet toBitSet(SortedArrays.SortedArrayList<T> src,
```
belliottsmith: nit: superset vs src (to match subset)

accord-core/src/main/java/accord/utils/SortedArrays.java
```diff
@@ -1793,4 +1794,29 @@ private static void swap(int[] values, int i, int j)
...
+                                                                    SortedArrays.SortedArrayList<T> subset)
+    {
+        BitSet bitSet = new BitSet(src.size());
+        for (int i = 0; i < src.size(); i++)
+        {
+            if (subset.contains(src.get(i)))
```
belliottsmith: we should use `foldlIntersection` for O(n) versus O(n.lg(n)) complexity

e.g.

`return foldlIntersection(superset.array, 0, superset.array.length, subset.array, 0, subset.array.length, (i, bs, li, ri) -> { bs.set(li); return bs; }, new SimpleBitSet(superset.size()));`

This does require copying foldlIntersection to support operating over objects rather than long:

```
    @Inline
    public static <T extends Comparable<? super T>, V> V foldlIntersection(T[] as, int ai, int alim, T[] bs, int bi, int blim, IndexedFoldIntersect<? super T, V> fold, V accumulate)
    {
        return foldlIntersection(1, Comparable::compareTo, as, ai, alim, bs, bi, blim, fold, accumulate);
    }

    @Inline
    public static <T, V> V foldlIntersection(int aiMatchIncrement, AsymmetricComparator<? super T, ? super T> comparator, T[] as, int ai, int alim, T[] bs, int bi, int blim, IndexedFoldIntersect<? super T, V> fold, V accumulate)
    {
        while (true)
        {
            long abi = findNextIntersection(as, ai, alim, bs, bi, blim, comparator);
            if (abi < 0)
                break;

            ai = (int)(abi);
            bi = (int)(abi >>> 32);

            accumulate = fold.apply(as[ai], accumulate, ai, bi);
            ai += aiMatchIncrement;
            ++bi;
        }

        return accumulate;
    }
```

and copying IndexedFoldIntersectToLong to IndexedFoldIntersect:

```
package accord.utils;

public interface IndexedFoldIntersect<P1, Accumulate>
{
    /**
     * Apply some merge function accepting a constant object parameter p1, and the prior output of this function
     * or the initial value, to some element of a collection, with the index of the element provided.
     *
     * This function is used for efficiently folding over some subset of a collection.
     */
    Accumulate apply(P1 p1, Accumulate p2, int leftIndex, int rightIndex);
}
```

accord-core/src/main/java/accord/utils/SortedArrays.java
dcapwell: done

accord-core/src/main/java/accord/utils/SortedArrays.java
dcapwell: > we should use foldlIntersection for O(n) versus O(n.lg(n)) complexity

Can you explain where this `lg(n)` is coming from?

```java
for (int i = 0; i < superset.size(); i++)
```

We are looping, so `O(N)` on that

We then call `accord.utils.SimpleBitSet#get`

```java
int index = indexOf(i);
long bit = bit(i);
return 0 != (bits[index] & bit);
```

which should be `O(1)`

We then call

```java
builder.add(superset.get(i));
```

`add` is `O(1)` and `get` is also `O(1)`

belliottsmith: `subset.contains` is `lg(n)` (as would be `superset.indexOf`, though this way round (looping over subset) would be more efficient)

dcapwell: ah thanks!

dcapwell: so i didn't go the `foldlIntersection` route and just did `findNextIntersection` directly; which is what `fold` is doing anyways

accord-core/src/main/java/accord/utils/SortedArrays.java
```diff
@@ -1793,4 +1793,36 @@ private static void swap(int[] values, int i, int j)
...
+        {
+            long ri = SortedArrays.findNextIntersection(superset.array, i, subset.array, subsetIndex);
+            if (ri < 0)
+                break;
+            i = (int) (ri);
+            subsetIndex = (int) (ri >>> 32);
+
+            bitSet.set(i);
+        }
+        Invariants.require(bitSet.getSetBitCount() <= subset.size(), "Generated bit set is larger than the subset!");
```
belliottsmith, line 1812: Maybe check lastSetBit is strictly less instead?

accord-core/src/main/java/accord/utils/SortedArrays.java
```diff
@@ -1793,4 +1793,36 @@ private static void swap(int[] values, int i, int j)
...
+        return bitSet;
+    }
+
+    public static <T extends Comparable<? super T>> SortedArrays.SortedArrayList<T> fromSimpleBitSet(SortedArrays.SortedArrayList<T> superset,
+                                                                                                     SimpleBitSet bitSet,
+                                                                                                     IntFunction<T[]> alloc)
+    {
+        SortedArrays.SortedArrayList.Builder<T> builder = new SortedArrays.SortedArrayList.Builder<>(alloc.apply(bitSet.getSetBitCount()));
+        for (int i = 0; i < superset.size(); i++)
```
belliottsmith: we can/should gallop here as well, using i = bitSet.firstSetBit(); i >= 0 ; i = bitSet.nextSetBit(i + 1)

accord-core/src/main/java/accord/utils/SortedArrays.java
dcapwell: in the test i do the following

```
testSerde(expected, expected);
```

which fails if you make that change

```
Caused by: java.lang.IllegalStateException: Generated bit set is larger than the subset!
	at accord.utils.Invariants.createIllegalState(Invariants.java:77)
	at accord.utils.Invariants.illegalState(Invariants.java:82)
	at accord.utils.Invariants.require(Invariants.java:236)
	at accord.utils.SortedArrays.toSimpleBitSet(SortedArrays.java:1812)
```

dcapwell: in the topology patch this shouldn't happen, but can't say that is true if this code gets used in any other context, so `<=` is safest

accord-core/src/main/java/accord/utils/SortedArrays.java
dcapwell: pushed and tests are passing

belliottsmith (approved):

dcapwell:
merged in 3c9b3077df1c74a14d8fe8dc58fb3f61b8257236

PR #211, opened 2025-06-26
Fixes 250622

ifesdjeen:


accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -309,15 +313,33 @@ public Cleanup shouldCleanup(Input input, RedundantBefore redundantBefore, Durab
...
                 return NO;
 
+            if (cleanup != null)
+            {
+                switch (cleanup)
+                {
+                    case EXPUNGE:
+                        return EXPUNGE;
+                    case ERASE:
+                        if (EXPUNGE == Cleanup.shouldCleanup(input, txnId, null, SaveStatus.Erased, NotDurable, null, redundantBefore, durableBefore))
```
ifesdjeen, line 323: Would be great to add a comment since it's nested under ERASE

accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -309,15 +313,33 @@ public Cleanup shouldCleanup(Input input, RedundantBefore redundantBefore, Durab
...
             StoreParticipants participants = this.participants;
             // TODO (expected): we need to filter participants to correctly compute doesStillExecute in Cleanup.shouldCleanup;
             //  would be better to break this dependency, or otherwise encode it better.
             //  In particular it would be nice to avoid doing this twice for each command on load, as we also do this in SafeCommandStore.
             //  Perhaps we can special-case loading, and simply update the participants here so we can avoid doing it again on access
-            if (input == Input.FULL && participants != null)
-                participants = participants.filter(LOAD, redundantBefore, txnId, saveStatus != null && saveStatus.known.isExecuteAtKnown() ? executeAt : null);
+            if (input == Input.FULL)
+            {
+                if (saveStatus == null)
```
ifesdjeen, line 338: Can this happen as a result of a partial compaction?

accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -461,6 +483,7 @@ public boolean addCleanup(boolean clearFields, Cleanup addCleanup)
 
             hasUpdate = true;
             cleanup = addCleanup;
+            flags |= setChanged(CLEANUP);
```
ifesdjeen, line 486: Don't we need to set the same flag in cleanup below too?

ifesdjeen (commented):
+1

PR #212, opened 2025-06-28
Fix:

belliottsmith:
-EphemeralReads should retry in a later epoch if replication factor changes
 - Fix no local epoch for NotAccept
 - Invalidate a command with no route but for which we know we own some key that has applied locally
 - Don't pre-merge existing DurableBefore, to reduce duplicate/redundant persistence
 - Don't reconstruct commands with null participants; must have been EXPUNGED

### DIFF
--- a/accord-core/src/main/java/accord/impl/CommandChange.java
+++ b/accord-core/src/main/java/accord/impl/CommandChange.java
@@ -336,6 +336,7 @@ public class CommandChange
             {
                 if (saveStatus == null)
                     return EXPUNGE;
+                
                 if (participants != null)
                     participants = participants.filter(LOAD, redundantBefore, txnId, saveStatus.known.isExecuteAtKnown() ? executeAt : null);
             }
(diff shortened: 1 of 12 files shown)


accord-core/src/main/java/accord/topology/Topology.java
```diff
@@ -46,6 +46,7 @@
 import accord.utils.ArrayBuffers.IntBuffers;
 import accord.utils.IndexedBiFunction;
 import accord.utils.IndexedConsumer;
+import accord.utils.IndexedFoldToLong;
```
ifesdjeen, line 49: Nit: unused

accord-core/src/main/java/accord/messages/GetEphemeralReadDeps.java
```diff
@@ -72,10 +73,10 @@ public GetEphemeralReadDepsOk apply(SafeCommandStore safeStore)
         long latestEpoch = Math.max(safeStore.node().epoch(), node.epoch());
 
         // TODO (desired): only return failure if we've participated in a sync point that could migrate coordination to the newer epoch
-        if (latestEpoch > executionEpoch && safeStore.ranges().removed(executionEpoch, latestEpoch).intersects(scope))
+        StoreParticipants participants = StoreParticipants.read(safeStore, scope, txnId, minEpoch, latestEpoch);
+        if (latestEpoch > executionEpoch && (safeStore.ranges().removed(executionEpoch, latestEpoch).intersects(participants.owns()) || node.topology().hasReplicationMaybeChanged(participants.owns(), executionEpoch)))
```
ifesdjeen, line 77: Are brackets intended around `(safeStore.ranges().removed(executionEpoch, latestEpoch).intersects(participants.owns())` ? Just making sure they're not misplaced.

accord-core/src/main/java/accord/impl/CommandChange.java
```diff
@@ -334,10 +334,9 @@ public Cleanup shouldCleanup(Input input, RedundantBefore redundantBefore, Durab
             //  Perhaps we can special-case loading, and simply update the participants here so we can avoid doing it again on access
             if (input == Input.FULL)
             {
-                if (saveStatus == null)
+                if (saveStatus == null || participants == null)
```
ifesdjeen: I'm a bit concerned here, but maybe incorrectly: if we mistakenly lose participants, we will end up purging entire command change (writing this actually only losing participants seems equally bad).

ifesdjeen (approved):

PR #213, opened 2025-06-30
CASSANDRA-20741: Add rebootstrap

ifesdjeen:


accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -699,6 +699,15 @@ protected boolean shouldBootstrap(Node node, Topology local, Topology newLocalTo
         return newLocalTopology.epoch() != 1;
     }
 
+    public AsyncChain<?> rebootstrap(Node node)
+    {
+        List<AsyncChain<?>> results = new ArrayList<>();
+        Snapshot snapshot = current;
+        for (ShardHolder shard : snapshot.shards)
+            results.add(shard.store.rebootstrap(node, shard.ranges.currentRanges(), snapshot.global.epoch()));
```
belliottsmith: do we always want to rebootstrap everything, or might we want to re bootstrap only a given range sometimes?

accord-core/src/main/java/accord/local/Bootstrap.java
```diff
@@ -146,7 +146,13 @@ void start(SafeCommandStore safeStore)
                      .flatMap(syncPoint -> node.withEpochAtLeast(epoch, null, () -> store.build(empty(), safeStore1 -> {
                          if (valid.isEmpty()) // we've lost ownership of the range
                              return AsyncResults.success(Ranges.EMPTY);
-                         return fetch = safeStore1.dataStore().fetch(node, safeStore1, valid, syncPoint, this);
+                         if (kind == Kind.Sync)
+                             return fetch = safeStore1.dataStore().sync(node, safeStore1, valid, syncPoint, this);
```
belliottsmith: maybe pass this parameter to `fetch` rather than having two methods?

accord-core/src/main/java/accord/local/CommandStores.java
ifesdjeen: My understanding is that we always want to rebootstrap everything, especially if we've been down for a while. 

I can add an option to rebootstrap only a specific range, for example, if we know that there was a single problem during replay, in which case I can grab command store's ranges for the command that failed.

WDYT?

belliottsmith: We may rebootstrap because we're _behind_ in some cases (not because replay has failed), and in this case we will know which ranges we might be behind on, and so can target the rebootstrap. So, I think it would be helpful to be able to target just a given sub-range even with a command store?

belliottsmith: FWIW, we also need a parameter (*somewhere*) that tells the CommandStore we're rebootstrapping _due to corruption_ (or loss of data). This should _reset_ RedundantBefore, because we don't know what parts of the log we know our prior responses/decisions for locally.

belliottsmith:
It would help to integrate this into the burn test so we can exercise its correctness in isolation (and would help me map out in my head all of the interactions, so I can be sure it's doing everything it needs to). Looks good overall though!

accord-core/src/main/java/accord/coordinate/ExecuteEphemeralRead.java
```diff
@@ -233,6 +233,23 @@ protected boolean cancel()
             return true;
         }
 
+        @Override
+        protected void onFailure(CommitOrReadNack failReply, Throwable throwable)
```
belliottsmith: why duplicate this? Is this to address the double invocation of callback, so we can call super.cancel? If so, I think it would be better to fix this by changing the cancel override to instead override timeout.

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
```diff
@@ -329,6 +329,24 @@ protected boolean cancel()
             return true;
         }
 
+        @Override
+        protected void onFailure(CommitOrReadNack failReply, Throwable throwable)
```
belliottsmith: same question here regarding overriding onFailure

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -832,7 +831,7 @@ public void updateExclusiveSyncPoint(Command prev, Command updated, boolean forc
                     }
                     if (isShadowedByPreBootstrap) continue;
                 }
-                illegalState();
+                illegalState(String.format("Prev: %s, updated: %s, command: %s", prev, updated, command));
```
belliottsmith: there's an override of illegalState that accepts format params

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -895,4 +938,11 @@ public NodeCommandStoreService node()
     {
         return node;
     }
+
+    /**
+     * Thrown when command store is not ready to serve the request for a given range.
+     */
+    public static class NotReadyException extends RuntimeException
```
belliottsmith, line 984: I think NotReady is perhaps the wrong name for this, since we will never be ready, unless I misunderstand the context. The point is we don't have the relevant log history anymore. The best we can ever hope to respond with is Vestigial. The only reason we might not convert immediately to Vestigial is that a transaction may only partially overlap the rebootstrap, so we need to support maintaining some local state for the transaction in this case.

accord-core/src/main/java/accord/local/RedundantBefore.java
```diff
@@ -924,6 +936,22 @@ public boolean isShardOnlyApplied(TxnId txnId, Unseekables<?> participants)
         return foldl(participants, Bounds::isShardOnlyApplied, false, txnId);
     }
 
+    public boolean isUnsafeBefore(TxnId txnId, Unseekables<?> participants)
+    {
+        // Definitely safe
+        if (maxUnsafeBefore == TxnId.NONE)
+            return false;
+
+        // TODO: Problem is that, logically, we do know that transactions after maxUnsafeBefore (txnId.compareTo(maxUnsafeBefore) > 0)
```
belliottsmith: Is this still true? Because if so we need to understand it, and using a weaker comparison here shouldn't have any bearing on that fact, as it should only decide whether we invoke the below foldl which provides the more precise answer?

accord-core/src/main/java/accord/local/RedundantStatus.java
```diff
@@ -159,6 +166,7 @@ public static class SomeStatus
         public static final SomeStatus MAJORITY_APPLIED_ONLY = oneSlow(MAJORITY_APPLIED);
         public static final SomeStatus SHARD_APPLIED_ONLY = oneSlow(SHARD_APPLIED);
         public static final SomeStatus GC_BEFORE_AND_LOCALLY_APPLIED = multi(GC_BEFORE, LOCALLY_APPLIED);
+        public static final SomeStatus SHARD_UNSAFE_BEFORE = oneSlow(UNSAFE_BEFORE);
```
belliottsmith: definitely not SHARD unsafe, in this class SHARD refers to a whole shard property. We mean LOCALLY_UNSAFE or perhaps better would be LOCALLY_CORRUPT or LOCALLY_LOST or LOCALLY_INCOMPLETE. This should be consistent with the Property name, and be suffixed with _ONLY here.

accord-core/src/main/java/accord/local/RedundantStatus.java
```diff
@@ -112,6 +114,11 @@ public enum Property
         // not persisted
         WAS_OWNED                          (false,  false, LT, LOCALLY_DEFUNCT),
         NOT_OWNED                          (false, false, LT),
+
+        /**
+         * TODO
+         */
+        UNSAFE_BEFORE                      (false, false, LT),
```
belliottsmith: LOCALLY_CORRUPT or LOCALLY_LOST or ...? Should imply PRE_BOOTSTRAP_OR_STALE, since we must be stale (we cannot execute anything for this range until we bootstrap)

accord-core/src/test/java/accord/impl/list/ListAgent.java
```diff
@@ -124,7 +125,7 @@ public Tracing trace(TxnId txnId, TraceEventType eventType)
     @Override
     public void onInconsistentTimestamp(Command command, Timestamp prev, Timestamp next)
     {
-        throw new AssertionError("Inconsistent execution timestamp detected for command " + command + ": " + prev + " != " + next);
+        throw new AssertionError(String.format("Inconsistent execution timestamp detected for command %s: %s != %s", command, prev, next));
```
belliottsmith: perhaps just invoke Invariants.illegalState if you want to reformat this?

accord-core/src/main/java/accord/local/SafeCommandStore.java
```diff
@@ -170,6 +170,8 @@ protected SafeCommand maybeCleanup(SafeCommand safeCommand, @Nonnull StorePartic
     {
         Command command = safeCommand.current();
         StoreParticipants participants = command.participants().supplementOrMerge(command.saveStatus(), supplemental);
+        if (!commandStore().safeToRespond(command.txnId(), command.route()))
```
belliottsmith: We should be able to roll this into the internal maybeCleanup call, by just checking status.any(UNSAFE/LOCALLY_CORRUPT/whatever)

We might want to somehow capture whether this load is to reply to an external request or process the transaction locally, else we may struggle to handle partial rebootstraps where a transaction is considered known for one range but not for another (i.e., the transaction is no doubt known, but the log is otherwise not safe for one part).

Either that, or we may want to define this as only affecting transactions that are fully covered by UNSAFE/LOCALLY_CORRUPT, in which case we would only throw the exception if status.all(UNSAFE/LOCALLY_CORRUPT). In this case, it might be the case that we can consider merging this behaviour with Vestigial handling. I don't have a well formed opinion about how best we should proceed here, but can experiment with the patch if you like, or we can discuss and you can explore the possibility space.

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -1432,12 +1434,20 @@ enum Validated { INSUFFICIENT, UPDATE_TXN_IGNORE_DEPS, UPDATE_TXN_KEEP_DEPS, UPD
...
     }
 
     private static Validated validate(@Nullable Ballot ballot, SaveStatus newStatus, Command cur, StoreParticipants participants,
                                       Route<?> addRoute, @Nullable Txn addPartialTxn, @Nullable Deps partialDeps,
                                       @Nullable Commit.Kind commitKind, @Nullable Timestamp executeAt)
+    {
+        return validate(ballot, newStatus, cur, participants, addRoute, addPartialTxn, partialDeps, commitKind, executeAt, null);
+    }
+
+    private static Validated validate(@Nullable Ballot ballot, SaveStatus newStatus, Command cur, StoreParticipants participants,
```
belliottsmith: this isn't wired up, so not sure how I feel about the tracing at present, need to understand how it can/will be used...

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -699,6 +700,18 @@ protected boolean shouldBootstrap(Node node, Topology local, Topology newLocalTo
         return newLocalTopology.epoch() != 1;
     }
 
+    public AsyncChain<Void> rebootstrap(Node node)
```
belliottsmith, line 703: I think we need to test partial rebootstrap to have confidence in the logic. I think it would help to also design this as something that can be invoked on a node that is already up, rather than only as an operation that may be invoked on restart operation

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -548,6 +563,34 @@ final Supplier<EpochReady> bootstrapper(Node node, Ranges newRanges, long epoch)
...
+        return !unsafeGetRedundantBefore().isUnsafeBefore(txnId, participants);
+    }
+
+    protected EpochReady rebootstrap(Node node, Ranges ranges, long epoch)
+    {
+        AsyncResult<EpochReady> metadata = submit(empty(), safeStore -> {
+            Bootstrap bootstrap = new Bootstrap(node, this, epoch, ranges, DataStore.RequestKind.Sync);
+            bootstraps.add(bootstrap);
+            // If rebootstrap can grab a later timestamp for subsequent attempts, but this timestamp is enough for us
+            // to establish what's safe to read
```
belliottsmith: safeToRead is a specific concept (related to serving a data read), let's rephrase the comment to avoid confusion?

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -548,6 +563,34 @@ final Supplier<EpochReady> bootstrapper(Node node, Ranges newRanges, long epoch)
...
+            TxnId unsafeBefore = bootstrap.start(safeStore);
+            logger.debug("Rebootstrap timestamp on {}@{}: {}", id, node.id(), unsafeBefore);
+            unsafeUpsertRedundantBefore(RedundantBefore.create(ranges, unsafeBefore, SHARD_UNSAFE_BEFORE));
+            return new EpochReady(epoch, null, null,
+                                  bootstrap.data,
+                                  bootstrap.reads);
+        });
+
+        AsyncResult<Void> readyToCoordinate = readyToCoordinate(ranges, epoch);
```
belliottsmith, line 607: I don't think this fully achieves its goal. We may have already notified other replicas that we are ready to coordinate, but we may no longer be. So while we need to recreate this collection and go through the motions again (to handle the case that we *haven't* notified other replicas of this), we need to refuse to answer *any* dependency requests for *any* intersecting transaction (regardless of its `TxnId`) until this completes.

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -548,6 +563,34 @@ final Supplier<EpochReady> bootstrapper(Node node, Ranges newRanges, long epoch)
...
+    protected EpochReady rebootstrap(Node node, Ranges ranges, long epoch)
+    {
+        AsyncResult<EpochReady> metadata = submit(empty(), safeStore -> {
+            Bootstrap bootstrap = new Bootstrap(node, this, epoch, ranges, DataStore.RequestKind.Sync);
+            bootstraps.add(bootstrap);
+            // If rebootstrap can grab a later timestamp for subsequent attempts, but this timestamp is enough for us
+            // to establish what's safe to read
+            TxnId unsafeBefore = bootstrap.start(safeStore);
+            logger.debug("Rebootstrap timestamp on {}@{}: {}", id, node.id(), unsafeBefore);
+            unsafeUpsertRedundantBefore(RedundantBefore.create(ranges, unsafeBefore, SHARD_UNSAFE_BEFORE));
```
belliottsmith: I think we need a safeStore here don't we? Else this won't be persisted, at least in the integration.

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -548,6 +563,34 @@ final Supplier<EpochReady> bootstrapper(Node node, Ranges newRanges, long epoch)
...
     }
 
+    public boolean safeToRespond(TxnId txnId, Unseekables<?> participants)
+    {
+        return !unsafeGetRedundantBefore().isUnsafeBefore(txnId, participants);
+    }
+
+    protected EpochReady rebootstrap(Node node, Ranges ranges, long epoch)
+    {
+        AsyncResult<EpochReady> metadata = submit(empty(), safeStore -> {
```
belliottsmith, line 590: We should probably markUnsafeToRead before we do anything else, since otherwise it happens asynchronously and may still be initialised from before

belliottsmith:
I think the patch is also ready to evolve to handling both variations of rebootstrap, i.e. we may rebootstrap because we're "stale" (but have a complete copy of our own transaction log history), and in this case we don't need to impose all of the new restrictions. I think it would be good to add the switch for this as part of this round of changes.

accord-core/src/main/java/accord/local/Commands.java
ifesdjeen: I can revert on commit. Was very useful when I was debugging, but I may as well just add printlns next time.

accord-core/src/main/java/accord/local/CommandStore.java
ifesdjeen: Renamed to TransactionLostException for now.

accord-core/src/main/java/accord/local/RedundantBefore.java
ifesdjeen: Implemented this via bounds check instead, so this logic doesn't exist anymore.

accord-core/src/main/java/accord/local/RedundantStatus.java
ifesdjeen: Renamed to LOCALLY_INCOMPLETE

accord-core/src/main/java/accord/local/RedundantStatus.java
ifesdjeen: mb LOCALLY_INCOMPLETE?

accord-core/src/main/java/accord/local/CommandStore.java
ifesdjeen: Added a command store-wide `rebootstrapping` flag, which is used to check if we're  ready to coordinate, together with `safeToRead`

accord-core/src/main/java/accord/local/CommandStore.java
ifesdjeen: Rephrased, expanded

PR #214, opened 2025-06-30
Fixes 250628

belliottsmith:


accord-core/src/main/java/accord/local/durability/DurabilityService.java
```diff
@@ -73,64 +74,97 @@ public GlobalDurability global()
...
+            @Override public void timeout()
+            {
+                request.timeout();
+                unregister(request);
+            }
+        }, request.timeoutAt, MICROSECONDS);
+
+        synchronized (this)
+        {
+            if (!request.isDone()) // guard against unlikely scenario of timeout firing before we register here
```
ifesdjeen, line 155: If my understanding is correct, could still get triggered between isDone check and adding to request, should we check _after_ adding and unregistering then?

ifesdjeen (approved):

accord-core/src/main/java/accord/local/durability/DurabilityService.java
belliottsmith: Since the timeout is added first, it might be that the request gets timedout before we register it. If the timeout is invoked _after_ we register then it will successfully remove us, so nothing to worry about. But if it has already fired it won't have removed us (because we hadn't been added yet)

PR #217, opened 2025-07-07
Fixes 250705

belliottsmith:


accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -1169,46 +1170,42 @@ else if (!CommandsForKey.managesExecution(txnId) && command.hasBeen(Status.Stabl
...
             return AsyncChains.success(commandStore.executeInContext(commandStore,
                                                                      context(command, ASYNC),
                                                                      (SafeCommandStore safeStore) -> loadInternal(command, safeStore)));
         }
 
-        private AsyncChain<Command> apply(Command command)
+        private AsyncChain<Void> apply(TxnId txnId)
         {
             return AsyncChains.success(commandStore.executeInContext(commandStore,
-                                                                     context(command, SYNC),
```
ifesdjeen, line 1182: nit: `import static accord.local.KeyHistory.SYNC;` is now unused

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -434,7 +436,9 @@ public void epochClosed(Ranges ranges, long epoch)
                 Invariants.require(epoch < minEpoch(), "Could not find epoch %d. Min: %d, current: %d", epoch, minEpoch(), currentEpoch);
                 return; // notification came for an already truncated epoch
             }
-            while (epochs[i].recordClosed(ranges) && ++i < epochs.length) {}
+
+            while (!ranges.isEmpty() && i < epochs.length)
+                ranges = epochs[i++].recordClosed(ranges);
```
ifesdjeen, line 443: should we comment that we are ensuring that later epochs are transitively closing earlier ones?

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -260,7 +227,54 @@ else if (after.durability().isDurableOrInvalidated() && (force || !before.durabi
         }
     }
 
-    private TxnState maybeFetch(SafeCommandStore safeStore, TxnId txnId, Command after, TxnState state)
+    private TxnState updateHomeState(SafeCommandStore safeStore, Command after, @Nullable TxnState state)
+    {
+        Route<?> route = after.route();
```
ifesdjeen, line 232: I personally would annotate this method. It might be self-evident, but adding some descriptions of the branches, especially since we're querying some conditions more than once, could be beneficial.

ifesdjeen (approved):

ifesdjeen (approved):

PR #219, opened 2025-07-10
Accord Observability Improvements

belliottsmith:


accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -628,7 +658,7 @@ public static void postApply(SafeCommandStore safeStore, TxnId txnId, long t0, b
             return;
 
         safeCommand.applied(safeStore, forceApply);
-        safeStore.agent().eventListener().onApplied(command, t0);
+        safeStore.agent().localEvents().onApplied(safeStore, command, -1);
```
dcapwell: why `-1` here? wasn't it applied as this is `postApply`?

accord-core/src/main/java/accord/api/Agent.java
```diff
@@ -40,22 +40,11 @@
...
     default @Nullable Tracing trace(TxnId txnId, TraceEventType eventType) { return null; }
 
-    /**
-     * For use by implementations to decide what to do about successfully recovered transactions.
-     * Specifically intended to define if and how they should inform clients of the result.
-     * e.g. in Maelstrom we send the full result directly, in other impls we may simply acknowledge success via the coordinator
-     *
-     * Note: may be invoked multiple times in different places
-     */
-    void onRecover(Node node, Result success, Throwable fail);
```
dcapwell, line 57: for me: this was replaced by `accord.api.CoordinatorEventListener#onRecoveryStopped`?

accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
```diff
@@ -114,16 +114,6 @@ void setFailure(Throwable failure)
         // we may already be complete, as we may receive a failure from a later phase; but it's fine to redundantly mark done
         isDone = true;
         callback.accept(null, failure);
-        if (failure instanceof CoordinationFailed)
```
dcapwell, line 117: why was this removed?  Is the assumption that the caller to `coordinator` will see this failure so can handle that there?

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
```diff
@@ -168,9 +167,8 @@ void onPreAccepted(Topologies topologies, Timestamp executeAt, SortedListMap<Nod
             if (txnId.is(ExclusiveSyncPoint) && txnId.epoch() == executeAt.epoch())
                 withFlags = txnId.addFlag(HLC_BOUND);
             Deps deps = Deps.merge(oks.valuesAsNullableList(), oks.domainSize(), List::get, ok -> ok.deps);
-            if (tracker.hasFastPathAccepted())
-                adapter.execute(node, executor, topologies, route, Ballot.ZERO, FAST, CoordinationFlags.none(), txnId, txn, withFlags, deps, deps, callback);
```
dcapwell, line 172: why was fast path removed?

accord-core/src/main/java/accord/coordinate/CoordinateTransaction.java
```diff
@@ -128,8 +129,8 @@ void onPreAccepted(Topologies topologies, Timestamp executeAt, SortedListMap<Nod
                 // note: we merge all Deps regardless of witnessedAt. While we only need fast path votes,
                 // we must include Deps from fast path votes from earlier epochs that may have witnessed later transactions
                 // TODO (desired): we might mask some bugs by merging more responses than we strictly need, so optimise this to optionally merge minimal deps
+                node.agent().coordinatorEvents().onPreAccepted(txnId, deps, true);
```
dcapwell: wonder if enum would help here?  Had to go read the method signature to see that this is `isStable`

accord-core/src/main/java/accord/coordinate/CoordinateTransaction.java
```diff
@@ -138,21 +139,21 @@ else if (tracker.hasMediumPathAccepted() && txnId.hasMediumPath())
...
             proposeAndCommitInvalidate(node, executor, Ballot.ZERO, txnId, route.homeKey(), route, executeAt, callback);
-            node.agent().eventListener().onRejected(txnId);
+            node.agent().coordinatorEvents().onRejected(txnId);
             return;
         }
 
         Deps deps = Deps.merge(oks.valuesAsNullableList(), oks.domainSize(), List::get, ok -> ok.deps);
-        proposeAdapter().propose(node, executor, topologies, route, Accept.Kind.SLOW, Ballot.ZERO, txnId, txn, executeAt, deps, callback);
-        node.agent().eventListener().onSlowPathTaken(txnId, deps);
+        node.agent().coordinatorEvents().onPreAccepted(txnId, deps, false);
```
dcapwell: should we the enum showing fast/med/slow so we can pass this detail into `onPreAccepted`?  We had 3 methods before and they get replaced with this single one, which looks like we loose that level of detail?

accord-core/src/main/java/accord/coordinate/Propose.java
```diff
@@ -197,9 +197,10 @@ void onAccepted()
         //  In this case either id' needs to wait (which requires potentially more states like the alternative medium path)
         //  Or we must pick it up as an Unstable dependency here.
         Deps newDeps = mergeNewDeps();
-        Deps stableDeps = mergeDeps(newDeps);
-        if (kind == Kind.MEDIUM) adapter().execute(node, executor, acceptTracker.topologies(), route, ballot, MEDIUM, CoordinationFlags.none(), txnId, txn, executeAt, stableDeps, newDeps, callback);
-        else adapter().stabilise(node, executor, acceptTracker.topologies(), route, ballot, txnId, txn, executeAt, stableDeps, callback);
+        Deps deps = mergeDeps(newDeps);
+        node.agent().coordinatorEvents().onAccepted(txnId, ballot, deps, kind == Kind.MEDIUM);
```
dcapwell: similar comment to other method, maybe pass in `kind` rather than `boolean isStable`?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -286,8 +297,9 @@ public static AcceptOutcome accept(SafeCommandStore safeStore, SafeCommand safeC
         PartialDeps partialDeps = prepareDeps(validated, participants, command, deps);
         participants = prepareParticipants(validated, participants, command);
 
-        safeCommand.accept(safeStore, newSaveStatus, participants, ballot, executeAt, partialTxn, partialDeps, ballot);
+        Command accepted = safeCommand.accept(safeStore, newSaveStatus, participants, ballot, executeAt, partialTxn, partialDeps, ballot);
         safeStore.notifyListeners(safeCommand, command);
+        safeStore.agent().localEvents().onAccepted(safeStore, accepted);
```
dcapwell, line 301: should this happen before `notifyListeners`?  That can cause other txn to be updated which can fire off first, so the fact that this was `accepted` before that wouldn't be directly visible if you just looked at the events?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -298,12 +310,16 @@ public static AcceptOutcome notAccept(SafeCommandStore safeStore, SafeCommand sa
...
+                safeStore.agent().localEvents().onRejectNotAccept(safeStore, command, reject);
                 return reject;
+            }
         }
 
         logger.trace("{}: not accepted ({})", command.txnId(), status);
-        safeCommand.notAccept(safeStore, status, ballot);
+        Command notAccepted = safeCommand.notAccept(safeStore, status, ballot);
         safeStore.notifyListeners(safeCommand, command);
+        safeStore.agent().localEvents().onNotAccepted(safeStore, notAccepted);
```
dcapwell, line 321: same as above comment, should this be before `notifyListeners`?

accord-core/src/main/java/accord/coordinate/AbstractCoordinatePreAccept.java
belliottsmith: Yes, we had both duplication and inconsistency. I wanted to pick one of the two approaches, and the most consistent and simplest to reason about was to keep the handler of the request as responsible.

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
belliottsmith: It isn't possible for a sync point to take the fast path. It was a relic of inclusive sync points that have been removed for various reasons. I think there's some other vestiges, but this one intersected this change.

accord-core/src/main/java/accord/coordinate/CoordinateTransaction.java
belliottsmith: Sure, I can add an enum

accord-core/src/main/java/accord/coordinate/CoordinateTransaction.java
belliottsmith: This information is now provided to onExecuting which is a consistent place to provide it. I don't mind providing it here as well, but we don't actually know yet if the path will end up being taken successfully. isStable implies the fast path was taken, but the relevant fact there is that we will proceed to execution, and so these are the deps we may want to count (and this is what the parameter is used for in both the onPreAccept and onAccept method) - unless we want to count deps separately for each phase (which seems perhaps a bit over the top unless we find a need for it later).

accord-core/src/main/java/accord/coordinate/Propose.java
belliottsmith: We have enough kind variables, but I can introduce an IsStable enum with IS_STABLE and NOT_STABLE

accord-core/src/main/java/accord/coordinate/CoordinateTransaction.java
belliottsmith: I changed my mind, and just removed it entirely. It makes more sense to just do this in `onExecuting`

accord-core/src/main/java/accord/coordinate/Propose.java
belliottsmith: as above, changed mind and moved logic to onExecuting where it can be handled consistently

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: Sure. I think it makes sense either way, so happy to do it first

dcapwell (approved):

PR #222, opened 2025-07-15
Obs 250625 (part 1)

belliottsmith:


accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -333,7 +333,7 @@ public void clearBefore(SafeCommandStore safeStore, TxnId clearWaitingBefore, Tx
             {
                 // the command might be invalidated, which should be established on load, so simply load the command
                 TxnId txnId = state.txnId;
-                safeStore.commandStore().execute(txnId, safeStore0 -> {
+                safeStore.commandStore().execute(PreLoadContext.contextFor(txnId, "Clear Progress"), safeStore0 -> {
```
ifesdjeen, line 336: nit: will Java "intern" this for us? If not - should we create local runnable statics?

accord-core/src/main/java/accord/impl/AbstractSafeCommandStore.java
```diff
@@ -81,8 +81,18 @@ public PreLoadContext canExecute(PreLoadContext with)
         if (with.keys().domain() == Routable.Domain.Range)
             return with.isSubsetOf(this.context) ? with : null;
 
-        if (!context().keyHistory().satisfiesIfPresent(with.keyHistory()))
-            return null;
+        {
```
ifesdjeen, line 86: Should we unindent, or is this intentional?

accord-core/src/main/java/accord/local/LoadKeysFor.java
```diff
@@ -0,0 +1,28 @@
...
+
+package accord.local;
+
+/**
+ * For operations that need information associated with keys or ranges,
+ * this indicates whether the data will be queried, updated, or both.
+ */
+public enum LoadKeysFor
+{
+    WRITE, READ_WRITE, RECOVERY
```
ifesdjeen: I do not fully understand: should it be queried (read?) -> updated (write) -> both (read/write)? 

I am assuming recovery is as strong as read/write, could you explain the distinction?

accord-core/src/main/java/accord/local/PreLoadContext.java
```diff
@@ -111,12 +114,14 @@ default boolean isSubsetOf(PreLoadContext superset)
         Unseekables<?> keys = keys();
         if (!keys.isEmpty())
         {
-            KeyHistory requiredHistory = keyHistory();
+            LoadKeys requiredHistory = loadKeys();
```
ifesdjeen, line 117: I feel like we should rename `LoadKeys` now, since `LoadKeysFor` probably has meaning closer to this now, LoadKeys seems more like mode?

accord-core/src/main/java/accord/local/PreLoadContext.java
```diff
@@ -91,15 +93,16 @@ default void forEachId(Consumer<TxnId> consumer)
      */
     default Unseekables<?> keys() { return RoutingKeys.EMPTY; }
 
-    default KeyHistory keyHistory() { return NONE; }
+    default LoadKeys loadKeys() { return NONE; }
+    default LoadKeysFor loadKeysFor() { return WRITE; }
```
ifesdjeen, line 97: Should we have a default here even? Could be surprising for an implementer, easy to overlook. Having to implement it at least would make a person think / consider.

accord-core/src/main/java/accord/local/LoadKeysFor.java
ifesdjeen: Also, it would be good to specify that write can not operate on ranges, which seems to be the whole point of the optimization. I think it is not entirely obvious from this context, or even without very close inspection.

ifesdjeen (commented):
Do we need to make `AccordInteropApply` sync? Looked over other places, WRITE seems to be set correctly.

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
belliottsmith: not sure what you mean here, this one has a parameter - the `TxnId`. It is a shame we can't use the `TxnId` itself anymore, but probably better all round to have descriptions

accord-core/src/main/java/accord/impl/AbstractSafeCommandStore.java
belliottsmith: it was intentional, but I think it dates from a time when it had more variables perhaps, or some reason to want to isolate its scope more clearly. removing

accord-core/src/main/java/accord/local/LoadKeysFor.java
belliottsmith: I have added comments in a later patch. Let me bring them here.

accord-core/src/main/java/accord/local/PreLoadContext.java
belliottsmith: Well, it's when you can expect the keys to be loaded. I could call it LoadKeysMode, but e.g. LoadKeys.ASYNC seems pretty clear, and I'm not sure LoadKeysMode.ASYNC is much better. Open to suggestions for sure.

accord-core/src/main/java/accord/local/PreLoadContext.java
belliottsmith: I don't know. I am inclined to keep it as is, because we have a LOT of these classes and most don't need to load key information. But I don't feel incredibly strongly about it.

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
ifesdjeen: Sorry I have misplaced this comment it should have been for `(PreLoadContext.Empty) () -> "Key Barrier` and alike.

belliottsmith: Right, these should become statically allocated lambdas, as they are non-capturing. That's the reason for the slight hoops I jump through rather than having a static method. Whether it's worth it is another matter!

ifesdjeen (approved):

PR #223, opened 2025-07-15
CASSANDRA-20767: Move durability service startup

ifesdjeen:


accord-core/src/main/java/accord/local/Node.java
```diff
@@ -225,7 +225,6 @@ public Node(Id id, MessageSink messageSink,
         scheduler.recurring(() -> commandStores.forEachCommandStore(store -> store.progressLog.maybeNotify()), 1, SECONDS);
         scheduler.recurring(timeouts::maybeNotify, 100, MILLISECONDS);
         configService.registerListener(this);
-        configService.registerListener(durabilityService);
```
belliottsmith, line 229: is this going to break integration with the burn test?

belliottsmith (approved):

accord-core/src/main/java/accord/local/Node.java
ifesdjeen: (fixed in a feedback commit)

PR #224, opened 2025-07-16
Fixes 250715

belliottsmith:


accord-core/src/main/java/accord/local/MaxDecidedRX.java
```diff
@@ -37,9 +47,64 @@ private MaxDecidedRX(boolean inclusiveEnds, Object[] tree)
...
+    {
+        Invariants.require(txnId.is(Txn.Kind.ExclusiveSyncPoint));
+        // first check max, as if this is later we don't know that we can safely filter
+        TxnId maxDecidedId = max(keysOrRanges);
+        if (maxDecidedId.compareTo(txnId) < 0)
+            return min(keysOrRanges);
+        return null;
+    }
+
+    public @Nullable TxnId minDecidedDependencyId(Unseekable keyOrRange, TxnId txnId)
```
ifesdjeen, line 83: Intended to be unused?

accord-core/src/main/java/accord/coordinate/MaybeRecover.java
```diff
@@ -82,19 +84,26 @@ protected void onDone(Success success, Throwable fail)
         //  this can be helpful in mitigating flakiness and helping forward progress for large transactions spanning many shards
         if (fail != null)
         {
+            if (tracing != null)
```
ifesdjeen, line 87: should we generally format tracing in a single line?

ifesdjeen (approved):

accord-core/src/main/java/accord/coordinate/MaybeRecover.java
belliottsmith: I wondered about this and even briefly flirted with a Tracing.trace static method for making a single line, but I decided it actually maybe helps to both be consistent (and not all places would be suitable for a single line) and also it acts as a kind of comment for the reader so giving it some body/presence felt positive to me, and I preferred the aesthetic to single lines.

accord-core/src/main/java/accord/local/MaxDecidedRX.java
belliottsmith: It was used in a later patch that this got back ported from. I’ll remove it now.

belliottsmith: one of them at least is used by the implementation actually, so for now I will mark them both VisibleForImplementation

PR #225, opened 2025-07-18
Improve gc rebased

ifesdjeen:
(had to re-post the patch since turned out I did merge Accord side yesterday)


accord-core/src/main/java/accord/local/Node.java
```diff
@@ -310,6 +309,8 @@ public AsyncResult<?> persist(DurableBefore addValue, DurableBefore newValue)
                 Invariants.require(addValue.maxEpoch() <= epoch());
                 return wrap.persist(addValue, newValue);
             }
+
+            @Override public boolean shouldPersist(DurableBefore addValue, DurableBefore newValue) { return wrap.shouldPersist(addValue, newValue); }
```
ifesdjeen, line 313: re-did the patch, changed it so that we use `shouldPersist` instead of abstract class

accord-core/src/main/java/accord/utils/PersistentField.java
```diff
@@ -104,24 +103,31 @@ private AsyncResult<?> mergeAndUpdate(@Nullable Input inputValue, @Nonnull BiFun
...
+                    boolean upd = false;
+                    Saved latest = null;
+                    while (!complete.isEmpty() && pending.peek().id == complete.first())
+                    {
+                        latest = pending.poll().saving;
+                        complete.pollFirst();
+                        upd = true;
+                    }
+                    if (upd) set.accept(latest);
+                    result.setSuccess(null);
```
ifesdjeen: here, the "biggest" change is that we set success outside synchronized block to avoid callbacks being called there by accident

accord-core/src/main/java/accord/utils/PersistentField.java
```diff
@@ -92,36 +92,43 @@ public synchronized AsyncResult<?> mergeAndUpdate(@Nonnull Input inputValue)
...
+                    while (!complete.isEmpty() && pending.peek().id == complete.first())
+                    {
+                        latest = pending.poll().saving;
+                        complete.pollFirst();
+                        upd = true;
+                    }
+                    if (upd) set.accept(latest);
                 }
-                if (upd) set.accept(latest);
                 result.setSuccess(null);
```
ifesdjeen, line 125: here, the "biggest" change is that we set success outside synchronized block to avoid callbacks being called in it by accident

ifesdjeen:
@belliottsmith could you review?

belliottsmith (approved):

PR #226, opened 2025-07-21
Replay nondurable

belliottsmith:


accord-core/src/main/java/accord/impl/AbstractReplayer.java
```diff
@@ -23,21 +23,46 @@
...
+
+    protected boolean maybeShouldReplay(TxnId txnId)
+    {
+        return txnId.compareTo(minReplay) >= 0;
+    }
+
+    protected boolean shouldReplay(TxnId txnId, StoreParticipants participants)
+    {
+        Participants<?> search = participants.route();
+        if (search == null) search = participants.hasTouched();
```
ifesdjeen, line 62: Can't it be that hasTouched is also null, in which case we can opt for a global min?

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -119,6 +123,62 @@ public abstract class InMemoryCommandStore extends CommandStore
     private static final Logger logger = LoggerFactory.getLogger(InMemoryCommandStore.class);
     private static final boolean CHECK_DEPENDENCY_INVARIANTS = false;
 
+    public static class Snapshot extends AsyncResults.SettableResult<Snapshot>
+    {
+        private static final AtomicLong nextId = new AtomicLong();
+
+        private final long id = nextId.incrementAndGet();
```
ifesdjeen, line 130: nit: unused. still usable for debug?

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -1128,88 +1197,75 @@ public static void logDependencyGraph(CommandStore commandStore, TxnId txnId)
...
+                    // initialise basic state, but don't call safeStore.update so we don't initialise listeners etc
+                    GlobalCommand global = commandStore.commands.computeIfAbsent(txnId, GlobalCommand::new);
+                    global.value(command);
+                    Timestamp executeAt = command.executeAtIfKnown();
+                    if (executeAt != null)
+                        commandStore.commandsByExecuteAt.put(executeAt, global);
+                }
+            }
+
+            if (command == null || !maybeShouldReplay(txnId) || !shouldReplay(txnId, command.participants()))
```
ifesdjeen, line 1265: Could you explain why we still need to replay even if we do not apply? 

For some redundantBefore states we should be able to fully avoid loading, too, right?

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -690,50 +727,59 @@ public int hashCode()
...
+        boolean result = pending(kind).remove(active);
+        if (runBufferIndex < runBufferCount)
+            accept(safeStore);
+        return result;
     }
 
     void clearActive(TxnStateKind kind, TxnId txnId)
     {
-        active(kind).remove(txnId);
+        pending(kind).remove(txnId);
```
ifesdjeen, line 782: Should we rename the method to clearPending?

accord-core/src/main/java/accord/local/RedundantStatus.java
```diff
@@ -86,32 +88,46 @@ public enum Property
...
         // not persisted
         WAS_OWNED                          (false,  false, LT, LOCALLY_DEFUNCT),
-        NOT_OWNED                          (false, false, LT),
+        NOT_OWNED                          (false,  false, LT),
+
+        // added later, so vaguely out of order (but no specific order to this enum)
+
+        /**
+         * We have applied the preceding transactions durably to the store, so that we can safely truncate the Write
+         * information as we will not need to replay it to the store (but may need to replay it to
```
ifesdjeen: unfinished sentence?

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -351,15 +353,31 @@ public void clearBefore(SafeCommandStore safeStore, TxnId clearWaitingBefore, Tx
...
+        commandStore.maybeExecuteImmediately(() -> {
+            stopped = false;
+            accept(null);
+        });
+    }
+
+    @Override
+    public void stop()
+    {
+        stopped = true;
```
ifesdjeen, line 368: Shouldn't stopped be changed in command store only?

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -466,40 +484,67 @@ public void invalidIfUncommitted(TxnId txnId)
...
+        if (prevBuffer.length >= ArrayBuffers.MIN_BUFFER_SIZE)
+            cachedAny().forceDiscard(prevBuffer, prevCount);
+    }
+
+    private void maybeShrinkRunBuffer()
+    {
+        if (runBuffer.length >= (runBufferCount - runBufferIndex)/2)
+        {
+            int newCount = runBufferCount - runBufferIndex;
+            Object[] newBuffer = new Object[newCount + (newCount/2)];
```
ifesdjeen, line 545: Just wondering: why not try using `cachedAny` when shrinking?

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -466,40 +484,67 @@ public void invalidIfUncommitted(TxnId txnId)
...
+        runBuffer = newBuffer;
+        runBufferIndex = 0;
+        runBufferCount = newCount;
+        if (prevBuffer.length >= ArrayBuffers.MIN_BUFFER_SIZE)
+            cachedAny().forceDiscard(prevBuffer, prevCount);
+    }
+
+    private void maybeShrinkRunBuffer()
+    {
+        if (runBuffer.length >= (runBufferCount - runBufferIndex)/2)
```
ifesdjeen: I might be misreading, but did you mean we should do `* 2` here?  

run buffer length will always be larger than number of elements in it, so it seems like the idea was to shrink when there's 50%+ empty?

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -767,6 +813,16 @@ public void maybeNotify()
         }
     }
 
+    public void setMaxConcurrency(int maxConcurrency)
+    {
+        this.maxConcurrency = maxConcurrency;
```
ifesdjeen, line 819: nit: should we add strictly pos check?

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -574,32 +615,41 @@ private void processAwaitingEpoch(@Nullable SafeCommandStore safeStore, List<Run
...
             }
+
+            ++runBufferIndex;
+            try { run.accept(safeStore); }
+            catch (Throwable t) { node.agent().onUncaughtException(t); }
         }
+
+        if (runBuffer.length > ArrayBuffers.MIN_BUFFER_SIZE)
+            cachedAny().forceDiscard(runBuffer, runBufferCount);
+        runBufferIndex = runBufferCount = 0;
```
ifesdjeen, line 651: should this one be 

```
runBufferIndex = runBufferCount = 0;
```

accord-core/src/main/java/accord/impl/AbstractReplayer.java
belliottsmith: `hasTouched` can never be null, and in fact is forbidden from being empty (except for `Uninitialised/Erased/Vestigial`)

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
belliottsmith: This is emulating the behaviour in C* - we don't read from the journal in active processing with InMemoryCommandStore, so we have to restore the state of every transaction from the journal into memory. Some of these could be filtered here, but this is not how C* journal works, we would filter on load. 

We then filter all of the later steps, which is effectively not calling apply when appropriate.

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
belliottsmith: I did play around with this a few ways, but to avoid making modifications to the AsyncExecutor API (that would be sort of inconsistent with ExecutorPls), I simply made it volatile, so it is safe to change whenever.

We do need to rationalise our async/executor APIs eventually, it continues to cause some friction

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
belliottsmith: Because it might give us a much larger buffer than we want. We specifically want something smaller than we got last time we asked for something from the cache.

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
belliottsmith: Oh, whoops. Good catch.

ifesdjeen (approved):

PR #229, opened 2025-07-23
CASSANDRA-20785: Journal.TopologyUpdate should not store the local topology as it can be inferred from the global on

dcapwell:


accord-core/src/test/java/accord/impl/basic/DelayedCommandStores.java
```diff
@@ -115,7 +115,7 @@ public void validateShardStateForTesting(Journal.TopologyUpdate lastUpdate)
             shards[i++] = shard;
         }
 
-        loadSnapshot(new Snapshot(shards, lastUpdate.local, lastUpdate.global));
+        loadSnapshot(new Snapshot(shards, lastUpdate.global.forNode(nodeId()), lastUpdate.global));
```
belliottsmith: should this also call .trim?

dcapwell: sorry missed that, ill fix

dcapwell: pushed

belliottsmith (approved):

PR #230, opened 2025-07-24
Fixes 250724

belliottsmith:


accord-core/src/main/java/accord/local/Node.java
```diff
@@ -222,7 +222,10 @@ public Node(Id id, MessageSink messageSink,
         this.persistDurableBefore = new PersistentField<>(() -> durableBefore,
                                                           (input, prev) -> {
                                                               if (input.fullyContainedIn(prev))
+                                                              {
```
ifesdjeen: Great idea to check it here

ifesdjeen (approved):

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -325,9 +325,18 @@ public boolean hasCommandsForKey(RoutingKey key)
     @Override
     protected void ensureDurable(Ranges ranges, RedundantBefore onCommandStoreDurable)
     {
-        ((InMemoryAgent)agent).snapshot(this).begin((success, fail) -> execute((Empty)() -> "Report CommandStore Durable", safeStore -> {
-            safeStore.upsertRedundantBefore(onCommandStoreDurable);
-        }).begin(agent));
+        // TODO (required): we need a more general Replaying state
+        if (!CommandsForKey.reportLinearizabilityViolations())
```
ifesdjeen, line 329: hah, yes, maybe we should add a static flag and use it in CommandsForKey.reportLinearizabilityViolations?

accord-core/src/main/java/accord/local/Node.java
```diff
@@ -221,12 +221,10 @@ public Node(Id id, MessageSink messageSink,
         this.random = random;
         this.persistDurableBefore = new PersistentField<>(() -> durableBefore,
                                                           (input, prev) -> {
-                                                              if (input.fullyContainedIn(prev))
-                                                              {
-                                                                  Invariants.expect(prev.equals(DurableBefore.merge(input, prev)));
+                                                              DurableBefore next = DurableBefore.merge(input, prev);
```
ifesdjeen, line 224: to be honest I was hoping for a check that is cheaper than full durable before comparison but looks like fully contained in did not work?

accord-core/src/main/java/accord/impl/AbstractReplayer.java
```diff
@@ -66,30 +66,32 @@ protected boolean shouldReplay(TxnId txnId, StoreParticipants participants)
     protected void initialiseState(SafeCommandStore safeStore, TxnId txnId)
     {
         SafeCommand safeCommand = safeStore.unsafeGet(txnId);
-        Command command = safeCommand.current();
-        if (command.saveStatus().compareTo(SaveStatus.Stable) >= 0 && command.saveStatus().compareTo(PreApplied) <= 0)
```
ifesdjeen, line 70: just wondering why indent here? to avoid leaking `command` ref?

ifesdjeen (approved):

PR #231, opened 2025-07-25
Fixes 250725

belliottsmith:


accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -399,6 +399,7 @@ private void recover()
                                            .intersecting(route, id -> !recoverOks.containsKey(id.node));
         InferredFastPath fastPath;
         if (txnId.hasPrivilegedCoordinator() && coordinatorInRecoveryQuorum) fastPath = Reject;
+        else if (txnId.is(Txn.Kind.ExclusiveSyncPoint)) fastPath = Reject;
```
ifesdjeen, line 402: nit: can be 
```
        if ((txnId.hasPrivilegedCoordinator() && coordinatorInRecoveryQuorum) ||
            txnId.is(Txn.Kind.ExclusiveSyncPoint)) fastPath = Reject;
```

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -211,23 +212,39 @@ public void update(SafeCommandStore safeStore, TxnId txnId, Command before, Comm
...
-                    state.home().atLeast(safeStore, this, AwaitReadyToExecute, Queued);
-                    break;
-                case PreApplied:
-                    state.home().atLeast(safeStore, this, ReadyToExecute, Queued);
-                    break;
-            }
+            case Stable:
+                if (!after.acceptedOrCommitted().equals(Ballot.ZERO) || (before != null && before.saveStatus() == SaveStatus.Committed))
+                    state.home().atLeast(safeStore, this, Decided, NoneExpected);
+            default:
```
ifesdjeen, line 233: should this be a fall-through? probably worth adding a comment

ifesdjeen (approved):
also, inform decided has an extra import

accord-core/src/main/java/accord/local/cfk/Pruning.java
```diff
@@ -539,22 +542,70 @@ static TxnInfo[] removeRedundantById(TxnInfo[] byId, boolean hasRedundantLoading
...
+                // after restoring from snapshot we may have an inconsistency between the command state and the CFK state,
+                // and won't know that we should update the CFK because the command is already invalidated.
+                // So, to avoid having to read all CFK on startup and process any potentially invalidated transactions,
+                // we instead apply this filtering aggressively whenever we know the transaction cannot apply,
+                // and rely on the applied RX to correctly reject recovery of the transaction.
+                for (int i = pos ; i < appliedPos ; ++i)
+                {
+                    if (byId[i].compareTo(APPLIED) < 0)
+                    {
+                        if (byId[i].mayExecute())
```
ifesdjeen, line 574: nit: should we collapse nested ifs into a single if?

ifesdjeen (approved):

PR #232, opened 2025-07-30
Improve DefaultProgressLog

belliottsmith:


accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -754,45 +762,88 @@ public String reason()
...
     {
-        pending(kind).remove(txnId);
+        PendingTask pending = pending(kind).remove(txnId);
+        if (pending instanceof CallbackInvoker<?,?>)
+            active.remove(((CallbackInvoker<?, ?>) pending).id);
+    }
+
+    public void requeue(SafeCommandStore safeStore, TxnStateKind kind, TxnId txnId)
+    {
+        PendingTask pending = pending(kind).remove(txnId);
```
ifesdjeen: should we just call `clearPendingAndActive` here?

ifesdjeen (approved):

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
belliottsmith: yep

PR #233, opened 2025-07-31
Fix multiple overlapping bootstraps; improve replay time and startup behaviours

belliottsmith:
…ure the DurableBefore Majority condition holds transitively,

since later quorums may include the bootstrapping node which does not participate in the durability of preceding transactions Also Improve:
 - Apply MaxDecidedRX filtering to CommandsForKey RX dependencies
 - optimise AbstractRanges.sliceMinimal (burn test hotspot)
 - Don't start shard schedulers on topology changes if not already started
 - Only registerTransitive if waitingOnSync
 - Reserve DurabilityRequest until ShardDurability is started
 - Don't notify progress log if stopped

### DIFF
--- a/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
+++ b/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
@@ -74,7 +74,6 @@ import accord.primitives.AbstractRanges;
 import accord.primitives.AbstractUnseekableKeys;
 import accord.primitives.PartialDeps;
 import accord.primitives.Participants;
-import accord.primitives.RangeDeps;
 import accord.primitives.Ranges;
 import accord.primitives.Routable.Domain;
 import accord.primitives.RoutableKey;
@@ -98,7 +97,6 @@ import static accord.local.LoadKeys.NONE;
 import static accord.local.LoadKeysFor.WRITE;
 import static accord.local.RedundantStatus.Coverage.ALL;
 import static accord.local.StoreParticipants.Filter.LOAD;
-import static accord.primitives.Known.KnownRoute.MaybeRoute;
 import static accord.primitives.Routable.Domain.Key;
 import static accord.primitives.Routable.Domain.Range;
 import static accord.primitives.Routables.Slice.Minimal;
@@ -1274,30 +1272,5 @@ public abstract class InMemoryCommandStore extends CommandStore
 
             return apply(command);
         }
-    }
-
-    @Override
-    protected void registerTransitive(SafeCommandStore safeStore, RangeDeps rangeDeps)
-    {
-        RangesForEpoch rangesForEpoch = this.rangesForEpoch;
-        Ranges allRanges = rangesForEpoch.all();
-
-        TreeMap<TxnId, RangeCommand> rangeCommands = this.rangeCommands;
-        rangeDeps.forEachUniqueTxnId(allRanges, null, (ignore, txnId) -> {
-            GlobalCommand global = commands.get(txnId);
-            if (global != null && global.value().known().has(MaybeRoute))
-                return;
-
-            Ranges ranges = rangeDeps.ranges(txnId);
-            ranges = ranges.without(rangesForEpoch.coordinates(txnId));  // already coordinates, no need to replicate
-            if (ranges.isEmpty())
-                return;
-
-            ranges = ranges.slice(rangesForEpoch.allSince(txnId.epoch()), Minimal); // never coordinated, no need to replicate for dependency or recovery calculations
-            if (ranges.isEmpty())
-                return;
-
-            rangeCommands.computeIfAbsent(txnId, RangeCommand::new).add(ranges);
-        });
     }
 }
--- a/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
+++ b/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
@@ -143,11 +143,23 @@ public abstract class InMemoryCommandStore extends CommandStore
             }
         }
 
-        void save(CommandsForKey cfk)
+        void saveCallback(CommandsForKey cfk)
         {
-            commandsForKey.put(cfk.key(), Serialize.toBytesWithoutKey(cfk.maximalPrune()));
+            save(cfk);
             if (--waitingForCfk == 0)
                 trySuccess(this);
+        }
+
+        private void save(CommandsForKey cfk)
+        {
+            cfk = cfk.maximalPrune();
+            ByteBuffer serialized = Serialize.toBytesWithoutKey(cfk);
+            CommandsForKey roundTrip = Serialize.fromBytes(cfk.key(), serialized);
+            if (roundTrip != null)
+            {
+                Invariants.require(cfk.equalContents(roundTrip));
+                commandsForKey.put(cfk.key(), serialized);
+            }
         }
 
         public static AsyncResult<Snapshot> snapshot(InMemoryCommandStore commandStore)
@@ -167,7 +179,7 @@ public abstract class InMemoryCommandStore extends CommandStore
                 }
                 else
                 {
-                    snapshot.commandsForKey.put(cfk.key(), Serialize.toBytesWithoutKey(cfk.maximalPrune()));
+                    snapshot.save(cfk);
                 }
             }
             if (snapshot.waitingForCfk == 0)
@@ -592,7 +604,7 @@ public abstract class InMemoryCommandStore extends CommandStore
         public GlobalState<CommandsForKey> value(CommandsForKey value)
         {
             if (!pendingSnapshots.isEmpty() && !value.isLoadingPruned())
-                pendingSnapshots.forEach(snapshot -> snapshot.save(value));
+                pendingSnapshots.forEach(snapshot -> snapshot.saveCallback(value));
             return super.value(value);
         }
 
--- a/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
+++ b/accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
@@ -58,6 +58,7 @@ import accord.local.CommandStore;
 import accord.local.CommandStores.RangesForEpoch;
 import accord.local.CommandSummaries;
 import accord.local.Commands;
+import accord.local.MaxDecidedRX;
 import accord.local.NodeCommandStoreService;
 import accord.local.PreLoadContext;
 import accord.local.PreLoadContext.Empty;
@@ -83,7 +84,6 @@ import accord.primitives.Timestamp;
 import accord.primitives.Txn;
 import accord.primitives.Txn.Kind.Kinds;
 import accord.primitives.TxnId;
-import accord.primitives.Unseekable;
 import accord.primitives.Unseekables;
 import accord.utils.Invariants;
 import accord.utils.async.AsyncChain;
@@ -114,7 +114,6 @@ import static accord.primitives.Status.Stable;
 import static accord.primitives.Status.Truncated;
 import static accord.primitives.Txn.Kind.EphemeralRead;
 import static accord.primitives.Txn.Kind.ExclusiveSyncPoint;
-import static accord.primitives.Txn.Kind.Read;
 import static accord.utils.Invariants.illegalState;
 import static java.lang.String.format;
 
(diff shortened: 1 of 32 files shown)


accord-core/src/main/java/accord/local/durability/ShardDurability.java
```diff
@@ -346,6 +346,9 @@ synchronized void retry(Ranges ranges)
 
         synchronized void start()
         {
+            if (!isStarted())
```
ifesdjeen: Just for readability, could you maybe make it `if (!ShardDurability.this.isStarted())` ? Probably superfluous, but I think it might be useful to disambiguate.

accord-core/src/main/java/accord/local/Bootstrap.java
```diff
@@ -135,20 +138,24 @@ void start(SafeCommandStore safeStore)
...
-                     .flatMap(i -> i)
-                     .begin(this);
+            CommandStore commandStore = safeStore.commandStore();
+            node.durability()
+                // we first make sure the sync point is durable to a majority, since any later durability conditions
+                // this node participates in will not guarantee a quorum for preceding transactions
+                .sync("Bootstrap " + commitRanges + " for " + safeStore.commandStore(), globalSyncId, commitRanges, NoLocal, Quorum, 1L, TimeUnit.HOURS)
+                .flatMap(success -> commandStore.build((PreLoadContext.Empty) () -> "Start Bootstrap RX", safeStore0 -> {
+                    // we submit a separate execution so that we know markBootstrapping is durable before we initiate the fetch
+                    store.markBootstrapping(safeStore0, globalSyncId, commitRanges);
```
ifesdjeen: Do we maybe want to mark bootstrapping before we wait for durability majority? This used to be the first step we were making previously if I understand the left hand side correctly.

accord-core/src/main/java/accord/local/CommandSummaries.java
```diff
@@ -60,185 +62,279 @@ enum SummaryStatus
...
-            protected final Timestamp maxTxnId;
-            @Nullable protected final TxnId findAsDep;
+        protected final RedundantBefore redundantBefore;
+        protected final MaxDecidedRX maxDecidedRX;
+        protected final Unseekables<?> searchKeysOrRanges;
+        // TODO (expected): separate out Kinds we need before/after primaryTxnId/executeAt
+        protected final Kinds testKind;
+        protected final TxnId primaryTxnId, findAsDep, minTxnId, minDecidedId;
+        protected final Timestamp maxTxnId;
+//        protected final TxnId minDecidedId;
```
ifesdjeen: nit: minDecidedId already added as a field above, can be cleaned up here

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -666,11 +666,33 @@ protected void updatedRedundantBefore(SafeCommandStore safeStore, RedundantBefor
...
+            if (e.getKey() > syncId.epoch())
+                break;
+
+            Ranges remaining = e.getValue().ranges;
+            boolean intersects = remaining.intersects(ranges);
+            if (intersects)
+                return true;
+        }
+
+        return true;
```
ifesdjeen: Should this return false maybe, since we didn't get any intersection matches that would return true?

accord-core/src/main/java/accord/local/CommandSummaries.java
```diff
@@ -60,185 +62,279 @@ enum SummaryStatus
...
+        public boolean is(IsDep isDep)
+        {
+            return (encoded >> IS_DEP_SHIFT) == isDep.ordinal();
         }
 
-        public Summary slice(Ranges slice)
+        public IsDep isDep()
         {
-            return new Summary(txnId, executeAt, status, participants.slice(slice, Minimal), dep, findAsDep);
+            int ordinal = encoded >> IS_DEP_SHIFT;
```
ifesdjeen: should we maybe first check for equality with min value and then bit shift?

accord-core/src/main/java/accord/local/CommandSummaries.java
```diff
@@ -60,185 +62,279 @@ enum SummaryStatus
...
-            {
-                return ifRelevant(cmd.txnId(), cmd.executeAtOrTxnId(), cmd.saveStatus(), cmd.participants(), cmd.partialDeps());
-            }
+        public boolean isRelevant(CommandsForKey cfk)
+        {
+            if (cfk == null || cfk.size() == 0)
+                return false;
 
-            private boolean isEligibleDep(SummaryStatus status, TxnId findAsDep, TxnId txnId, Timestamp executeAt)
+            // NOTE: we CANNOT safely filter on first element, as we may have pruned dependencies we need to witness
```
ifesdjeen, line 227: It is not clear to me here what you mean by first element; also because we then filter by _last_

accord-core/src/main/java/accord/local/CommandSummaries.java
```diff
@@ -60,185 +62,279 @@ enum SummaryStatus
...
-                    return null;
+                intersecting = newIntersecting;
+            }
 
-                // start in search key domain, since this is what we consult to decide if can be recovered
-                Unseekables<?> intersecting = searchKeysOrRanges.intersecting(touches, Minimal);
-                if (intersecting.isEmpty())
+            if (mayFilterAsDecided)
+            {
+                TxnId minDecidedId = minDecidedDependencyId(maxDecidedRX, intersecting, primaryTxnId);
```
ifesdjeen, line 346: Variable shadowing looks intentional in both cases, but maybe rename to include "interesting" or something similar?

accord-core/src/main/java/accord/local/Bootstrap.java
belliottsmith: The reason I haven't, and I should add a comment to this effect, is that I don't want us to participate in any quorum until after a majority quorum has been reached (so that we aren't counted towards the DurableBefore Majority condition, as we don't really count for anything pre-bootstrap, as we don't wait for it). 

There is certainly a cleaner and more robust way to do this, where we track this explicitly and ensure this information is incorporated when counting votes. But, this suffices for now.

belliottsmith: To be fair, it might be a relatively small thing to add that more robust logic as part of this patch. I'll mull on it overnight.

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: you're right, though this would have been fine (just caused some unnecessary work). I've actually tweaked this a little in the follow-up that addresses potential issues with `registerTransitive`.

accord-core/src/main/java/accord/local/CommandSummaries.java
belliottsmith: hrm. I didn't realise I had changed this in this patch. something must have gone wrong with a rebase. I don't quite follow what you mean, but let me check what happened in some rebase before we figure that out.

accord-core/src/main/java/accord/local/CommandSummaries.java
belliottsmith: I mean we are filtering based on the last element in the CFK, and while it might seem like we could do equivalent filtering based on the first/min element and maxTxnId, we can't. I will tweak the wording.

accord-core/src/main/java/accord/local/CommandSummaries.java
belliottsmith: Oh, I remember. This was intertwined with the changes to CFK filtering.

belliottsmith: Ok, I get it. Have pushed a change.

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -666,7 +663,62 @@ protected void updatedRedundantBefore(SafeCommandStore safeStore, RedundantBefor
...
+        if (waitingOnSync.isEmpty())
+            return;
+
+        for (Map.Entry<Long, WaitingOnSync> e : waitingOnSync.entrySet())
+        {
+            if (e.getKey() > syncId.epoch())
+                break;
+
+            Ranges remaining = e.getValue().waitingOn.without(ranges);
+            if (e.getValue().waitingOn != remaining)
```
ifesdjeen, line 700: nit: we can just always assign: if it's equal, it will remain unchanged, and new value will get assigned otherwise

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -420,30 +429,32 @@ protected final InMemorySafeStore createSafeStore(PreLoadContext context, Ranges
...
-                case Range:
-                    // load range cfks here
-                    break;
+                for (RoutingKey key : (AbstractUnseekableKeys)unseekables)
+                    commandsForKey.put(key, commandsForKey(key).createSafeReference());
+            }
+            else
+            {
+                CommandSummaries.SummaryLoader loader = CommandSummaries.SummaryLoader.loader(unsafeGetRedundantBefore(), unsafeGetMaxDecidedRX(), context);
+                for (GlobalCommandsForKey global : this.commandsForKey.values())
```
ifesdjeen, line 444: There should be a way to avoid scanning over all keys linearly here, maybe with a range multi-map or something? Not saying that we should do this in this patch, but maybe worth adding a TODO?

accord-core/src/main/java/accord/local/cfk/Serialize.java
```diff
@@ -77,6 +77,7 @@ public class Serialize
     private static final int HAS_BALLOT_HEADER_BIT_SHIFT = 2;
     private static final int HAS_STATUS_OVERRIDES_HEADER_BIT = 0x8;
     private static final int HAS_STATUS_OVERRIDES_HEADER_BIT_SHIFT = 3;
+    private static final int HAS_MISSING_DEPS_FLAGS_HEADER_BIT = 0x10;
```
ifesdjeen, line 80: nit: maybe add something like "bit 4 is set if there are any queries with missing deps flags" into preamble header description

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -666,7 +663,62 @@ protected void updatedRedundantBefore(SafeCommandStore safeStore, RedundantBefor
...
+        if (waitingOnSync.isEmpty())
+            return;
+
+        for (Map.Entry<Long, WaitingOnSync> e : waitingOnSync.entrySet())
+        {
+            if (e.getKey() > syncId.epoch())
+                break;
+
+            Ranges unmark = e.getValue().waitingOnDurable.slice(ranges, Minimal);
+            if (!unmark.isEmpty())
```
ifesdjeen, line 716: should we disassociate the key if empty?

accord-core/src/main/java/accord/local/CommandSummaries.java
```diff
@@ -60,185 +62,279 @@ enum SummaryStatus
...
-                this.testKind = testKind;
-                this.minTxnId = minTxnId;
-                this.maxTxnId = maxTxnId;
-                this.findAsDep = findAsDep;
-            }
+        public static <L extends SummaryLoader> L loader(RedundantBefore redundantBefore, MaxDecidedRX maxDecidedRX, TxnId primaryTxnId, LoadKeysFor loadKeysFor, Unseekables<?> keysOrRanges, Factory<L> factory)
+        {
+            Invariants.require(primaryTxnId != null);
+            TxnId minTxnId = redundantBefore.min(keysOrRanges, Bounds::gcBefore);
+            Timestamp maxTxnId = loadKeysFor == RECOVERY || !primaryTxnId.is(ExclusiveSyncPoint) ? Timestamp.MAX : primaryTxnId;
```
ifesdjeen, line 200: nit: maybe best remove `!` and flip arguments?

ifesdjeen: or perhaps:

```
       185 +              Timestamp maxTxnId;
       186 +              TxnId findAsDep;
       187 +              Kinds kinds;
       188 +              if (loadKeysFor == RECOVERY)
       189 +              {
       190 +                  maxTxnId = Timestamp.MAX;
       191 +                  findAsDep = primaryTxnId;
       192 +                  kinds = primaryTxnId.witnesses().or(primaryTxnId.witnessedBy());
       193 +              }
       194 +              else
       195 +              {
       196 +                  maxTxnId = primaryTxnId.is(ExclusiveSyncPoint) ? primaryTxnId : Timestamp.MAX;
       197 +                   findAsDep = null;
       198 +                  kinds = primaryTxnId.witnesses().or(Nothing);
       199 +              }

```

ifesdjeen (approved):

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
belliottsmith: This doesn't really need to be a range multi map, we can already search it just fine if we want to by iterating over the keys and then slicing to a subMap - we do it elsewhere, just wasn't super important here. Happy to do it now if you like.

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: I don't follow. Do you mean remove the key from the collection?

PR #237, opened 2025-08-08
Enrich Durability

belliottsmith:
Enrich Durability with each phase, so we can both prune unapplied dependencies to mitigate replicas that are behind causing Deps growth,
    and more reliably avoid initialising recovery progress log state of transactions that cannot yet make progress
     - Fix infer invalid in MaybeRecover and FetchData
     - MaybeRecover sometimes aborts before home shard knows outcome
     - Fix epoch sync with VisibilitySyncPoint
     - Retired implies synced
     - Introduce LOCALLY_LOST RedundantStatus


accord-core/src/main/java/accord/local/RedundantBefore.java
```diff
@@ -351,7 +357,7 @@ private static Timestamp maybeClearStaleUntilAtLeast(Timestamp staleUntilAtLeast
         {
             for (int i = 0 ; staleUntilAtLeast != null && i < bounds.length && bounds[i].compareTo(staleUntilAtLeast) > 0 ; ++i)
             {
-                if (any(statuses[i], PRE_BOOTSTRAP))
+                if (any(0xFFFF & statuses[i], PRE_BOOTSTRAP))
```
ifesdjeen: could you explain why mask it if it's short and we have an overload for short?

accord-core/src/main/java/accord/local/RedundantBefore.java
```diff
@@ -360,7 +366,7 @@ private static Timestamp maybeClearStaleUntilAtLeast(Timestamp staleUntilAtLeast
         public Bounds with(TxnId newBound, SomeStatus addStatus)
         {
             // TODO (desired): introduce special-cased faster merge for adding a single value
-            return merge(range, this, new Bounds(range, Long.MIN_VALUE, Long.MAX_VALUE, new TxnId[] { newBound }, new short[] { addStatus.encoded }, null));
+            return merge(range, this, new Bounds(range, Long.MIN_VALUE, Long.MAX_VALUE, new TxnId[] { newBound }, new short[] { (short)addStatus.encoded }, null));
```
ifesdjeen, line 368: Don't we want to mask here while down casting?

accord-core/src/main/java/accord/local/DepsCalculator.java
```diff
@@ -137,13 +135,14 @@ public Deps calculate(SafeCommandStore safeStore, TxnId txnId, Participants<?> t
...
             {
-                Invariants.require(maxRedundantBefore.is(ExclusiveSyncPoint));
+                Invariants.require(maxRedundantBefore.isSyncPoint());
                 return null;
             }
         }
 
         // NOTE: ExclusiveSyncPoint *relies* on STARTED_BEFORE to ensure it reports a dependency on *every* earlier TxnId that may execute (before or after it).
         MinDependencyCalculator minDepCalc = null;
+        // the main different between RX and RV is whether we apply this filtering
```
ifesdjeen: nit: difference?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -1097,38 +1100,40 @@ public static Command setDurability(SafeCommandStore safeStore, SafeCommand safe
...
         participants = updated.participants();
         if (executeAt != null && command.status().hasBeen(Committed) && !command.executeAt().equals(executeAt))
             safeStore.agent().onInconsistentTimestamp(command, command.asCommitted().executeAt(), executeAt);
 
-        if (command.durability().compareTo(durability) < 0)
+        updated = safeCommand.update(safeStore, updated.updateDurability(newDurability));
+        TxnId txnId = command.txnId();
+        DependencyElision updates = OFF;
+        if (newDurability.isDurable() && !oldDurability.isDurable()) updates = IF_DURABLY_PREAPPLIED;
+        else if (newDurability.isDurablyCommitted() && !durability.isDurablyCommitted()) updates = IF_DURABLY_COMMITTED;
```
ifesdjeen: Should `durability.isDurablyCommitted()` actually be `oldDurability`?

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -1097,38 +1100,40 @@ public static Command setDurability(SafeCommandStore safeStore, SafeCommand safe
         if (command.is(Truncated))
             return command;
 
-        if (command.durability().compareTo(durability) >= 0)
+        Durability oldDurability = command.durability();
```
ifesdjeen, line 1102: Should we rename `durability` in `setDurability` to `durabilityUpdate` or `durabilityChange` ?

accord-core/src/main/java/accord/local/cfk/CommandsForKey.java
```diff
@@ -1604,18 +1639,21 @@ public CommandsForKey updateUniqueHlc(long minUniqueHlc)
...
-        return new CommandsForKey(key, bounds, byId, minUndecidedById, maxAppliedPreBootstrapWriteById, committedByExecuteAt, maxAppliedWriteByExecuteAt, minUniqueHlc, loadingPruned, prunedBeforeById, unmanageds);
+        return new CommandsForKey(key, bounds, byId, minUndecidedById, maxAppliedPreBootstrapWriteById, committedByExecuteAt, maxAppliedWriteByExecuteAt, minUniqueHlc, loadingPruned, prunedBeforeById, unmanageds, true);
     }
 
-    public CommandsForKey setDurable(TxnId txnId)
+    public CommandsForKey setDurable(TxnId txnId, Durability durability)
     {
         TxnInfo txn = get(txnId);
-        if (txn == null || !txn.is(APPLIED_NOT_DURABLE))
+        if (txn == null || !durability.isDurablyCommitted())
```
ifesdjeen, line 1648: I might be misunderstanding: do we _not_ want to set durable in CFK if not yet durably committed? Should we even be calling this method then?

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -219,60 +223,26 @@ public void update(SafeCommandStore safeStore, TxnId txnId, Command before, Comm
...
-        TxnState state = get(txnId);
-        if (state != null && state.isHomeInitialised())
-            state.home().atLeast(safeStore, this, Decided, NoneExpected);
-    }
-
-    private void updateHomeState(SafeCommandStore safeStore, TxnState state, Command before, Command after)
-    {
-        switch (after.saveStatus())
+        boolean updateWaiting = force || beforeSaveStatus != afterSaveStatus;
+        boolean updateHome = updateWaiting || (afterRoute != null && beforeRoute == null) || !before.durability().isAtLeast(after.durability());
```
ifesdjeen: should the third condition be `after.durability().isAtLeast(before.durability()`?

ifesdjeen (approved):

accord-core/src/main/java/accord/local/RedundantBefore.java
belliottsmith: masking to the range of the down-casted type is a no-op

accord-core/src/main/java/accord/local/Commands.java
belliottsmith: good catch

accord-core/src/main/java/accord/local/cfk/CommandsForKey.java
belliottsmith: I don't quite follow, but if you're asking whether this method should not have been invoked, the answer is we already should not invoke it if we think it won't help, but the method signature doesn't specify what Durability it would be valid to invoke it with, and we don't want to corrupt the state if we for some reason invoke setDurability with a Durability that doesn't imply anything useful for CommandsForKey, so the most sensible thing IMO is to simply do nothing if the parameter doesn't tell us anything we want to know.

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
belliottsmith: I think we would need an `isAtMost` method if we wanted it that way around.

accord-core/src/main/java/accord/topology/Topologies.java
```diff
@@ -88,9 +88,16 @@ default boolean isEmpty()
      */
     SortedArrayList<Id> nodes();
 
-    default SortedArrayList<Id> staleNodes()
+    default SortedArrayList<Id> staleOrRemovedIds()
```
ifesdjeen, line 91: Won't this break if we have a node that got removed but then re-added?

ifesdjeen (approved):

PR #238, opened 2025-08-09
Fixes 250809

belliottsmith:


accord-core/src/main/java/accord/coordinate/FetchData.java
```diff
@@ -90,6 +90,13 @@ public FetchRequest(SequentialAsyncExecutor executor, Known fetch, TxnId txnId,
             this.reportTo = reportTo;
             this.tracing = tracing;
         }
+
+        @Override
+        public void accept(FetchResult success, Throwable fail)
+        {
+            if (fail != null && tracing != null) tracing.trace(null, "Failed fetch: %s", Tracing.format(fail));
```
ifesdjeen, line 97: nit: should this use the new wrap instead?

ifesdjeen (approved):

accord-core/src/main/java/accord/coordinate/FetchData.java
belliottsmith: I introduced wrap for this purpose, but this avoids an allocation. I left wrap for future use cases (by which I mean when we have time to clean up other call sites).

PR #243, opened 2025-08-18
Fixes 250816

belliottsmith:


accord-core/src/main/java/accord/local/MaxDecidedRX.java
```diff
@@ -22,103 +22,198 @@
...
 
-    public @Nullable TxnId minDecidedDependencyId(Unseekables<?> keysOrRanges, TxnId txnId)
+//
+//    TxnId minAny(Routables<?> keysOrRanges)
+//    {
+//        return noneIfNull(foldlWithDefault(keysOrRanges, Entry::minAny, null, TxnId.NONE));
+//    }
+//
+//    TxnId maxAny(Routables<?> keysOrRanges)
+//    {
```
ifesdjeen: nit: commented code

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
```diff
@@ -78,12 +86,63 @@
...
+
+        @Override
+        public void onFailure(Id from, Throwable failure)
+        {
+        }
+
+        void informStable()
+        {
+            Invariants.require(hasReachedQuorum());
+            hasInformedDecidedOrSucceeded = true;
```
ifesdjeen: nit: should this call `setSucceeded`?

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
```diff
@@ -78,12 +86,63 @@
...
+        public StableTracker(Topologies topologies)
+        {
+            super(topologies);
+        }
+
+        @Override
+        public void onSuccess(Id from, ReadReply reply)
+        {
+            if ((reply.isOk() || reply == Waiting) && RequestStatus.Success == recordSuccess(from) && informOnSuccess)
+                informStable();
```
ifesdjeen, line 110: looks like we might be able to call `informStable` more Ethan once as nodes keep responding after reaching quorum. Should we call it just once to avoid sending InformDurable more than once?

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
```diff
@@ -258,8 +387,22 @@ protected void onDone(Success success, Throwable failure)
...
+            if (sendOnlyReadStableMessages())
+            {
+                // send additional stable messages to record the transaction outcome
+                Commit.Kind kind = commitKind();
+                if (!candidates.isEmpty())
+                {
+                    for (int i = 0, size = candidates.size() ; i < size ; ++i)
+                        sendStableOnly(candidates.get(i), kind);
+                }
+                if (unstableFastReads != null && sendOnlyReadStableMessages())
```
ifesdjeen: nit: covered by the above if

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
belliottsmith: No, that was for the outer class to invoke on its success. But perhaps I will simply call this isDone and setDone

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
belliottsmith: we only return `RequestStatus.Success` once

ifesdjeen (approved):

PR #244, opened 2025-08-21
Observability Improvements 250802

belliottsmith:


accord-core/src/main/java/accord/local/Node.java
```diff
@@ -998,6 +970,39 @@ public final long currentStamp()
...
+        long startedAtNanos = time.elapsed(NANOSECONDS);
+        long nextId = nextCoordinationId.get();
+        if (startedAtNanos >= nextId && nextCoordinationId.compareAndSet(nextId, startedAtNanos))
+            return startedAtNanos;
+        return nextCoordinationId.incrementAndGet();
+    }
+
+    public void register(Coordination coordination)
+    {
+        coordinations.register(coordination);
```
ifesdjeen, line 984: Could you wrap calls to coordinations register/unregister into try/catch? Just to prevent txns failing in case something goes wrong in coordinations.

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -417,6 +417,9 @@ public void syncComplete(Id node, long epoch)
          */
         public Ranges epochClosed(Ranges ranges, long epoch)
         {
+            if (epochs.length == 0)
```
ifesdjeen, line 420: Did you see any test failures related to this? I am concerned if we have problems with startup order now, since usually we should not report closed epochs while epochs is still empty. I thought I have precluded this, but maybe not everywhere? Another potential problem could be that we won't report the epoch as closed again.

accord-core/src/test/java/accord/impl/TestAgent.java
```diff
@@ -158,6 +158,12 @@ public long slowCoordinatorDelay(Node node, SafeCommandStore safeStore, TxnId tx
         return units.convert(1L, SECONDS);
     }
 
+    @Override
+    public boolean isSlowCoordinator(long elapsed, TimeUnit units, TxnId txnId, int attempt)
+    {
+        return units.toSeconds(elapsed) > 1;
```
ifesdjeen, line 164: Won't this round it down too much? Should we maybe do `>=`. Maybe does not matter that much for test agent though.

accord-core/src/main/java/accord/impl/RequestCallbacks.java
```diff
@@ -110,7 +112,7 @@ public void onExpire(long nowMicros)
             @Override
             public void onExpire(long nowMicros)
             {
-                safeInvoke(RegisteredCallback::unsafeOnFailure, new accord.coordinate.Timeout(null, null));
+                safeInvoke(RegisteredCallback::unsafeOnFailure, (Throwable)null);
```
ifesdjeen, line 116: Should w maybe use some static variable in place of timeout if the idea here was to avoid additional allocations?

accord-core/src/main/java/accord/coordinate/ReadCoordinator.java
```diff
@@ -301,4 +308,74 @@ protected RequestStatus trySendMore()
         contact.forEach(this::contact);
         return status;
     }
+
+    @Override
+    public SortedList<Id> inflight()
+    {
+        Node.Id[] ids = new Node.Id[inflight.size()];
```
ifesdjeen, line 315: Should we just use int arrays here? Especially if we're using this for debug output only

accord-core/src/main/java/accord/coordinate/AbstractCoordination.java
```diff
@@ -0,0 +1,343 @@
...
+        if (replyState == null)
+            return null;
+
+        Object[] replies = replyState.clone();
+        for (int i = 0 ; i < replies.length ; ++i)
+        {
+            if (replies[i] != null && !(replies[i] instanceof accord.messages.Reply))
+                replies[i] = null;
+        }
+        return new SortedListMap<Node.Id, Ok>(nodes, replies, replyCount);
```
ifesdjeen: nit: I think this can even be `new SortedListMap<>` but generally reply type is erased, do you want to specify here?

accord-core/src/main/java/accord/coordinate/AbstractCoordination.java
```diff
@@ -0,0 +1,343 @@
...
+import accord.utils.async.AsyncChain;
+import accord.utils.async.Cancellable;
+
+// TODO (expected): move message sending here, so we can invalidate any pending callbacks when we advance the state machine
+public abstract class AbstractCoordination<Result, Reply extends accord.messages.Reply, Ok> extends AbstractSimpleCoordination implements Callback<Reply>
+{
+    private BiConsumer<? super Result, Throwable> callback;
+    private final SortedArrayList<Node.Id> nodes;
+    private final SimpleBitSet expectingReply;
+    private Object[] replyState;
```
ifesdjeen, line 54: It is unclear what is `Ok`, and whether `Ok` is always cancellable, and what this `Object` can be safely cast to.

accord-core/src/main/java/accord/coordinate/Recover.java
```diff
@@ -244,17 +212,57 @@ public void onSuccess(Id from, RecoverReply reply)
...
+            {
+                callback.accept(ProgressToken.APPLIED, null);
+                node.agent().coordinatorEvents().onRecoveryStopped(node, txnId, ballot, success, failure);
+            }
+            else if (failure instanceof Redundant)
+            {
+                Timestamp committedExecuteAt = ((Redundant) failure).committedExecuteAt;
+                if (tracing != null)
+                    tracing.trace(null, "Recover found Redundant; retrying with known committedExecuteAt " + committedExecuteAt);
+                retry(committedExecuteAt, callback);
```
ifesdjeen, line 248: Could you add an invariant that current recover is done before calling retry? Maybe superfluous.

ifesdjeen (approved):

PR #245, opened 2025-08-22
Add a way to grab topology iif it exists

ifesdjeen:
Relax assertion after allowing to load epochs at a later stage during startup


belliottsmith:
Think this is defunct; closing for now

PR #246, opened 2025-09-03
CASSANDRA-20883: bitset changes to support cassandra serializer needs

dcapwell:


.gitignore
```diff
@@ -4,3 +4,10 @@ build/
 accord-core/build/
 accord-maelstrom/build/
 .rat-excludes.txt
+
```
dcapwell, line 7: this was pulled from CASSANDRA-20828, its super annoying now that i started to use NeoVim more, need this so we don't think the git module is dirty

belliottsmith:
I think this is defunct; closing for now

PR #247, opened 2025-09-04
Fix:

belliottsmith:
- removeRedundantMissing could erroneously remove missing dependencies occurring after the new appliedBeforeIndex
 - Invariant failure for some propagate calls supplying 'wrong' addRoute
 - Disambiguate nextTxnId and nextTxnIdWithDefaultFlags
 - All pending topology notifications must be propagated to a newly published epoch, not only the pending notification for that epoch


iamaleksey (approved):
+1

PR #249, opened 2025-09-08
CASSANDRA-20883: fixed a bug with the bit set when querying outside of the size

dcapwell:


accord-core/src/main/java/accord/utils/LargeBitSet.java
```diff
@@ -320,7 +320,7 @@ public final int nextSetBitBefore(int i, int exclBound, int ifNotFound)
     private int nextSetBitInternal(int i, int exclIndexBound, int ifNotFound)
     {
         Invariants.requireArgument(i >= 0);
-        Invariants.requireArgument(i <= size());
```
dcapwell, line 323: the interface doesn't expose a way to know the size, so you need to rely on `nextSetBitInternal` to know when to complete iterating, so rather than failing return `-1`

belliottsmith: My only concern here is that this is inconsistent with get/set etc. We should either consistently change all of these methods to support indexes outside of the range, or we should retain the existing behaviour and simply expose `capacity()` in the interface. I don't have a strong opinion, though I lean towards the `capacity()` approach as it is is more likely than not to be a bug to access outside of the allocated range.

dcapwell:
this was replaced by https://github.com/apache/cassandra-accord/pull/264

PR #251, opened 2025-09-16
Rebootstrap

ifesdjeen:


accord-core/src/main/java/accord/local/cfk/Serialize.java
```diff
@@ -326,7 +326,7 @@ private static ByteBuffer unsafeToBytesWithoutKey(CommandsForKey cfk)
                               | (executeAtCount          > 0  ? HAS_EXECUTE_AT_HEADER_BIT         : 0)
                               | (ballotCount             > 0  ? HAS_BALLOT_HEADER_BIT             : 0)
                               | (overrideCount           > 0  ? HAS_STATUS_OVERRIDES_HEADER_BIT   : 0)
-                              | (cfk.bootstrappedAt() != null ? HAS_BOOTSTRAPPED_AT_HEADER_BIT    : 0)
+                              | (cfk.readyAt() != null ? HAS_BOOTSTRAPPED_AT_HEADER_BIT : 0)
```
ifesdjeen, line 329: nit: nice formatting is now broken)

accord-core/src/main/java/accord/local/RedundantStatus.java
```diff
@@ -151,15 +155,15 @@ public enum Property
         }
 
         final boolean overrideWasOwned;
-        final boolean mergeWithPreBootstrapOrStale;
+        final boolean mergeWithUnready;
```
ifesdjeen, line 158: This is the closest line to 148. We have a comment there saying "we have 32 integer bits to use"; but we now have 64 right?

Also, do we want to relax this invariant then?

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
```diff
@@ -190,7 +186,7 @@ protected void startOnceInitialised()
         Node.Id self = node.id();
         if (permitLocalExecution() && tryIfUniversal(self))
         {
-            isPrivilegedVoteCommitting = true;
+            isPrivilegedVoteCommitting = txnId.hasPrivilegedCoordinator() && path == FAST;
```
ifesdjeen, line 189: Just to make sure: previously we would not take `nack` into account, but now we would because it can assign `isPrivilegedVoteCommitting`. Is this intended? Probably yes but I am missing context to confirm.

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -892,28 +1057,28 @@ public final DataStore unsafeGetDataStore()
...
         if (at == TxnId.NONE)
         {
-            for (Ranges rs : bootstrappedAt.values())
+            for (Ranges rs : readyAt.values())
                 Invariants.require(!ranges.intersects(rs));
         }
         else
         {
-            bootstrappedAt.floorEntry(at).getValue().containsAll(ranges);
+            readyAt.floorEntry(at).getValue().containsAll(ranges);
```
ifesdjeen, line 1078: I understand this patch only has renamed the field, but this probably should be an assert

accord-core/src/main/java/accord/api/Journal.java
```diff
@@ -67,7 +67,12 @@ enum Load
     void saveTopology(TopologyUpdate topologyUpdate, Runnable onFlush);
 
     void purge(CommandStores commandStores, EpochSupplier minEpoch);
-    void replay(CommandStores commandStores);
+
+    /**
+     * Repays all messages from journal to rehydrate CommandStores state. Returns whether it has seen (and ignored)
```
ifesdjeen, line 72: nit: repay

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -1053,6 +950,57 @@ public synchronized void initializeTopologyUnsafe(Journal.TopologyUpdate update)
         loadSnapshot(new Snapshot(shards, update.global.forNode(supplier.node.id()).trim(), update.global));
     }
 
+    public synchronized void resetTopology(Journal.TopologyUpdate update)
+    {
+        // TODO: assert
+        Snapshot current = this.current;
+        Invariants.require(update.global.epoch() == current.local.epoch());
+        ShardHolder[] shards = new ShardHolder[current.commandStores.size()];
+        int i = 0;
```
ifesdjeen, line 959: nit: unused

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -673,15 +808,45 @@ protected void updatedRedundantBefore(SafeCommandStore safeStore, RedundantBefor
...
+                if (e.getKey() > epoch)
+                    break;
+
+                Ranges remaining = e.getValue().waitingOn;
+                Ranges intersecting = remaining.slice(ranges, Minimal);
+                if (!intersecting.isEmpty())
+                {
+                    awaiting.add(e.getValue().whenDone);
+                    ranges = ranges.without(intersecting);
+                    waitingOn = waitingOn.with(intersecting);
```
ifesdjeen, line 831: Do we need to accumulate it here if we never use it?

accord-core/src/main/java/accord/local/cfk/Serialize.java
belliottsmith: [rages at intellij]

accord-core/src/main/java/accord/local/RedundantStatus.java
belliottsmith: This is very inconsistent, perhaps from a previous attempt to change the encoding. I think it's correct to say we have 32 bits to use here, now. But you're right we could match the condition to this number. I ended up not yet needing the extra bits (yet!), though, so I have for now also not upgraded the serialization format. So I might leave the restriction to short. It's a comparatively easy thing to fix later. But, I don't feel super strongly.

accord-core/src/main/java/accord/coordinate/ExecuteTxn.java
belliottsmith: You might be mixing up two different mostly unrelated changes. This particular line of code is just correctly limiting the privileged fast path commit behaviour to those cases where we really need it. For instance, if we don't have a privileged coordinator we can take the fast path whether or not we successfully commit locally. So this is an isolated minor change just to ensure we switch the behaviour on the correct condition.

If we ARE enacting a privileged fast path vote, then a NACK means we have to take the slow path (because the decision won't be durable to recovery - this optimisation relies on the coordinator recording the vote outcome before anyone else does).

accord-core/src/main/java/accord/api/Journal.java
belliottsmith: I'm fairly sure this is yours :D

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: nope!

accord-core/src/main/java/accord/api/Journal.java
ifesdjeen: Sorry!

ifesdjeen:
Since I have opened this PR originally, I don't seem to be able to click "approve", so I will just say +1 and hope that's ok

PR #253, opened 2025-09-17
Improve tracing

belliottsmith:


accord-core/src/main/java/accord/api/Tracing.java
```diff
@@ -34,35 +30,61 @@ default void trace(CommandStore store, String fmt, Object ... args)
...
+            if (++next == fmt.length())
+                throw new IllegalArgumentException("Invalid substitution declaration: % not followed by d, s or %");
+
+            char ch = fmt.charAt(next);
+            prev = next + 1;
+
+            if (ch == '%')
+            {
+                out.append('%');
+                prev = next + 1;
```
ifesdjeen: Looks like `%` handling might not work the way you expect it to:

```
public class TracingTest
{
    @Test
    public void testFormatting()
    {
        assertTracingFormatIsEquivalentToStringFormat("%% %s %s", "arg1", "arg2");
        assertTracingFormatIsEquivalentToStringFormat("%% %% %s %s", "arg1", "arg2");
    }

    private static void assertTracingFormatIsEquivalentToStringFormat(String format, Object... args)
    {
        String tracing = Tracing.safeFormat(format, args);
        String string = String.format(format, args);
        Assertions.assertEquals(tracing, string);
    }
}
```

```
org.opentest4j.AssertionFailedError: 
Expected :% arg2 %s
Actual   :% arg1 arg2
```

(assuming you intended to implement `%%` substitution)

ifesdjeen: To be honest, I am not sure why we want to parse inputs here. I think it was reasonable to just use exception id there.

ifesdjeen (approved):

ifesdjeen:
I am approving but I still think the added tracing test should pass before we commit

accord-core/src/main/java/accord/api/Tracing.java
belliottsmith: Mostly to make the exception-safe version built-in and exercised as normal, rather than an after thought we don't properly exercise.

PR #254, opened 2025-09-17
Split AsyncChain and AsyncResult; normalise AsyncResult with C* Future

belliottsmith:


accord-core/src/main/java/accord/api/AsyncExecutor.java
```diff
@@ -20,32 +20,102 @@
...
+import accord.utils.async.AsyncCallbacks.RunOrFail;
 import accord.utils.async.AsyncChain;
-import accord.utils.async.RunnableWithResult;
+import accord.utils.async.AsyncChains;
+import accord.utils.async.Cancellable;
 
 public interface AsyncExecutor extends Executor
 {
-    default AsyncChain<Void> build(Runnable task)
+    // unlike execute, throws no exceptions, nor will not wrap the runnable
```
ifesdjeen, line 37: Could you elaborate why _unlike_ `execute`? do you mean unlike `execute(executor, call)` (but then, it does call it internally, so it will have equivalent functionality?

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
```diff
@@ -71,75 +73,81 @@ public class CoordinateSyncPoint<R> extends CoordinatePreAccept<R>
...
-            SettableByCallback<SyncPoint<U>> result = new SettableByCallback<>();
-            CoordinateSyncPoint<SyncPoint<U>> coordinate = new CoordinateSyncPoint<>(node, node.someSequentialExecutor(), txnId, adapter.forDecision(node, route, SHARE, txnId, txnId), node.agent().emptySystemTxn(txnId.kind(), txnId.domain()), route, adapter, result);
-            coordinate.start();
-            return result;
+            Invariants.requireArgument(txnId.isSyncPoint());
+            TopologyMismatch mismatch = TopologyMismatch.checkForMismatch(node.topology().globalForEpoch(txnId.epoch()), txnId, route.homeKey(), route);
+            if (mismatch != null)
+                throw mismatch;
+
+            return new AsyncChains.Head<>()
```
ifesdjeen, line 137: Are there any advantages of returning `Head` here over just returning a completed async chain, i.e. `AsyncChains.success(null)?

accord-core/src/main/java/accord/api/AsyncExecutor.java
belliottsmith: The point here is that RunOrFail doesn't propagate any exceptions to the caller (except via the RunOrFail interface), including e.g. rejected execution exceptions. It also guarantees to run the parameter without modification. This is just making explicit the contract that allows it to be the terminal method that all other APIs invoke to achieve their semantics.

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
```diff
@@ -367,19 +369,22 @@ protected void updatedRedundantBefore(SafeCommandStore safeStore, RedundantBefor
...
-            Invariants.require(globalCommand != null && !globalCommand.isEmpty());
-            Command command = globalCommand.value();
-            StoreParticipants participants = command.participants().filter(LOAD, safeStore, txnId, command.executeAtIfKnown());
-            Cleanup cleanup = Cleanup.shouldCleanup(FULL, txnId, command.executeAtIfKnown(), command.saveStatus(), command.durability(), participants, unsafeGetRedundantBefore(), durableBefore());
-            Invariants.require(command.hasBeen(Applied)
-                               || cleanup.compareTo(Cleanup.TRUNCATE) >= 0
-                               || (durableBefore().min(txnId) != Universal &&
-                                      ((command.participants().stillExecutes() != null && command.participants().stillExecutes().isEmpty())
-                                      || !Route.isFullRoute(command.route()))));
+        if (progressLog instanceof DefaultProgressLog)
```
ifesdjeen, line 372: Should we maybe at least warn otherwise? That said, probably this is fine since that's used for tests only for now.

accord-core/src/main/java/accord/local/Bootstrap.java
```diff
@@ -273,7 +273,7 @@ public void fail(Ranges ranges, Throwable failure)
 
             store.agent().onFailedBootstrap(attempt, "PartialFetch", newFailures, () -> {
                 node.scheduler().selfRecurring(() -> {
-                    store.execute((PreLoadContext.Empty) () -> "Restart Bootstrap", safeStore -> restart(safeStore, newFailures.slice(allValid, Minimal), attempt + 1), store.agent());
+                    store.execute((PreLoadContext.Empty) () -> "Restart Bootstrap", safeStore -> { restart(safeStore, newFailures.slice(allValid, Minimal), attempt + 1); }, store.agent());
```
ifesdjeen, line 276: Noticed we have to disambiguate between Consumer and Function by adding brackets, but at least in Accord the Function overload is never used. Should we keep it?

accord-core/src/main/java/accord/local/CommandStores.java
```diff
@@ -1062,10 +1064,10 @@ public synchronized Supplier<EpochReady> updateTopology(Node node, Topology newT
             return () -> {
                 EpochReady ready = update.bootstrap.get();
                 return new EpochReady(ready.epoch,
-                                      flush.flatMap(ignore -> ready.metadata).beginAsResult(),
-                                      flush.flatMap(ignore -> ready.coordinate).beginAsResult(),
-                                      flush.flatMap(ignore -> ready.data).beginAsResult(),
-                                      flush.flatMap(ignore -> ready.reads).beginAsResult());
+                                      flush.flatMap(ignore -> ready.metadata),
```
ifesdjeen, line 1067: This is nice!

accord-core/src/main/java/accord/utils/async/AsyncResult.java
```diff
@@ -18,38 +18,127 @@
...
+    default <T> AsyncResult<T> flatMap(Function<? super V, ? extends AsyncResult<T>> mapper, AsyncExecutor executor)
+    {
+        return chain().flatMap(mapper.andThen(AsyncResult::chain), executor).beginAsResult();
+    }
+
+    /**
+     * When the chain has failed, this allows the chain to attempt to recover if possible.  The provided function may return a {@code null} to represent
+     * that recovery was not possible and that the original exception should propgate.
+     * <p/>
+     * This is similiar to {@link java.util.concurrent.CompletableFuture#exceptionally(Function)} but with async handling; would have the same semantics as the following
```
ifesdjeen: nit: similar

accord-core/src/main/java/accord/utils/async/AsyncResult.java
```diff
@@ -18,38 +18,127 @@
...
+    }
+
+    default <T> AsyncResult<T> flatMap(Function<? super V, ? extends AsyncResult<T>> mapper, AsyncExecutor executor)
+    {
+        return chain().flatMap(mapper.andThen(AsyncResult::chain), executor).beginAsResult();
+    }
+
+    /**
+     * When the chain has failed, this allows the chain to attempt to recover if possible.  The provided function may return a {@code null} to represent
+     * that recovery was not possible and that the original exception should propgate.
```
ifesdjeen: nit: propgate

accord-core/src/test/java/accord/utils/async/AsyncChainsTest.java
```diff
@@ -267,13 +242,13 @@ protected Cancellable start(BiConsumer<? super Integer, Throwable> callback) {
...
 
-        Assertions.assertEquals(15, AsyncChains.getBlocking(chain));
+        Assertions.assertEquals(15, AsyncChainUtils.getBlocking(chain));
     }
 
     private static void assertCombinerSize(int size, AsyncChain<?> chain)
     {
-        Assertions.assertTrue(chain instanceof AsyncChains.ReducingAsyncChain, () -> String.format("%s is not an instance of AsyncChainCombiner", chain));
-        AsyncChains.ReducingAsyncChain<?> combiner = (AsyncChains.ReducingAsyncChain<?>) chain;
+        Assertions.assertTrue(chain instanceof AsyncChains.AccumulatingReducer, () -> String.format("%s is not an instance of AsyncChainCombiner", chain));
```
ifesdjeen: nit: renamed to "AsyncChains.AccumulatingReducer" ?

accord-core/src/test/java/accord/utils/async/AsyncChainsTest.java
```diff
@@ -331,7 +306,7 @@ protected Cancellable start(BiConsumer<? super Integer, Throwable> callback)
         topLevel.add(() -> {
             AsyncResult.Settable<Integer> settable = AsyncResults.settable();
             settable.setSuccess(42);
-            return settable;
+            return settable.chain();
```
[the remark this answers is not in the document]
ifesdjeen: `Settable` does not implement `AsyncChain`, only `AsyncResult`

ifesdjeen (approved):

accord-core/src/main/java/accord/coordinate/CoordinateSyncPoint.java
belliottsmith: I don't quite follow. The AsyncChain.Head defines the latent work we haven't yet submitted, and that can be notified of the result if you call begin with a callback. Maybe this comment is attributed to the wrong spot in the code?

accord-core/src/main/java/accord/impl/InMemoryCommandStore.java
belliottsmith: We don't use this for some tests, which is the only reason we have this check. It's all test only, so not sure the noise would be worth it...

belliottsmith: (but happy to if you think it is)

accord-core/src/main/java/accord/local/Bootstrap.java
belliottsmith: good spot, let me check if we use it in C*. If not I'll remove it, would be nice to clean that up

belliottsmith: Hmm, ok. So we don't use this specific override but we do use the other Function overrides. So, I don't have a strong opinion but I do like balancing the various APIs for consistency.

PR #255, opened 2025-09-24
Fixes 250919

belliottsmith:


accord-core/src/main/java/accord/impl/DefaultLocalListeners.java
```diff
@@ -530,4 +551,177 @@ public void clear()
             }
         });
     }
+
+    @Override
+    public Iterable<TxnId> txnsWaitingOn(SaveStatus saveStatus)
```
ifesdjeen, line 556: nit: unclear if intentional or not, but this param is unused

accord-core/src/main/java/accord/impl/progresslog/DefaultProgressLog.java
```diff
@@ -552,38 +553,52 @@ public void accept(@Nullable SafeCommandStore safeStore)
...
-        Object[] prevBuffer = runBuffer;
-        int prevCount = runBufferCount;
-        int newCount = prevCount - runBufferIndex;
-        System.arraycopy(prevBuffer, runBufferIndex, newBuffer, 0, newCount);
-        runBuffer = newBuffer;
-        runBufferIndex = 0;
-        runBufferCount = newCount;
-        if (prevBuffer.length >= ArrayBuffers.MIN_BUFFER_SIZE)
-            cachedAny().forceDiscard(prevBuffer, prevCount);
+        if (runBuffer.length <= runBufferTotalCount() / 2 && runBuffer.length > 8)
```
ifesdjeen: did you mean an inverted statement here? 

I am assuming run buffer total count is always smaller than its length, or?

ifesdjeen (approved):

PR #256, opened 2025-11-03
Improve Topology Management

belliottsmith:
Merge ConfigurationService with TopologyManager to remove cyclic dependency and duplicated work. Also:
 - Improve visibility of work blocking topology processing
 - Ensure we cannot double-count peers when deciding an epoch's distributed readiness
 - Remove the possibility of distributed stalls, by processing new topologies as soon as a contiguous sequence is known locally, regardless of whether anyprior local epoch is ready to coordinate or has recorded this fact to peers. The notification of local readiness continues to be processed only once all prior epochs are ready, but we can begin using and readying later epochs immediately.


accord-core/src/main/java/accord/api/TopologyService.java
```diff
@@ -0,0 +1,65 @@
...
+
+package accord.api;
+
+import accord.local.Node;
+import accord.topology.EpochReady;
+import accord.topology.Topology;
+import accord.utils.async.AsyncResult;
+
+/**
+ * ConfigurationService is responsible for:
```
ifesdjeen: nit: TopologyService

accord-core/src/main/java/accord/topology/ActiveEpochs.java
```diff
@@ -0,0 +1,723 @@
...
+        // TODO (required): what was this trying to do? Seems potentially faulty (but maybe correct given other assumptions)
+        if (isEmpty())
+            return false;
+
+        ActiveEpoch cur = epochs[0];
+        ranges = ranges.without(cur.addedRanges);
+        if (!cur.addedRanges.containsAll(ranges))
+            return false;
+
+        for (long i = 0 ; i < epochs.length; i++)
```
ifesdjeen: It seems like we need`int` here and 
```
            ActiveEpoch retiredIn = epochs[i];
```
below

accord-core/src/main/java/accord/topology/ActiveEpochs.java
```diff
@@ -0,0 +1,723 @@
...
+     */
+    public synchronized boolean isFullyRetired(Ranges ranges)
+    {
+        // TODO (required): what was this trying to do? Seems potentially faulty (but maybe correct given other assumptions)
+        if (isEmpty())
+            return false;
+
+        ActiveEpoch cur = epochs[0];
+        ranges = ranges.without(cur.addedRanges);
+        if (!cur.addedRanges.containsAll(ranges))
```
ifesdjeen, line 123: we remove cur.addedRanges from ranges but then check if cur.addedRanges containsAll ranges

from what i understand, this could be true iif ranges are empty, right?

accord-core/src/main/java/accord/topology/ActiveEpochs.java
```diff
@@ -0,0 +1,723 @@
...
+    public TopologyRange between(long minEpoch, long maxEpoch)
+    {
+        // No epochs known to Accord
+        if (firstNonEmptyEpoch == -1 || minEpoch > currentEpoch)
+            return new TopologyRange(minEpoch(), currentEpoch, firstNonEmptyEpoch, Collections.emptyList());
+
+        minEpoch = Math.max(minEpoch, minEpoch());
+        int diff =  Math.toIntExact(currentEpoch - minEpoch + 1);
+        List<Topology> topologies = new ArrayList<>(diff);
+        for (int i = 0; minEpoch() + i <= maxEpoch && i < diff; i++)
```
ifesdjeen: Was `minEpoch + i` meant here? 
(just double checked, looks like this was pre-existing code, so this is my bad...)

accord-core/src/main/java/accord/topology/ActiveEpochs.java
```diff
@@ -0,0 +1,723 @@
...
+            topologies.add(get(minEpoch + i).global);
+
+        return new TopologyRange(minEpoch(), currentEpoch, firstNonEmptyEpoch, topologies);
```
ifesdjeen: I think this one should also be `minEpoch`

accord-core/src/main/java/accord/topology/ActiveEpochs.java
```diff
@@ -0,0 +1,723 @@
...
+    }
+
+    public Topologies preciseEpochs(Unseekables<?> select, long minEpoch, long maxEpoch, Topologies.SelectNodeOwnership selectNodeOwnership, SelectTopology selectTopology)
+    {
+        // TODO (expected): we should disambiguate minEpoch we can bump (i.e. historical epochs) and those we cannot (i.e. txnId.epoch())
+        minEpoch = Math.max(minEpoch(), minEpoch);
+        maxEpoch = validateMax(maxEpoch);
+        ActiveEpoch max = get(maxEpoch);
+
+        if (max == null)
```
ifesdjeen: I think `max` can not be `null` anymore

accord-core/src/main/java/accord/topology/PendingEpoch.java
```diff
@@ -0,0 +1,158 @@
...
+            if (isActive)
+                return AsyncResults.success(null);
+
+            long timeoutMicros = manager.node.agent().expireEpochWait(MICROSECONDS);
+            long deadlineMicros = manager.time.elapsed(MICROSECONDS) + timeoutMicros;
+            result = last = waiting.peekLast();
+            if (last == null || last.deadlineMicros < deadlineMicros)
+                waiting.add(result = new WaitingForEpoch(deadlineMicros + (timeoutMicros / 10)));
+        }
+        if (last == null)
```
ifesdjeen, line 111: Could you help me understand why it is safe for this to be outside `syncrhonized` block? `timeout` is a non-volatile field, so I am concerned that we will add `waiting` but not have a corresponding timeout.

accord-core/src/main/java/accord/topology/PendingEpoch.java
```diff
@@ -0,0 +1,158 @@
...
+    @GuardedBy("TopologyManager.this")
+    Ranges closed = Ranges.EMPTY, retired = Ranges.EMPTY;
+
+    public PendingEpoch(long epoch, TopologyManager manager)
+    {
+        this.epoch = epoch;
+        this.manager = manager;
+    }
+
+    @GuardedBy("TopologyManager")
```
ifesdjeen: Looks like it is guarded by `TopologyManager.this` (also subsequent mentions).

accord-core/src/test/java/accord/impl/TopologyManager2Test.java
```diff
@@ -0,0 +1,321 @@
...
+//
+//import accord.api.TopologyService;
+//import accord.local.Node.Id;
+//import accord.primitives.Range;
+//import accord.topology.Shard;
+//import accord.topology.Topology;
+//import org.junit.jupiter.api.Assertions;
+//import org.junit.jupiter.api.Test;
+//
+//public class TopologyManager2Test
```
ifesdjeen: Should we remove this test?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -18,1625 +18,445 @@
...
-            return prev && shard.isInFastPath(self);
-        }
+        pending.fetching.invoke((success, fail) -> {
+            if (fail == null) reportTopology(success);
+            else if (active.currentEpoch < pending.epoch && pending.topology() == null)
+            {
+                // TODO (expected): special casing of TopologyRetiredException?
+                logger.warn("Failed to fetch epoch {}. Retrying.", pending.epoch, fail);
+                node.agent().onCaughtException(fail, "Fetch epoch " + pending.epoch);
+                long retryInMicros = node.agent().retryTopologyDelay(node, 1 + ++pending.fetchAttempts, TimeUnit.MICROSECONDS);
```
ifesdjeen, line 414: looks like `pending.fetchAttempts` can be incremented concurrently, since lock is not held at that moment anymore.

accord-core/src/main/java/accord/topology/PendingEpochs.java
```diff
@@ -0,0 +1,156 @@
...
+ */
+
+package accord.topology;
+
+import accord.local.Node;
+import accord.primitives.Ranges;
+import accord.utils.Invariants;
+
+/**
+ * Thread safety is managed by a synchronized lock on this object, and extending classes can use the same lock when needed.
```
ifesdjeen: Looks like thread safety is guaranteed by `TopologyManager` synchornization, or?

accord-core/src/main/java/accord/topology/PendingEpochs.java
```diff
@@ -0,0 +1,156 @@
...
+            int capacity = Math.max(epochs.length, size * 2);
+            resize(capacity, Math.max(1, capacity / 4));
+        }
+        epochs[--start] = append;
+    }
+
+    private void resize(int newCapacity, int newStart)
+    {
+        int size = size();
+        PendingEpoch[] nextEpochs = epochs.length == newCapacity ? epochs : new PendingEpoch[newCapacity];
```
ifesdjeen, line 75: When copying to "self", could you add filling the rest of array with nulls? This should not impact functionality but is easier on the eye when debugging.

accord-core/src/main/java/accord/topology/PendingEpochs.java
```diff
@@ -0,0 +1,156 @@
...
+        epochs[end++] = append;
+    }
+
+    private void prepend(PendingEpoch append)
+    {
+        if (start == 0)
+        {
+            int size = size();
+            int capacity = Math.max(epochs.length, size * 2);
+            resize(capacity, Math.max(1, capacity / 4));
```
ifesdjeen, line 67: Any reason why not shift by one all the time? I think it is reasonable to assume that epochs will be mostly observed in order, and prepend will be only rarely exercised

accord-core/src/main/java/accord/topology/EpochReady.java
```diff
@@ -0,0 +1,120 @@
...
+
+package accord.topology;
+
+import accord.utils.Invariants;
+import accord.utils.async.AsyncResult;
+import accord.utils.async.AsyncResults;
+import accord.utils.async.NestedAsyncResult;
+
+/**
+ *
```
ifesdjeen: nit: empty comment

ifesdjeen (approved):

accord-core/src/main/java/accord/topology/PendingEpoch.java
belliottsmith: The safety depended on the fact that only one `whenActive` call could witness the `last = null`, and the worst possible race would be that the `RegisteredTimeout` might be overwritten by a (much) later one, which is implausible but would be innocuous (just wouldn't be cancelled). But, I found it was easier/clearer to just make the field managed by an `AtomicReference`. The reason to not perform this inside the lock is to avoid reentrancy from within `timeouts.registerAt`

accord-core/src/main/java/accord/topology/TopologyManager.java
belliottsmith: it should effectively have mutual exclusivity, as we can only have one fetch inflight at once, and only retry from the preceding fetch

accord-core/src/main/java/accord/topology/PendingEpochs.java
belliottsmith: We may have to prepend several epochs in a row, and this guarantees we have constant time amortised complexity (rather than quadratic), and once we run out of space appending we will reclaim the space.

PR #257, opened 2025-11-06
Fixes 251030

belliottsmith:


accord-core/src/main/java/accord/coordinate/tracking/DurabilityTracker.java
```diff
@@ -192,6 +212,41 @@ public Set<Node.Id> excluding()
...
+        ReducingRangeMap.Builder<DurabilityLevel> builder = new ReducingRangeMap.Builder<>(endInclusive, trackers.length);
+        for (int topologyIndex = 0 ; topologyIndex < topologies.size() ; ++topologyIndex)
+        {
+            for (int i = topologyOffset(topologyIndex), max = topologyOffset(topologyIndex + 1); i < max ; ++i)
+            {
+                DurabilityShardTracker tracker = trackers[i];
+                if (tracker == null)
+                    continue;
+
+                builder.append(tracker.shard.range.start(), tracker.result(self), (a, b) -> { throw new IllegalStateException(); });
```
ifesdjeen: Nit: could you add error messages to ILEs? I think they can also explain why reduce function should ILE on overlapping shards.

accord-core/src/main/java/accord/local/durability/DurabilityLevel.java
```diff
@@ -0,0 +1,140 @@
...
+               ", including=" + including +
+               ", excluding=" + excluding +
+               '}';
+    }
+
+    public static DurabilityLevel min(DurabilityLevel a, DurabilityLevel b)
+    {
+        SyncLocal local = min(a.local, b.local);
+        SyncRemote remote = min(a.remote, b.remote);
+        SortedArrayList<Node.Id> including = union(a.including, b.including);
```
ifesdjeen, line 82: I might be misunderstanding this: if we want to find out a minimum durability level, don't we want to subtract `b.including` from `a.including` (i.e. find an intersection)?

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
```diff
@@ -73,325 +86,702 @@ public class DurabilityQueue
...
+        pending.unsetParent();
+        requeue.add(pending);
+    }
+
+    void setParent(List<Pending> deferred, Pending parent)
+    {
+        for (Pending p : deferred)
+        {
+            p.unsetParent();
+            p.parent = parent;
```
ifesdjeen: Do we also need to add `deferred` to `deferred` on `parent`?

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
```diff
@@ -73,325 +86,702 @@ public class DurabilityQueue
...
-                retry(exclusiveSyncPoint, request, attempt + 1);
+                    retry(pending);
+                }
             }
-            else if (request == null)
+            finally
             {
-                // if request != null, the request will log the necessary information
-                logger.debug("{}: Successfully achieved durability for {}.", txnId, ranges);
+                synchronized (this)
```
ifesdjeen: Maybe that's just me, but could you add a DurabilityQueue qualifier to this, since we're in lambda?

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
```diff
@@ -73,325 +86,702 @@ public class DurabilityQueue
...
+
+                if (nxt == null)
+                    return test;
+                return BY_PRIORITY.compare(nxt, test) <= 0 ? nxt : test;
+            }, supersedes, submit, next);
+        }
+        return new Conflicts(next, supersedes);
+    }
+
+    private Pending minConflict(List<Pending> conflicts)
```
ifesdjeen: nit: unused (disclaimer: I have not yet checked C* implementation)

accord-core/src/main/java/accord/local/durability/DurabilityLevel.java
ifesdjeen, line 1: Would you mind to sketch up a block diagram or word explanation of how DurabilityLevel works (i.e. submit->register->enqueue->abandon/conflicts -> process  + priorities + retries)? Or a high-level bullet-point explanation.

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
```diff
@@ -73,325 +86,702 @@ public class DurabilityQueue
...
+        for (Pending p : superseded)
+        {
+            if (p.status != QUEUED) continue;
+            if (p.request != null && !p.request.isDone()) continue;
+            if (p.request == null && (p.achieved == null || p.achieved.min.remote.compareTo(Quorum) < 0)) continue;
+
+            if (p.parent != null)
+                p.parent.removeDeferred(p);
+            abandon(p, true, true);
+            return;
```
ifesdjeen, line 391: It is not obvious for me, could you elaborate why we return after pruning just one?

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
```diff
@@ -73,325 +86,702 @@ public class DurabilityQueue
...
+        pending.status = COMPLETING;
+        --activeCount;
+        requeueDeferred(pending);
+        enqueue(requeue.poll());
+    }
+
+    private void requeueDeferred(Pending parent)
+    {
+        Invariants.require(parent.status.compareTo(COMPLETING) >= 0);
+        List<Pending> deferreds = parent.deferred;
```
ifesdjeen, line 666: I am a bit confused after reading this method, since we are setting items by index to this array; do we need to call `truncate` in the end to remove all pre-existing items from there?

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -169,6 +169,7 @@ CommandStore create(int id,
     private int maxConflictsUpdates = 0;
     protected RangesForEpoch rangesForEpoch;
     protected @Nullable Ranges refuses;
+    private List<SyncPointListener> syncPointListeners;
```
ifesdjeen: I have not checked Cassandra, and maybe there's a code path where this is exercised, but do we want to actually iterate and notify anywhere?

accord-core/src/main/java/accord/utils/async/AsyncChains.java
```diff
@@ -414,7 +414,7 @@ private AccumulatingReducer(Reduce<V, V> reducer)
             this.reducer = reducer;
         }
 
-        public void add(AsyncChain<V> add)
+        public synchronized void add(AsyncChain<V> add)
```
ifesdjeen, line 417: Could you elaborate how this `synchnornized` is enough? `size` and `start` are still unsynchronized; 
I think it might be good to describe visibility guarantee, ie from what I understand we rely on volatile + snapshot of `state` with a single writer.

accord-core/src/main/java/accord/local/Catchup.java
```diff
@@ -22,51 +22,72 @@
...
-            waitingOn = redundantBefore.removeLostOrStale(waitingOn);
-            waitingOn = durableBefore.foldlWithBounds(waitingOn, (DurableBefore.Entry entry, Ranges ranges, RoutingKey entryStart, RoutingKey entryEnd) -> {
-                Ranges entryRange = Ranges.of(Range.create(entryStart, entryEnd));
-                return redundantBefore.foldlWithBounds(entryRange, (RedundantBefore.Bounds bounds, Ranges rs, RoutingKey boundStart, RoutingKey boundEnd) -> {
-                    if (bounds.maxBound(LOCALLY_APPLIED).compareTo(entry.quorumBefore) >= 0)
-                        rs = rs.without(Ranges.of(Range.create(boundStart, boundEnd)));
+            RedundantBefore redundantBefore = safeStore.redundantBefore();
+            Ranges newWaitingOn = redundantBefore.removeLostOrStale(waitingOn);
+            if (newWaitingOn != waitingOn)
+                logger.info("{}: {} are retired (or stale)", safeStore.commandStore(), newWaitingOn.without(waitingOn));
```
ifesdjeen: isn't it `waitingOn.without(newWaitingOn)` that are retired/stale?

accord-core/src/main/java/accord/local/Catchup.java
```diff
@@ -22,51 +22,72 @@
...
+
+            newWaitingOn = durableBefore.foldlWithBounds(newWaitingOn, (DurableBefore.Entry entry, Ranges ranges, RoutingKey entryStart, RoutingKey entryEnd) -> {
+                Ranges entryRanges = Ranges.of(Range.create(entryStart, entryEnd));
+                return redundantBefore.foldlWithBounds(entryRanges, (RedundantBefore.Bounds bounds, Ranges rs, RoutingKey boundStart, RoutingKey boundEnd) -> {
+                    TxnId locallyApplied = bounds.maxBound(LOCALLY_APPLIED);
+                    if (locallyApplied.compareTo(entry.quorumBefore) >= 0)
+                    {
+                        Ranges boundRanges = Ranges.of(Range.create(boundStart, boundEnd));
```
ifesdjeen, line 79: nit: `Ranges.of(Range.create(boundStart, boundEnd))` computed twice

ifesdjeen (approved):

accord-core/src/main/java/accord/local/durability/DurabilityLevel.java
belliottsmith: including/excluding are opposites, and we union both then subtract excluding from the combination of including. The reason to do it this way is we might contact different sets of nodes, so have included in one round but not another doesn't mean we haven't succeeded on that node. But if the round contacted the node but had no reply, it will be included in the `excluding` collection

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
belliottsmith: I think this is no longer present

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
belliottsmith: lambdas don't have a `this` qualifier, but I can add this if it helps

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
belliottsmith: already gone :)

accord-core/src/main/java/accord/local/durability/DurabilityLevel.java
belliottsmith: Do you mean `DurabilityLevel` or `DurabilityQueue`?

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
belliottsmith: The assumption is that we if we are adding only one item, we should only exceed the overlap threshold by one. In reality this isn't necessarily true, given how we define overlaps, so we could keep going if you like.

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
belliottsmith: No, the list is discarded; we only set the items so we can loop over the new parents (that are non-overlapping); we could probably do this more efficiently with more work, but I will simply leave a comment explaining for now

accord-core/src/main/java/accord/local/CommandStore.java
belliottsmith: whoops...

accord-core/src/main/java/accord/utils/async/AsyncChains.java
belliottsmith: I didn't end up using it here so I could just revert, but to explain: the only thing we worry about here is collecting items to reduce in the list, which can safely be done asynchronously if this method is synchronized. We must already assume that all of the items we want to collect are already collected when we start the chain, and that this is done by a single owner, otherwise the whole thing is sort of broken at a more fundamental level.

accord-core/src/main/java/accord/local/durability/DurabilityLevel.java
ifesdjeen: Sorry yes I did mean DurabilityQueue

accord-core/src/main/java/accord/local/durability/DurabilityQueue.java
ifesdjeen: Either way works for me really; I can see why we might want to amortize pruning so that makes sense.

ifesdjeen (approved):
Thank you for the adjustments and added comments, this all looks great!

PR #259, opened 2025-11-21
Topology exceptions

belliottsmith:


accord-core/src/main/java/accord/topology/TopologyMismatch.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private TopologyMismatch(String message, Throwable cause)
+    {
+        super(message, cause);
+    }
+
+    @Override
+    public TopologyMismatch rethrowable()
+    {
+        Invariants.require(this.getClass() == TopologyMismatch.class);
```
ifesdjeen: Maybe could just make the class final above?

accord-core/src/main/java/accord/topology/TopologyMismatch.java
```diff
@@ -0,0 +1,82 @@
...
+import javax.annotation.Nullable;
+
+import accord.primitives.Ranges;
+import accord.primitives.Routables;
+import accord.utils.Invariants;
+import accord.utils.UnhandledEnum;
+
+public class TopologyMismatch extends TopologyException
+{
+    public enum TopologyMatch { LATEST, ANY }
```
ifesdjeen: Could you add a short description? From what I understand, any is used for sync points right?

accord-core/src/main/java/accord/primitives/KnownMap.java
```diff
@@ -286,25 +286,45 @@ private static Known reduceKnownFor(MinAndMaxKnown bounds, @Nullable Known prev)
...
+                    int i = prev.find(start, FAST);
+                    if (i < 0)
+                        return prev.slice(0, -1 - i);
+
+                    if (prev.domain() == Routable.Domain.Key)
+                        return prev.slice(0, i + 1);
+
+                    Range r = prev.get(i).asRange();
+                    Range newR = r.start().equals(start) ? start.asRange() : r.newRange(r.start(), start);
+                    return prev.slice(0, i).with((Participants)Ranges.of(newR));
```
ifesdjeen: I might be misunderstanding, but shouldn't it be `I + 1` here?

accord-core/src/main/java/accord/primitives/KnownMap.java
```diff
@@ -286,25 +286,45 @@ private static Known reduceKnownFor(MinAndMaxKnown bounds, @Nullable Known prev)
...
+                        return prev.slice(-1 - i, prev.size());
 
-                Range r = prev.get(i).asRange();
-                Range newR = r.start().equals(start) ? start.asRange() : r.newRange(r.start(), start);
-                return prev.slice(0, i).with((Participants)Ranges.of(newR));
+                    if (prev.domain() == Routable.Domain.Key)
+                        return prev.slice(i + 1, prev.size());
+
+                    Range r = prev.get(i).asRange();
+                    Range newR = r.end().equals(end) ? end.asRange() : r.newRange(r.end(), end);
```
ifesdjeen: Again I might be misunderstanding, but shouldn't it be `end, r.end()`?

ifesdjeen (approved):
+1 with several questions

PR #262, opened 2025-12-11
Make sure to always call execute before results

dcapwell:


accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -104,7 +109,7 @@ protected void checkWithTimeout(Runnable fn)
             }
             catch (ExecutionException e)
             {
-                throw new PropertyError(propertyError(this, e.getCause()));
+                throw new PropertyError(propertyError(this, e.getCause()), e.getCause());
```
dcapwell, line 112: the cause got logged in the message but not the stack trace, so adding in for better debugging

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -213,10 +218,11 @@ private static String propertyError(Common<?> input, Throwable cause, Object...
         return sb.toString();
     }
 
-    private static String statefulPropertyError(StatefulBuilder input, Throwable cause, Object state, List<String> history)
+    private static String statefulPropertyError(StatefulBuilder input, Throwable cause, Object state, int failingStep, List<String> history)
     {
         StringBuilder sb = propertyErrorCommon(input, cause);
         sb.append("Steps: ").append(input.steps).append('\n');
+        sb.append("Failing Step: ").append(failingStep).append('\n');
```
dcapwell, line 225: helpful for larger histories; you can just call the process method with the right debug step and run the test to that point

accord-core/src/test/java/accord/utils/Property.java
```diff
@@ -1126,7 +1138,7 @@ public Setup<State, SystemUnderTest> next(RandomSource rs)
                                 for (Setup<State, SystemUnderTest> s : unknownWeights)
                                     weights.put(s, unknownWeightGen.nextInt(rs));
                             }
-                            nonConditional = Gens.pick(weights);
+                            nonConditional = weights.isEmpty() ? null : Gens.pick(weights);
```
dcapwell, line 1141: found a bug in my other branch where weights was empty so we would fail as we can't pick from an empty set

maedhroz (approved):

maedhroz (approved):

PR #263, opened 2025-12-12
Fixes 251205

belliottsmith:


accord-core/src/main/java/accord/topology/Topology.java
```diff
@@ -560,22 +578,29 @@ public <P, T> T foldlWithDefault(Routables<?> select, IndexedTriFunction<Shard,
...
 
-            if (nextai > amax + 1)
-                accumulator = function.apply(null, param, accumulator, -1 - bi);
+            if (nextai > ai)
+                accumulator = function.apply(ifNull, param, accumulator, -1 - bi);
             accumulator = function.apply(shards[supersetIndexes[bi]], param, accumulator, bi);
-            amax = nextai + 1;
-            ai = nextai;
+
+            Range r = ranges.get(bi);
```
ifesdjeen: Can we use `bi` here to address `ranges` not `subsetOfRanges`?

accord-core/src/main/java/accord/topology/TopologyMismatch.java
```diff
@@ -45,31 +47,19 @@ public TopologyMismatch rethrowable()
...
-            default: throw new UnhandledEnum(result);
-            case PENDING_REMOVAL:
-                message = String.format("Txn attempted to access keys or ranges that are being removed in epoch %d (%s)", topology.epoch(), keysOrRanges);
-                break;
-            case NOT_KNOWN:
-                message = String.format("Txn attempted to access keys or ranges that are not known in the epoch %d (%s)", topology.epoch(), keysOrRanges);
-                break;
-        }
+        if (kind.isSyncPoint() || !e.all.ranges.containsAll(keysOrRanges))
+            message = String.format("Txn attempted to access keys or ranges that are being removed in epoch %d (%s)", topology.epoch(), keysOrRanges.without(e.all.ranges));
```
ifesdjeen: (discussed off-thread) 
Here, it seems like ranges aren't present in active, so the message might have to be "not known" (i.e. below clause)

accord-core/src/main/java/accord/topology/PendingEpoch.java
```diff
@@ -171,12 +177,35 @@ public int stripe()
...
             this.deadlineMicros = deadlineMicros;
         }
+
+        AsyncChain<Void> chainImmediatelyElse(@Nullable AsyncExecutor executor)
+        {
+            AsyncChain<Void> chain = result.chain();
+            if (result.isDone())
+                return chain;
+
+            waitingUpdater.incrementAndGet(this);
```
ifesdjeen, line 202: Do we also want to decrement on completion?

ifesdjeen (approved):

PR #264, opened 2025-12-15
bitset impl are not consistent, so updated so both implementations are in-sync and to fix tests in Cassandra

dcapwell:


accord-core/src/main/java/accord/utils/LargeBitSet.java
```diff
@@ -174,9 +179,12 @@ public boolean unset(int i)
         return true;
     }
 
+    @Override
     public final boolean get(int i)
     {
         int index = indexOf(i);
+        if (index == UNKNOWN)
```
dcapwell: all read APIs default to the negative case if you ask out of bounds.  This keeps things consistent and easy to work with.  

The write APIs all reject as its clear you did something wrong.

One thing we talked about in slack was exposing capacity, but that also implies that every caller gets updated to check that in the loop where as the pattern has been 

```
for (int index = set.nextSetBit(0); index != -1; index = set.nextSetBit(index))
```

accord-core/src/main/java/accord/utils/SmallBitSet.java
```diff
@@ -40,29 +40,51 @@ public long bits()
...
-        long bit = bit(i);
+        long bit = bitSafe(i);
         boolean result = 0 == (bits & bit);
         bits |= bit;
         return result;
     }
 
     @Override
-    public void setRange(int from, int to)
+    public void setRange(int fromInclusive, int toExclusive)
```
dcapwell, line 70: `to` was exclusive but wouldn't work if you gave the last index as it would cause the bits to be `0`; found in fuzz testing

accord-core/src/main/java/accord/utils/SmallBitSet.java
```diff
@@ -40,29 +40,51 @@ public long bits()
...
     {
-        bits |= ~(-1L << to) & (-1L << from);
+        validateIndex(fromInclusive);
+        Invariants.requireArgument(fromInclusive <= toExclusive, "from > to (%s > %s)", fromInclusive, toExclusive);
+        if (fromInclusive == toExclusive)
+            return;
```
dcapwell, line 76: making this consistent with large; which does this same work

accord-core/src/main/java/accord/utils/SmallBitSet.java
```diff
@@ -40,29 +40,51 @@ public long bits()
...
+
+        long maskTo = (toExclusive >= 64) ? -1L : ~(-1L << toExclusive);
+        bits |= maskTo & (-1L << fromInclusive);
```
dcapwell: in slack benedict recommend we use the logic from large: `|= (-1L >>> (64 - (to & 63))) & (-1L << (from & 63))`

Need to test this out

accord-core/src/test/java/accord/utils/SimpleBitSetTest.java
```diff
@@ -167,9 +234,27 @@ public int nextSetBit(int fromIndex)
...
+        {
+            StringBuilder sb = new StringBuilder();
+            for (int i = 0; i < size; i++)
+            {
+                boolean m = model.get(i);
+                boolean s = sut.get(i);
+                if (m == s)
+                    sb.append(m ? 1 : 0);
+                else
+                    sb.append("(expected=").append(m ? 1 : 0).append(", actual=").append(s ? 1 : 0).append(')');
```
dcapwell, line 268: this format was hacked together to aid in debugging `setRange` issue.  Was simple and showed exactly what was needed, but open to other formats.

accord-core/src/main/java/accord/utils/SmallBitSet.java
```diff
@@ -40,29 +40,50 @@ public long bits()
...
     }
 
     @Override
-    public void setRange(int from, int to)
+    public void setRange(int fromInclusive, int toExclusive)
     {
-        bits |= ~(-1L << to) & (-1L << from);
+        validateIndex(fromInclusive);
+        Invariants.requireArgument(fromInclusive <= toExclusive, "from > to (%s > %s)", fromInclusive, toExclusive);
+        if (fromInclusive == toExclusive)
```
belliottsmith, line 75: this shouldn't be needed I don't *think*, as the below should simply become 0

accord-core/src/main/java/accord/utils/SmallBitSet.java
```diff
@@ -40,29 +40,50 @@ public long bits()
...
+            return;
+
+        bits |= (-1L >>> (64 - (toExclusive & 63))) & (-1L << (fromInclusive & 63));
     }
 
+    @Override
     public boolean get(int i)
     {
+        if (i >= 64) return false;
```
belliottsmith: shouldn't this be IIOBE?

accord-core/src/main/java/accord/utils/SmallBitSet.java
```diff
@@ -40,29 +40,50 @@ public long bits()
...
 
+    private static long bitSafe(int i)
+    {
+        validateIndex(i);
+        return bit(i);
+    }
+
+    private static void validateIndex(int i)
+    {
+        if (i >= 64 || i < 0)
```
belliottsmith: `i >>> 6 > 0` or `(i & ~63) != 0`?

accord-core/src/main/java/accord/utils/SmallBitSet.java
dcapwell: i mostly kept it as the logic was in large, i can remove and confirm with the tests (tests don't hit this case, so will update)

accord-core/src/main/java/accord/utils/SmallBitSet.java
dcapwell: my argument is that read methods shouldn't IIOBE and should return the negative case as it simplifies the calling logic (would have to update caller to find the capacity and always make sure to avoid crossing it).

`get` isn't the strongest argument for that, but `nextSetBit` i feel is.  Having every loop check if `nextSetBit` returned the last element pushes this to every caller and adds clutter all over.

`accord.utils.SortedArrays#fromSimpleBitSet`

```
for (int i = bitSet.firstSetBit(); i >= 0 ; i = bitSet.nextSetBit(i + 1))
```

`accord.coordinate.AbstractCoordination#inflight`

```
for (int i = expectingReply.nextSetBit(0) ; i >= 0 ; i = expectingReply.nextSetBit(i + 1))
```

All these examples are unsafe if the read apis throws IIOBE and really expect out of bounds to return negative case; this is the common assumption throughout the current usage.

We spoke before in my previous patch and you were not a fan of the `get` method having different semantic from `nextSetBit`, so returning the negative case keeps the read logic simple/safe for callers

dcapwell: we spoke in slack, switched to IIOBE.  if `i == size()` then we accept and return `-1` for `nextSetBit` but `get` now rejects.

If `i > size()` we will reject for all input

accord-core/src/main/java/accord/utils/SmallBitSet.java
dcapwell: went with `i >>> 6 > 0` to be consistent

accord-core/src/main/java/accord/utils/LargeBitSet.java
```diff
@@ -443,10 +448,16 @@ private int indexOf(int i)
     {
         int index = i >>> 6;
         if (index >= length)
-            throw new IndexOutOfBoundsException(String.format("%d >= %d", index, length));
+            throw new IndexOutOfBoundsException(String.format("Unable to access bit %d; must be between 0 and %d", i, size() - 1));
```
dcapwell, line 451: this keeps the 2 impl consistent even with the error type + msg.  I also find `index >= length` harder to work with than the actual input / size

accord-core/src/main/java/accord/utils/SmallBitSet.java
dcapwell: not safe it seems; need to keep this logic

```
		40: Clear
		41: setRange(0, 0)
```

```
Expected :0
Actual   :64
```

dcapwell:
got +1 from Benedict in slack

PR #266, opened 2026-01-12
Fixes 251228

belliottsmith:
InMemoryRangeIndex: Cassandra's secondary range index integration appears to be faulty in some way, producing sync points that have missing dependencies. Introduce a burn-test integrated in-memory implementation we can be more efficiently confident in.
Also:
     - SemiSyncValue
     - Simulate CFR loading/listening
     - KeyDeps.forEach
     - Timestamp.tryParse
Fix:
     - Minor burn test edge case when cancelling bootstrap and resuming
     - NotifyWaitingOn stillExecutes->stillWaitsOn
     - Privileged fast path commit failure should not send stable message
     - DefaultLocalListeners iterators do not handle multiple waiters on the same transaction correctly
Improve:
     - Don't compute Deps for SyncPoint Accept or Recover phase
     - Catchup should report bounds to progress log

### DIFF
--- a/accord-core/src/main/java/accord/impl/cfr/IdEntry.java
+++ b/accord-core/src/main/java/accord/impl/cfr/IdEntry.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package accord.impl.cfr;
+
+import accord.primitives.Range;
+import accord.primitives.Ranges;
+import accord.primitives.SaveStatus;
+import accord.primitives.Status;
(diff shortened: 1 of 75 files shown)


accord-core/src/main/java/accord/impl/cfr/IdEntry.java
```diff
@@ -0,0 +1,81 @@
...
+import accord.primitives.Ranges;
+import accord.primitives.SaveStatus;
+import accord.primitives.Status;
+import accord.primitives.Timestamp;
+import accord.primitives.TxnId;
+import accord.utils.Invariants;
+
+public abstract class IdEntry extends TxnId
+{
+    static final int SAVE_SATUS_SHIFT = Status.Durability.ENCODING_BITS;
```
ifesdjeen: nit: "satus"

accord-core/src/main/java/accord/impl/cfr/IdSingleEntry.java
```diff
@@ -0,0 +1,61 @@
...
+    @Override
+    public IdEntry id()
+    {
+        return this;
+    }
+
+    @Override
+    IdEntry copy()
+    {
+        accord.impl.cfr.IdSingleEntry copy = new accord.impl.cfr.IdSingleEntry(this, range);
```
ifesdjeen: nit: do not need to fully qualify here

accord-core/src/main/java/accord/impl/cfr/InMemoryRangeSummaryIndex.java
```diff
@@ -0,0 +1,345 @@
...
+            }
+            return f.apply(id, accumulator);
+        }
+    }
+
+    public void update(Command prev, Command updated, boolean force)
+    {
+        if (!force
+            && updated.saveStatus() == prev.saveStatus()
+            && (updated.saveStatus().known.executeAt().isDecided() && !prev.saveStatus().known.executeAt().isDecided())
```
ifesdjeen: how can isDecided be different if save statuses are identity-equal?

ifesdjeen (approved):

PR #268, opened 2026-02-24
CASSANDRA-21183 Safely regain ranges that were previously owned by a node

alanwang67:
Currently nodes that are regaining ranges can result in subtle edge cases when we still have command stores left over from previous epochs that overlap with the range we are regaining. 

To address this we: 
1. Allow the regaining of ranges only if the epoch in which we previously owned the range is already retired
2. Mark as unsafe to read the ranges that we are regaining for the command store that previously owned the range and prevent future `setSafeToRead` invocations to revert this change. 
3. Throw an error when we query two command stores for the same range

### DIFF
--- a/accord-core/src/main/java/accord/local/CommandStore.java
+++ b/accord-core/src/main/java/accord/local/CommandStore.java
@@ -187,7 +187,7 @@ public abstract class CommandStore implements AbstractAsyncExecutor, SequentialA
      * But they may still be ordered for other key ranges they participate in.
      */
     private NavigableMap<Timestamp, Ranges> safeToRead = emptySafeToRead();
-    private Ranges retiredRanges =  Ranges.EMPTY;
+    private Ranges regainedRanges =  Ranges.EMPTY;
     private final Set<Bootstrap> bootstraps = Collections.synchronizedSet(new DeterministicIdentitySet<>());
     @Nullable private RejectBefore rejectBefore;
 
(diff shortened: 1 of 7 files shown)


accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -283,6 +288,62 @@ protected Executor executor()
         return Runnable::run;
     }
 
+    public static Map<Id, Ranges> getAdditions(Topology current, Topology next)
```
belliottsmith: perhaps move this to Topology, and call it e.g. computeNodeAdditions?

accord-core/src/main/java/accord/topology/TopologyManager.java
```diff
@@ -283,6 +288,62 @@ protected Executor executor()
...
+            Ranges added = next.rangesForNode(node).without(prev);
+            if (added.isEmpty())
+                continue;
+
+            additions.put(node, added);
+        }
+        return additions;
+    }
+
+    public List<Long> epochsWeAreRegainingRangesFor(Topology curr, Topology next)
```
belliottsmith: I think this and block... aren't used, so could remove them (and the associated notify)?

accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -187,6 +187,7 @@ CommandStore create(int id,
      * But they may still be ordered for other key ranges they participate in.
      */
     private NavigableMap<Timestamp, Ranges> safeToRead = emptySafeToRead();
+    private Ranges retiredRanges =  Ranges.EMPTY;
```
belliottsmith: We probably don't need this; we can simply query redundantBefore, as this won't happen on any fast paths

alanwang67: Maybe I should come up with a better name, but this variable is for keeping track of the ranges that we marked as unsafe to read in the `CommandStore` because we are regaining the range for it, so it is a subset of the actual `retiredRanges`. I use this value to ensure that whenever we set ranges as `safeToRead` we exclude these ranges as it is no longer safe to read from them if there exists a `CommandStore` with a greater epoch that has the same range. If we query redundantBefore, I'm not sure if we will be able to get this value.

PR #269, opened 2026-03-11
CASSANDRA-21212 Deleted Retired Command Stores

alanwang67:
We delete `CommandStore` when the range they own is fully removed and all the epochs in which it owned the ranges are retired, but keep a lighter weight map to store previously owned ranges to maintain the invariants in https://issues.apache.org/jira/browse/CASSANDRA-21183.


aweisberg (approved):
+1

PR #270, opened 2026-03-25
CASSANDRA-19433 Nodetool cleanup can drop data Accord might need

alanwang67:
Adds a new method to get ranges owned by CommandStores to be used by `nodetool cleanup` to determine when it is safe to cleanup a range.

### DIFF
--- a/accord-core/src/main/java/accord/local/CommandStore.java
+++ b/accord-core/src/main/java/accord/local/CommandStore.java
@@ -1219,7 +1219,7 @@ public abstract class CommandStore implements AbstractAsyncExecutor, SequentialA
         return ImmutableSortedMap.copyOf(build);
     }
 
-    private static ImmutableSortedMap<Timestamp, Ranges> purgeHistory(NavigableMap<Timestamp, Ranges> in, Ranges remove)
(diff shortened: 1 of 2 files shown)


accord-core/src/main/java/accord/local/CommandStore.java
```diff
@@ -1219,7 +1219,7 @@ private static <T extends Timestamp> ImmutableSortedMap<T, Ranges> purgeAndInser
         return ImmutableSortedMap.copyOf(build);
     }
 
-    private static ImmutableSortedMap<Timestamp, Ranges> purgeHistory(NavigableMap<Timestamp, Ranges> in, Ranges remove)
+    protected static ImmutableSortedMap<Timestamp, Ranges> purgeHistory(NavigableMap<Timestamp, Ranges> in, Ranges remove)
```
aweisberg, line 1222: This could be package private.

aweisberg (changes requested):

accord-core/src/main/java/accord/local/CommandStore.java
alanwang67: I think setting it to package private means that we can't access it from `CommandStores`.

PR #271, opened 2026-04-13
Fixes 260226

belliottsmith:


aweisberg:
+1

accord-core/src/main/java/accord/local/Commands.java
```diff
@@ -836,8 +840,8 @@ protected static WaitingOn initialiseWaitingOn(SafeCommandStore safeStore, TxnId
     protected static Update updateWaitingOn(SafeCommandStore safeStore, ICommand waiting, Timestamp executeAt, WaitingOn.Update initialise)
     {
         RedundantBefore redundantBefore = safeStore.redundantBefore();
-        TxnId minWaitingOnTxnId = initialise.minWaitingOnTxnId();
-        if (minWaitingOnTxnId != null && redundantBefore.hasLocallyRedundantDependencies(initialise.minWaitingOnTxnId(), executeAt, waiting.participants().waitsOn()))
+        TxnId minWaitingOnTxnId = initialise.minWaitingOnTxn();
+        if (minWaitingOnTxnId != null && redundantBefore.hasLocallyRedundantDependencies(initialise.minWaitingOnTxn(), executeAt, waiting.participants().waitsOn()))
```
ifesdjeen: probably more of a nit, but still, on L844, `initialise.minWaitingOnTxn()` can just be `minWaitingOnTxnId`

accord-core/src/main/java/accord/local/CatchupHard.java
```diff
@@ -0,0 +1,176 @@
...
+import accord.utils.async.AsyncResults.SettableResult;
+
+import static accord.local.RedundantStatus.Property.LOG_UNAVAILABLE;
+import static accord.local.RedundantStatus.SomeStatus.LOG_UNAVAILABLE_ONLY;
+import static accord.primitives.Routables.Slice.Minimal;
+
+/**
+ * Catch up this data store to the point of any quorum
+ */
+public class CatchupHard
```
ifesdjeen, line 55: Do we want to override `onNewFailure` here like on Bootstrap?

belliottsmith: Good question. I think for now it's fine, it will just not retry, which is less important for CatchupHard. But I will be considering this logic again soon, and will consider what might be good to add here.

PR #276, opened 2026-05-18
CoordinateTransaction should not abort if LocalExecute fails, esp. in…

belliottsmith:
… case of LogUnavailableException


aweisberg (approved):
+1

aweisberg (approved):
+1 TY!

PR #277, opened 2026-05-28
CASSANDRA-21410: ShardDurability.markDefunct() called O(N²) times across topology updates, causing log spam and OOM in tests

dcapwell:


belliottsmith:
The test is tightly coupled to the implementation, and highly specific to this behaviour that only primarily affects tests anyway. As such I don't think it is valuable enough to justify its inclusion in the patch.

belliottsmith (approved):
+1 if test removed

belliottsmith:
If a test is strongly desired (I don't have a strong opinion, I am only opposed to introducing such a complex coupled test for a behaviour so trivial): an alternative might be to have `markDefunct` return a boolean indicating if it has been marked defunct, and to have a trivial test to confirm that any shard scheduler returns true exactly once. Although, we don't generally introduce such trivial tests.

PR #278, opened 2026-06-08
CommandChanges.shouldCleanup short-circuits to NO if there is no data…

belliottsmith:
…, but this is incorrect as Cleanup.EXPUNGE may have dropped the data and the record must receive cleanup EXPUNGE and be reported as ERASED. This can lead to lienarizability violations.

### DIFF
--- a/accord-core/src/main/java/accord/impl/DefaultLocalListeners.java
+++ b/accord-core/src/main/java/accord/impl/DefaultLocalListeners.java
@@ -42,6 +42,7 @@ import accord.local.SafeCommand;
 import accord.local.SafeCommandStore;
 import accord.primitives.SaveStatus;
 import accord.primitives.TxnId;
+import accord.utils.ArrayBuffers.BufferList;
 import accord.utils.AsymmetricComparator;
 import accord.utils.Invariants;
 import accord.utils.btree.BTree;
@@ -292,7 +293,6 @@ public class DefaultLocalListeners implements LocalListeners
         static final RegisteredComplexListener[] NO_LISTENERS = new RegisteredComplexListener[0];
         RegisteredComplexListener[] listeners = NO_LISTENERS;
         int count, length;
-        boolean notifying;
 
         /**
          * Append to the end of the list; if we aren't reentering from notify then if the next position
@@ -307,10 +307,11 @@ public class DefaultLocalListeners implements LocalListeners
             Invariants.require(listeners[index] == remove);
             listeners[index] = null;
             remove.index = -1;
-            // we don't decrement length even if count==length so as to simplify reentry
-            --count;
-            if (Invariants.isParanoid() && !notifying) checkIntegrity();
-            return count > 0 || notifying ? this : null;
+            if (--count == 0)
+                return null;
+
+            if (Invariants.isParanoid()) checkIntegrity();
+            return this;
         }
 
         /**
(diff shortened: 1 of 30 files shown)


alanwang67:
+1

ifesdjeen (approved):

accord-core/src/main/java/accord/impl/DefaultLocalListeners.java
```diff
@@ -307,10 +307,17 @@ RegisteredComplexListeners remove(RegisteredComplexListener remove)
             Invariants.require(listeners[index] == remove);
             listeners[index] = null;
             remove.index = -1;
-            // we don't decrement length even if count==length so as to simplify reentry
-            --count;
-            if (Invariants.isParanoid() && !notifying) checkIntegrity();
-            return count > 0 || notifying ? this : null;
+            if (--count == 0)
```
ifesdjeen, line 310: Looks like `DefaultLocalListeners` would now fail because of this behavior change 

```
Caused by: org.opentest4j.AssertionFailedError: expected: <[14, 15]> but was: <[15, 14]>
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
	at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1124)
	at accord.impl.LocalListenersTest$TestCase.notifyOne(LocalListenersTest.java:187)
	at accord.impl.LocalListenersTest$TestCase.run(LocalListenersTest.java:126)
	at accord.impl.LocalListenersTest.lambda$testOne$0(LocalListenersTest.java:80)
```

Looks like we need to copy with shift, or are we ok with the order change here?

accord-core/src/main/java/accord/local/RedundantBefore.java
```diff
@@ -1129,8 +1132,14 @@ public Participants<?> expectToOwn(TxnId txnId, @Nullable Timestamp executeAt, P
 
     public boolean isAtLeast(RedundantBefore atLeast)
     {
+        if (!ranges().containsAll(atLeast.ranges()))
+            return false;
+
         return foldl((b, v, al, e) -> {
             return al.foldl(Ranges.of(b.range), (lb, v2, ub) -> {
+                if (!v2 || ub.endEpoch > lb.endEpoch || ub.startEpoch != lb.startEpoch)
```
ifesdjeen, line 1140: I might be not fully understanding, but is it enough to check for `!=`, don't we need to check for `ub`  as smaller than?

ifesdjeen (approved):

accord-core/src/main/java/accord/local/RedundantBefore.java
belliottsmith: end epoch is only ever Long.MAX_VALUE or the actual end epoch. So we could tighten it to `(ub.endEpoch != lb.endEpoch && lb.endEpoch != Long.MAX_VALUE)` if you like

accord-core/src/main/java/accord/impl/DefaultLocalListeners.java
belliottsmith: Oh, yeah I don't think we guarantee (or rely on any such guarantee) any order of visitation. That was a function of some shenanigans we did to avoid entering the hash map twice (once to notify, once to remove). But, we cannot both not hold the lock while notifying and not enter the hash map twice, and we have to not hold the lock, so there's no point to the extra complexity.

belliottsmith: I'll take a look at the test

belliottsmith: The test is actually quite sophisticated, and tests behaviour when listeners modify other listeners. I think having the same level of test sophistication without a predefined visitation order will be hard, so I will restore the guaranteed visitation order.

PR #280, opened 2026-07-09
CASSANDRA-21553 Fix Sorted Arrays Linear Intersection gives wrong results

alanwang67:
Fixes bug with `linearIntersection`

### DIFF
--- a/accord-core/src/main/java/accord/utils/SortedArrays.java
+++ b/accord-core/src/main/java/accord/utils/SortedArrays.java
@@ -526,8 +526,9 @@ public class SortedArrays
             {
                 if (!hasMatch)
                     return left.length == 0 ? left : NO_INTS;
-                else
-                    return Arrays.copyOfRange(left, leftStart, leftIdx);
+                if (leftStart == 0 && leftIdx == left.length)
+                    return left;
+                return Arrays.copyOfRange(left, leftStart, leftIdx);
             }
         }
 
--- a/accord-core/src/main/java/accord/utils/SortedArrays.java
+++ b/accord-core/src/main/java/accord/utils/SortedArrays.java
@@ -480,6 +480,8 @@ public class SortedArrays
      */
     public static int[] linearIntersection(int[] left, int leftStart, int leftEnd, int[] right, int rightStart, int rightEnd, ArrayBuffers.IntBuffers buffers)
     {
+        Invariants.require(leftStart <= leftEnd && rightStart <= rightEnd);
+
         if (leftEnd - leftStart > rightEnd - rightStart)
         {
             int[] tmp = left;
--- a/accord-core/src/main/java/accord/utils/SortedArrays.java
+++ b/accord-core/src/main/java/accord/utils/SortedArrays.java
@@ -519,6 +519,7 @@ public class SortedArrays
                 {
                     resultSize = leftIdx - leftStart;
                     result = buffers.getInts(resultSize + Math.min(leftEnd - leftIdx, rightEnd - rightIdx));
+                    leftIdx++;
                     System.arraycopy(left, leftStart, result, 0, resultSize);
                     break;
                 }
--- a/accord-core/src/main/java/accord/utils/SortedArrays.java
+++ b/accord-core/src/main/java/accord/utils/SortedArrays.java
@@ -518,8 +518,8 @@ public class SortedArrays
                 else
                 {
                     resultSize = leftIdx - leftStart;
-                    result = buffers.getInts(resultSize + Math.min(leftEnd - leftIdx, rightEnd - rightIdx));
                     leftIdx++;
+                    result = buffers.getInts(resultSize + Math.min(leftEnd - leftIdx, rightEnd - rightIdx));
                     System.arraycopy(left, leftStart, result, 0, resultSize);
                     break;
                 }
--- a/accord-core/src/main/java/accord/utils/SortedArrays.java
+++ b/accord-core/src/main/java/accord/utils/SortedArrays.java
@@ -515,9 +515,9 @@ public class SortedArrays
                 }
                 else
                 {
-                    resultSize = leftIdx++;
+                    resultSize = leftIdx - leftStart;
                     result = buffers.getInts(resultSize + Math.min(leftEnd - leftIdx, rightEnd - rightIdx));
-                    System.arraycopy(left, 0, result, 0, resultSize);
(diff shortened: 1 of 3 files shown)


accord-core/src/test/java/accord/utils/SortedArraysTest.java
```diff
@@ -205,6 +205,68 @@ public void testLinearIntersection()
...
+
+            int leftStart = left.length == 1 ? 0 : rs.nextInt(0, left.length);
+            int leftEnd = left.length == 1 ? 1 : rs.nextInt(leftStart + 1, left.length + 1);
+
+            int rightStart = left.length == 1 ? 0 : rs.nextInt(0, left.length);
+            int rightEnd = left.length == 1 ? 1 : rs.nextInt(rightStart + 1, left.length + 1);
+
+            var actual = SortedArrays.linearIntersection(left, leftStart, leftEnd,
+                    left, rightStart, rightEnd,
+                    ArrayBuffers.uncachedInts());
```
dcapwell: can you fix the formatting?  line 260 and 261 should be indented to be under `left` above (after the `(`)

dcapwell (approved):

accord-core/src/main/java/accord/utils/SortedArrays.java
```diff
@@ -515,9 +517,9 @@ public static int[] linearIntersection(int[] left, int leftStart, int leftEnd, i
                 }
                 else
                 {
-                    resultSize = leftIdx++;
+                    resultSize = leftIdx - leftStart;
```
aweisberg, line 520: Not incrementing leftIdx at all in this branch has the result the loop has to continue a little farther in the next iteration to increment it and allow the loop to terminate.

aweisberg:
Of the changes here if you revert `leftIdx == left.length` the test only fails 47% of the time. If you revert the identity fast path it doesn't fail at all and the caller actually uses the identity fast path to determine that no changes were made so it's part of the contract of this method.

Also allegedly
```
  - Latent flakiness in testIntLinearIntersection: rs.nextInt(a[0] - 100,
  a[a.length - 1] + 100). a comes from sortedUniqueIntegerArray(0), i.e.
  Gens.ints().all() over the full int range, so a[0] - 100 can underflow and
      // identity contract: left is returned unchanged iff nothing was filtered out
      int[] subset = { 1, 3 };
      Assertions.assertSame(subset,
          SortedArrays.linearIntersection(subset, 0, subset.length, right, 0, right.length, ArrayBuffers.uncachedInts()));
  }

  Step 1 — the generator can produce a near-MIN_VALUE element.
  sortedUniqueIntegerArray(0) is Gens.arrays(Integer.class, Gens.ints().all()).unique().ofSizeBetween(0, 100) and Gens.ints().all() is RandomSource::nextInt — uniform over
  all 2³² ints. Underflow needs a[0] - 100 < Integer.MIN_VALUE, i.e. a[0] ∈ [MIN_VALUE, MIN_VALUE + 99] — exactly 100 of the 2³² values. Since a is sorted, that element
  lands at a[0].

  Say the draw includes -2147483600 (= MIN_VALUE + 48), so after sorting a[0] = -2147483600, and say a[a.length-1] = 1500000000.

  Step 2 — the loop is entered.
  int mutations = a.length == 0 ? 0 : rs.nextInt(1, a.length + 1); → non-zero for any non-empty a, so while (mutations != 0) runs.

  Step 3 — the else branch is reached.
  if (rs.nextBoolean()) picks the "sample from a" branch, else the random-value branch. With mutations averaging ~25, the probability that the else branch is never reached
  is ~2⁻²⁵ — negligible.

  Step 4 — the subtraction silently wraps, in int arithmetic. This is the crux:

  b.add(rs.nextInt(a[0] - 100, a[a.length - 1] + 100));

  a[0] - 100 = -2147483600 - 100. The true value -2147483700 isn't representable, so Java wraps it: -2147483700 + 2³² = 2147483596 (= MAX_VALUE - 51). No exception here —
  this is silent two's-complement wraparound.

  So the arguments become minInclusive = 2147483596, maxExclusive = 1500000100.

  Step 5 — nextInt's guard trips, unconditionally. RandomSource.java:83-87:
```

aweisberg:
Allegedly you can reproduce the flaky test with seed `-2123152008219503357L` and it reproduced after 448 runs of the test.

aweisberg:
accord.utils.SortedArraysTest.testIntLinearIntersection fails with seed -5740875519181198423

accord-core/src/main/java/accord/utils/SortedArrays.java
```diff
@@ -515,9 +517,10 @@ public static int[] linearIntersection(int[] left, int leftStart, int leftEnd, i
...
                     result = buffers.getInts(resultSize + Math.min(leftEnd - leftIdx, rightEnd - rightIdx));
-                    System.arraycopy(left, 0, result, 0, resultSize);
+                    leftIdx++;
```
aweisberg, line 521: Move this before getInts so it doesn't allocate an extra slot

accord-core/src/test/java/accord/utils/SortedArraysTest.java
```diff
@@ -205,6 +205,88 @@ public void testLinearIntersection()
...
+
+            int[] expected = intersection.stream().mapToInt(i->i).toArray();
+            Arrays.sort(expected);
+
+            // First two conditions cover the case where we are expected to return the exact copy
+            // if the intersection is an exact match of the smaller array
+            if (bStart == 0 && bEnd == bInt.length && aSet.containsAll(bSet))
+                Assertions.assertSame(bInt, SortedArrays.linearIntersection(aInt, aStart, a.length, bInt, bStart, bInt.length, new ArrayBuffers.IntBufferCache(4, 1 << 14)));
+            else if (aStart == 0 && aEnd == aInt.length && bSet.containsAll(aSet))
+                Assertions.assertSame(aInt, SortedArrays.linearIntersection(aInt, aStart, a.length, bInt, bStart, bInt.length, new ArrayBuffers.IntBufferCache(4, 1 << 14)));
```
aweisberg: This seems to be unreachable

belliottsmith:
Might be worth asking Claude to look for equivalent issues, as I think this is probably the second or third time this bug has been fixed, maybe due to copy/paste versions of similar methods.

aweisberg (approved):
+1 TY!

PR #285, opened 2026-08-06
CASSANDRA-21551: DefaultLocalListeners.remove adds nulls which then leads to NPE when .clear gets called

dcapwell:


accord-core/src/main/java/accord/impl/DefaultLocalListeners.java
```diff
@@ -558,8 +558,16 @@ public void clear()
             RegisteredComplexListeners listeners = complexListeners.remove(key);
             if (listeners != null)
             {
-                for (int i = 0 ; i < listeners.count ; i++)
-                    listeners.listeners[i].index = -1;
+                if (Invariants.isParanoid()) listeners.checkIntegrity();
+                // listeners may contain null holes (removal nulls a slot in place); iterate the
```
belliottsmith: a single line comment should be possible here

dcapwell: fixed

alanwang67:
For posterity, this is the same bug as [CASSANDRA-21502](https://issues.apache.org/jira/browse/CASSANDRA-21502).

alanwang67:
+1

cscotta:
+1

PR #286, opened 2026-08-11
CASSANDRA-21566 Accord range based tests have their inclusive field flipped

alanwang67:
Fix test failures related to AbstractRangesTest.java and KeyDepsTest.java

### DIFF
--- a/accord-core/src/test/java/accord/primitives/KeyDepsTest.java
+++ b/accord-core/src/test/java/accord/primitives/KeyDepsTest.java
@@ -113,11 +113,11 @@ public class KeyDepsTest
     }
 
     private static void testMerge(long seed, int uniqueTxnIdsRange, int epochRange, int hlcRange, int nodeRange,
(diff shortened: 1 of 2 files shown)


accord-core/src/test/java/accord/primitives/KeyDepsTest.java
```diff
@@ -222,19 +223,20 @@ public void testForEachOnUniqueEndInclusive()
     }
 
     @Test
-    public void testForEachOnUniqueStartInclusive()
+    public void testForEachOnUniqueStartInclusiveEndExclusive()
```
dcapwell, line 226: this is copy/paste of testForEachOnUniqueStartExclusiveEndInclusive and doesn't actually do `StartInclusiveEndExclusive`; shifting the keys by 1 doesn't actually test that code path.

The only way to test this is to create a new class so you can set the global state as this is now a global config set at class load time...

dcapwell (approved):

aweisberg (approved):
+1

